PR #3, opened 2018-04-06
feat(debug_helper): Adding debug helper when a get call fails

antoaravinth:
Fix for https://github.com/kentcdodds/react-testing-library/issues/31

@kentcdodds Let me know if things are good to go!

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -146,4 +146,30 @@ test('using jest helpers to check element attributes', () => {
   ).toThrowError()
 })
 
+test('test the debug helper prints the dom state here', () => {
+  const Large = `<div>
+        ${Array.from({length: 7000}, (v, key) => key).map(() => {
+          return `<div data-testid="debugging" data-otherid="debugging">
+                Hello World!
+            </div>`
+        })}
+    </div>`
+
+  const {getByText} = render(Large) // render large DOM which exceeds 7000 limit
+  expect(() => expect(getByText('not present')).toBeInTheDOM()).toThrowError()
+
+  const Hello = `<div data-testid="debugging" data-otherid="debugging">
+        Hello World!
+    </div>`
+  const {getByTestId} = render(Hello)
+  process.env.DEBUG_PRINT_LIMIT = 10 // user should see `. . .`
+  expect(() => expect(getByTestId('not present')).toBeInTheDOM()).toThrowError()
+
+  const {getByLabelText} = render(Hello)
+  process.env.DEBUG_PRINT_LIMIT = 10000 // user shouldn't see `. . .`
+  expect(() =>
+    expect(getByLabelText('not present')).toBeInTheDOM(),
+  ).toThrowError()
+})
+
 /* eslint jsx-a11y/label-has-for:0 */
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -194,6 +194,7 @@ test('using jest helpers to check element class names', () => {
 })
 
 test('test the debug helper prints the dom state here', () => {
+  const originalDebugPrintLimit = process.env.DEBUG_PRINT_LIMIT
   const Large = `<div>
         ${Array.from({length: 7000}, (v, key) => key).map(() => {
           return `<div data-testid="debugging" data-otherid="debugging">
@@ -209,14 +210,19 @@ test('test the debug helper prints the dom state here', () => {
         Hello World!
     </div>`
   const {getByTestId} = render(Hello)
-  process.env.DEBUG_PRINT_LIMIT = 10 // user should see `. . .`
-  expect(() => expect(getByTestId('not present')).toBeInTheDOM()).toThrowError()
+  process.env.DEBUG_PRINT_LIMIT = 5 // user should see `. . .`
+  expect(() => expect(getByTestId('not present')).toBeInTheDOM()).toThrowError(
+    /\.\.\./,
+  )
 
   const {getByLabelText} = render(Hello)
   process.env.DEBUG_PRINT_LIMIT = 10000 // user shouldn't see `. . .`
   expect(() =>
-    expect(getByLabelText('not present')).toBeInTheDOM(),
+    expect(getByLabelText('not present')).toBeInTheDOM(/^((?!\.\.\.).)*$/),
   ).toThrowError()
+
+  //all good replacing it with old value
+  process.env.DEBUG_PRINT_LIMIT = originalDebugPrintLimit
 })
 
 /* eslint jsx-a11y/label-has-for:0 */
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -210,13 +210,13 @@ test('test the debug helper prints the dom state here', () => {
         Hello World!
     </div>`
   const {getByTestId} = render(Hello)
-  process.env.DEBUG_PRINT_LIMIT = 5 // user should see `. . .`
+  process.env.DEBUG_PRINT_LIMIT = 5 // user should see `...`
   expect(() => expect(getByTestId('not present')).toBeInTheDOM()).toThrowError(
     /\.\.\./,
   )
 
   const {getByLabelText} = render(Hello)
-  process.env.DEBUG_PRINT_LIMIT = 10000 // user shouldn't see `. . .`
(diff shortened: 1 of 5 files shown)


src/queries.js
```diff
@@ -1,4 +1,6 @@
 import {matches} from './matches'
+import prettyFormat from 'pretty-format' // eslint-disable-line
```
sompylasar: Why eslint-disable-line here?

src/queries.js
```diff
@@ -129,11 +165,40 @@ function queryByAltText(container, alt) {
...
+      }),
+    )
+    throw new Error(
+      `Unable to find an element with the alt text: ${alt} \nHere is the state of your container: \n${htmlElement}`,
+    )
   }
   return el
 }
 
+function checkDebugLimit(debugContent) {
```
sompylasar: The name is confusong: it shouldn't be "check" (would return boolean), it is more of a "trimToDebugLimit".

src/queries.js
```diff
@@ -129,11 +165,40 @@ function queryByAltText(container, alt) {
...
+  const limit = process.env.DEBUG_PRINT_LIMIT
```
sompylasar: What if the limit needed in browser where there's no env, e.g. via Cypress?

I suggest exporting a mutable "defaultOptions" object from the module instead, like Node's util.inspect.defaultOptions.

src/queries.js
```diff
@@ -129,11 +165,40 @@ function queryByAltText(container, alt) {
...
+    if (limit <= contentLength)
+      // there are more chars we have stripped, just show to the user
+      debugContent += '\n . . .'
+  } else {
+    // default limit is 7000
+    const contentLength = debugContent.length
+    debugContent = debugContent.slice(0, 7000)
+    if (contentLength >= 7000)
+      // there are more chars we have stripped, just show to the user
+      debugContent += '\n . . .'
```
sompylasar: The above "if" and "else" are essentially copy-paste except `limit` is replaced with a default value.

I suggest pulling the "magic number" into a "defaultOptions" object and use it like so:
```
    const limit = defaultOptions.limit;
    // OR:  const limit = defaultOptions.limit || 7000;
    const contentLength = debugContent.length
    debugContent = debugContent.slice(0, limit)
    if (limit <= contentLength)
      // there are more chars we have stripped, just show to the user
      debugContent += '\n . . .'
```

antoaravinth: Yes, I agree. Can be improved. This is just a rough draft, I have to clean up few things. Thanks for your comments.

src/queries.js
antoaravinth: Yes, thats a good point.

src/queries.js
antoaravinth: So basically `pretty-format` is not part of the package.json, dependencies but its part of `jest` I believe. Hence I have added the `eslint-disable-line`, yes, I need to clean up a bit by adding the exact eslint-ignore.

sompylasar: But what if I don't use jest with dom-testing-library?

antoaravinth: Sorry, `pretty-format` is from `jest-matcher-utils` (https://github.com/facebook/jest/blob/master/packages/jest-matcher-utils/package.json#L14) which is our dependencies. Sorry for the rushed comment, saying it as `jest`!

antoaravinth:
@kentcdodds Hey Kent.. @sompylasar suggested the following:

> What if the limit needed in browser where there's no env, e.g. via Cypress?
> 
> I suggest exporting a mutable "defaultOptions" object from the module instead, like Node's util.inspect.defaultOptions.

w.r.t to process env for `DEBUG_PRINT_LIMIT`, as per his suggestion, `DEBUG_PRINT_LIMIT` wont work in no env mode. I was thinking it for a moment, if we do like `defaultOptions`, then we need to export it. So was thinking, what your thoughts on that? Yes to me, thats a good way to support across all the env.

src/queries.js
lgandecki: I'm not against using pretty-format, I'm just trying to understand if there is a reason to not put it in package.json as a direct dependency. Then there is no need to fight around eslint complaining (rightly). And if jest-matcher-utils switch to something else internally and stop requiring pretty-format, our package will still work.

kentcdodds: Agreed

antoaravinth: @lgandecki Yes, I agree. Will make the changes. Thanks your comments!

antoaravinth:
Ping, any thoughts on my last comment? Guess people missed it ;)

antoaravinth:
Ping, any thoughts on my last comment? Guess people missed it ;)  cc @kentcdodds

kentcdodds:
I have thoughts. Patience. Not all of us have time to address this within hours or days. We'll get to it when we can 😉

One thing you could do is rebase your branch and fix merge conflicts 😀

antoaravinth:
@kentcdodds Oops! Sorry thought people had missed it. will do the rebase and update the branch.

src/__tests__/element-queries.js
```diff
@@ -193,4 +193,30 @@ test('using jest helpers to check element class names', () => {
...
+    </div>`
+  const {getByTestId} = render(Hello)
+  process.env.DEBUG_PRINT_LIMIT = 10 // user should see `. . .`
+  expect(() => expect(getByTestId('not present')).toBeInTheDOM()).toThrowError()
+
+  const {getByLabelText} = render(Hello)
+  process.env.DEBUG_PRINT_LIMIT = 10000 // user shouldn't see `. . .`
+  expect(() =>
+    expect(getByLabelText('not present')).toBeInTheDOM(),
+  ).toThrowError()
```
kentcdodds: Make sure to reset `DEBUG_PRINT_LIMIT` to clean things up.

It should look something like this:

```javascript
test('test the debug helper prints the dom state here', () => {
  const originalDebugPrintLimit = process.env.DEBUG_PRINT_LIMIT

  // the test contents

  process.env.DEBUG_PRINT_LIMIT = originalDebugPrintLimit
})
```

That way it doesn't mess up other tests.

src/__tests__/element-queries.js
```diff
@@ -193,4 +193,30 @@ test('using jest helpers to check element class names', () => {
...
+
+  const {getByText} = render(Large) // render large DOM which exceeds 7000 limit
+  expect(() => expect(getByText('not present')).toBeInTheDOM()).toThrowError()
+
+  const Hello = `<div data-testid="debugging" data-otherid="debugging">
+        Hello World!
+    </div>`
+  const {getByTestId} = render(Hello)
+  process.env.DEBUG_PRINT_LIMIT = 10 // user should see `. . .`
+  expect(() => expect(getByTestId('not present')).toBeInTheDOM()).toThrowError()
```
kentcdodds: This is a good assertion, but we should add another assertion to make sure the error includes the `...`

You can do this by adding a regex to the `toThrowError` call:

`.toThrowError(/\.\.\./)`

In the one below we should do the opposite


```javascript
.toThrowError(/^((?!\.\.\.).)*$/)
```

I'm _pretty_ sure that should verify that there is no `...` in the error message.

sompylasar:
@kentcdodds Did we go with unchecked process.env for configuration?

sompylasar:
> That said, anyone using this in the browser will be using a bundler so I'm pretty sure this'll be fine.

I thought about this, too. Although I'm not 100% sure.

@antoaravinth Could you please check if this holds true at least for Cypress? You can do so from `cypress-testing-library`: clone it, update `dom-testing-library` dependency, `npm run -s test:cypress` or `test:cypress:open`. Thanks!

antoaravinth:
@sompylasar Yes it will. I have checked anyways:

```
Opening Cypress...
  (Tests Starting)


  dom-testing-library commands
    ✓ getByPlaceholderText (537ms)
    ✓ getByText (129ms)
    ✓ getByLabelText (495ms)
    ✓ getByAltText (134ms)
    ✓ getByTestId (134ms)


  5 passing (2s)


  (Tests Finished)

  - Tests:           5
  - Passes:          5
  - Failures:        0
  - Pending:         0
  - Duration:        2 seconds
  - Screenshots:     0
  - Video Recorded:  false
  - Cypress Version: 2.1.0


  (All Done)
```

sompylasar:
@antoaravinth Awesome! :+1:

PR #4, opened 2018-04-06
Add toHaveClass custom matcher (closes #2)

### LINKED ISSUE #2: .toHaveClass custom matcher
At risk of sounding like I want to bloat the set of custom matchers provided by this lib, I'd like to propose `.toHaveClass('dropdown-toggle')`. The `class` attribute in html is designed to have multiple classes encoded in it, and it's often nice to be able to reliably check if an element has a certain class while ignoring the rest of what's encoded in the `class` attribute value.

```javascript
// ...
// <button data-testid="delete-button" type="submit" class="btn btn-danger">
//   Delete item
// </button>
expect(getByTestId('delete-button')).toHaveClass('btn-danger')
```

Raising it as an issue first to see if it's considered useful for this lib, and also to discuss potential names:

- `toHaveClass`
- `toIncludeClass`
- `toContainClass`
- something else?

Or all of the above but ending with `*ClassName` instead of just `*Class`.

gnapse:
As discussed in #2 (closes #2)

One thing to note about it: I did this custom matcher so that it can even be used to match more than one class at the same time:

```javascript
// <button data-testid="delete-button" class="btn extra btn-danger">
//   Delete item
// </button>

// This passes, even though `"btn btn-danger"` are not contiguous in the actual class attr
expect(getByTestId('delete-button')).toHaveClass('btn btn-danger')
```

I think this makes it much more useful, and was not difficult to support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ when a real user uses it.
   * [`toBeInTheDOM`](#tobeinthedom)
   * [`toHaveTextContent`](#tohavetextcontent)
   * [`toHaveAttribute`](#tohaveattribute)
+  * [`toHaveClass`](#tohaveclass)
   * [Custom Jest Matchers - Typescript](#custom-jest-matchers---typescript)
 * [`TextMatch`](#textmatch)
 * [`query` APIs](#query-apis)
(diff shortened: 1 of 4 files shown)


antoaravinth:
@gnapse You are fond of jest custom matchers :) Looks cool though!

gnapse:
Yeah, I'd like to strike some balance between usefulness and avoiding bloating this lib, so I don't really think I'll propose any new ones for the time being, but these two (`.toHaveAttribute` and now `.toHaveClass`) came up from tests I was writing myself with `react-testing-library`, and I thought that testing for these two things is really common. Plus they fit very nicely in the context testing the dom, which is what this new library extracted from `react-testing-library` is for. That was a nice move!

PR #10, opened 2018-04-08
feat(waitForElement): a wrapper around MutationObserver

sompylasar:
* Add `waitForElement` implementation, tests, docs
* Add `mutationobserver-shim` to `dependencies` to provide
`MutationObserver` in jest tests where `jsdom` has no built-in support
for it: https://github.com/jsdom/jsdom/issues/639
* Update @sompylasar contributions


README.md
```diff
@@ -305,6 +306,50 @@ The default `interval` is `50ms`. However it will run your callback immediately
 on the next tick of the event loop (in a `setTimeout`) before starting the
 intervals.
 
+### `waitForElements`
+
+Defined as:
+
+```typescript
+function waitForElements<T>(
```
kentcdodds: Normally I think people will be looking for a single element. Should this be called `waitForElement`?

In addition, we should show an example (in the FAQ?) of getting an array of elements:

```javascript
const elements = waitForElement(() => [
  getByText('username'),
  getByText('password')
])
```

README.md
```diff
@@ -305,6 +306,50 @@ The default `interval` is `50ms`. However it will run your callback immediately
...
+// wait until the callback does not throw an error and returns a truthy value. In this case, that means
+// it'll wait until we can get a form control with a label that matches "username"
+// the difference from `wait` is that it reacts to DOM changes in the container
+// and returns the value returned by the callback
+const element = await waitForElements(() => getByLabelText(container, 'username'))
+element.value = 'chucknorris'
+// ...
+```
+
+Using `MutationObserver` is more efficient than polling the DOM at regular intervals with `wait`.
```
kentcdodds: Let's add a link to the MDN docs for this so folks know that it's actually a browser feature. Most people have probably never heard of it. I only heard of it a few weeks ago when building the cypress command waiter.

README.md
```diff
@@ -305,6 +306,50 @@ The default `interval` is `50ms`. However it will run your callback immediately
...
+
+The default `callback` is a no-op function (used like `await waitForElements()`). This can
+be helpful if you only need to wait for the next DOM change (see `mutationObserverOptions` to learn which changes are detected).
```
kentcdodds: Add a link to `mutationObserverOptions` :)

README.md
```diff
@@ -305,6 +306,50 @@ The default `interval` is `50ms`. However it will run your callback immediately
...
+
+The default `timeout` is `4500ms` which will keep you under
+[Jest's default timeout of `5000ms`](https://facebook.github.io/jest/docs/en/jest-object.html#jestsettimeouttimeout).
+
+The default `mutationObserverOptions` is `{subtree: true, childList: true}` which will detect
+additions and removals of child elements (including text nodes) in the `container` and any of its descendants.
+It won't detect attribute changes unless you add `attributes: true` to the options.
```
kentcdodds: Perfect :+1:

package.json
```diff
@@ -39,7 +39,8 @@
   },
   "devDependencies": {
     "jest-in-case": "^1.0.2",
-    "kcd-scripts": "^0.36.1"
+    "kcd-scripts": "^0.36.1",
+    "mutationobserver-shim": "^0.3.2"
```
kentcdodds: This should be a regular dependency.

src/waitForElements.js
```diff
@@ -0,0 +1,43 @@
...
+  callback = () => {},
+  {
+    container = document,
+    timeout = 4500,
+    mutationObserverOptions = {subtree: true, childList: true},
+  } = {},
+) {
+  return new Promise((resolve, reject) => {
+    // Disabling eslint prefer-const below: either prefer-const or no-use-before-define triggers.
+    let lastError, observer, timer // eslint-disable-line prefer-const
```
kentcdodds: Really odd that both of these rules trigger 🤔 I don't see how this code is in violation. An ESLint bug?

src/waitForElements.js
```diff
@@ -0,0 +1,43 @@
...
+        // If `callback` throws an error, wait for the next mutation or timeout.
+      }
+    }
+    function onTimeout() {
+      onDone(lastError || new Error('Timed out in waitForElements.'), null)
+    }
+    timer = setTimeout(onTimeout, timeout)
+    observer = new MutationObserver(onMutation)
+    observer.observe(container, mutationObserverOptions)
+  })
```
kentcdodds: Could we run `onMutation` above this line? That way if there's no reason to wait for a mutation to proceed we don't have to wait for one.

sompylasar:
> So I guess we're not providing them with the mutation observer polyfill ourselves? They have to install that? I'd rather polyfill it for people I think.

We could bundle it I guess, yes.

src/waitForElements.js
sompylasar: If I have it as `const`, I have to initialize it at definition, i.e. at this line:
```
timer = setTimeout(onTimeout, timeout)
```
but then:
```
    function onDone(error, result) {
      clearTimeout(timer)
```
is a usage before definition of `timer`

sompylasar:
@kentcdodds All requested changes addressed here: https://github.com/kentcdodds/dom-testing-library/pull/10/commits/1f46848a95735fd8c13a167b72f5b7109decf02d

README.md
```diff
@@ -305,6 +306,62 @@ The default `interval` is `50ms`. However it will run your callback immediately
...
+  getByLabelText(container, 'password'),
+])
+```
+
+Using [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) is more efficient than polling the DOM at regular intervals with `wait`. This library sets up a [`'mutationobserver-shim'`](https://github.com/megawac/MutationObserver.js) on the global `window` object for cross-platform compatibility with older browsers and the [`jsdom`](https://github.com/jsdom/jsdom/issues/639) that is usually used in Node-based tests.
+
+The default `callback` is a no-op function (used like `await waitForElement()`). This can
+be helpful if you only need to wait for the next DOM change (see [`mutationObserverOptions`](#mutationobserveroptions) to learn which changes are detected).
+
+The default `container` is the global `document`. Make sure the elements you wait for will be attached to it, or set a different `container`.
```
kentcdodds: Could you add a simple example of this. It's probably the most common use case

kentcdodds: Or maybe just make the example above use it like that.

sompylasar:
Turned out the no-argument `waitForElement()` stopped working as advertised after we've added the synchronous `onMutation()` call. Fixed it here: https://github.com/kentcdodds/dom-testing-library/pull/10/commits/88bab8733cbbf393404fc3fd37062163348cd2c1

Added `container` option to the examples here: https://github.com/kentcdodds/dom-testing-library/pull/10/commits/b7158cf1b497043dff526ad0b31de31d566a508c

src/index.js
```diff
@@ -6,4 +6,5 @@ export {queries}
 
 export * from './queries'
 export * from './wait'
+export * from './waitForElement'
```
kentcdodds: This is a bit of a nit-pick and I really should have an eslint rule for this, but due to issues with file casing + different operating systems + git, I prefer filenames to always use kebab-case rather than camelCase. Could you update the files to use kebab-case please?

src/__tests__/waitForElement.js
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  expect(callback).toHaveBeenCalledTimes(1 + mutationsAndCallbacks.length)
+  expect(successHandler).toHaveBeenCalledTimes(1)
+  expect(successHandler).toHaveBeenCalledWith(testEl)
+  expect(errorHandler).toHaveBeenCalledTimes(0)
+  expect(container).toMatchSnapshot()
+  expect(testEl.parentNode).toBe(container)
+
+  await promise
```
kentcdodds: Rather than `await promise`, I'd prefer `return promise`. Maybe I'm being irrational here, but it feels better to me.

src/__tests__/waitForElement.js
```diff
@@ -0,0 +1,278 @@
...
+// adds special assertions like toBeInTheDOM
+import '../extend-expect'
+import {render} from './helpers/test-utils'
+
+async function skipSomeTime(delayMs) {
+  await new Promise(resolve => setTimeout(resolve, delayMs))
+}
+
+async function skipSomeTimeForMutationObserver(delayMs = 50) {
+  // Using `setTimeout` >30ms instead of `wait` here because `mutationobserver-shim` uses `setTimeout` ~30ms.
```
kentcdodds: Thanks for this comment :+1:

src/__tests__/waitForElement.js
```diff
@@ -0,0 +1,278 @@
...
+    timeout: 300,
+    mutationObserverOptions: {attributes: true},
+  }).then(successHandler, errorHandler)
+
+  expect(callback).toHaveBeenCalledTimes(1)
+  expect(successHandler).toHaveBeenCalledTimes(0)
+  expect(errorHandler).toHaveBeenCalledTimes(0)
+
+  container.setAttribute('data-test-attribute', 'something changed once')
+  await skipSomeTimeForMutationObserver(200)
```
kentcdodds: Does this time need to be so long? I'd prefer it to be as short as possible so these tests run quickly :)

src/__tests__/waitForElement.js
```diff
@@ -0,0 +1,278 @@
...
+
+  expect(callback).toHaveBeenCalledTimes(2)
+  expect(successHandler).toHaveBeenCalledTimes(0)
+  expect(errorHandler).toHaveBeenCalledTimes(0)
+
+  container.setAttribute('data-test-attribute', 'something changed twice')
+  await skipSomeTimeForMutationObserver(150)
```
kentcdodds: Same here.

src/__tests__/waitForElement.js
```diff
@@ -0,0 +1,278 @@
...
+    timeout: 300,
+    mutationObserverOptions: {attributes: true},
+  }).then(successHandler, errorHandler)
+
+  expect(callback).toHaveBeenCalledTimes(1)
+  expect(successHandler).toHaveBeenCalledTimes(0)
+  expect(errorHandler).toHaveBeenCalledTimes(0)
+
+  container.setAttribute('data-test-attribute', 'something changed once')
+  await skipSomeTimeForMutationObserver(200)
```
kentcdodds: And here

src/__tests__/waitForElement.js
```diff
@@ -0,0 +1,278 @@
...
+
+  expect(callback).toHaveBeenCalledTimes(2)
+  expect(successHandler).toHaveBeenCalledTimes(0)
+  expect(errorHandler).toHaveBeenCalledTimes(0)
+
+  container.setAttribute('data-test-attribute', 'something changed twice')
+  await skipSomeTimeForMutationObserver(150)
```
kentcdodds: And here :)

src/__tests__/waitForElement.js
```diff
@@ -0,0 +1,278 @@
...
+  // The promise callbacks are expected to be called asyncronously.
+  expect(successHandler).toHaveBeenCalledTimes(0)
+  expect(errorHandler).toHaveBeenCalledTimes(0)
+  await wait()
+  expect(successHandler).toHaveBeenCalledTimes(1)
+  expect(successHandler).toHaveBeenCalledWith(container.firstChild)
+  expect(errorHandler).toHaveBeenCalledTimes(0)
+
+  container.setAttribute('data-test-attribute', 'something changed once')
+  await skipSomeTimeForMutationObserver(200)
```
kentcdodds: Annnnd here

src/__tests__/waitForElement.js
sompylasar: No it doesn't, will reduce to 50.

src/index.js
sompylasar: Okay, but tell this to React team. It's for the sake of searchability that the file is literally named the same as the thing it contains and exports. Better have a lint rule that ensures the names strictly match case with the filesystem (and there is such, I think in the eslint-plugin-import).

kentcdodds: I have told them and they rejected it. But I've had too many people bitten by file casing issues across operating systems to do anything else.

kentcdodds: If searchability is key then you say: "Oh, here's a thing called `waitForElement` in the code, so I'll turn that to kebab-case and search for `wait-for-element`." Believe me, I've felt these pains before as well. We just have different solutions.

In addition, it's why I'm starting to avoid `default export`s now because the greppability of a named export is better.

sompylasar: I just select the thing, cmd+shift+f, no need to convert. Especially when doing large refactorings. Default exports I agree can become messy if coders don't follow to name them the same at import site, it's a matter of discipline which is not enforced by the program structure.

src/__tests__/waitForElement.js
sompylasar: I'm changing this, but if this line was the only one with `await` expression, then `eslint` would say `async` function must have an `await`. We are lucky that there are more `await`s in all the tests.

P.S. I don't agree with this rule because `async` functions have their use for the error handling; `async` function is just the one that is wrapped in a `return new Promise(...)`.

src/index.js
sompylasar: 👉  https://github.com/kentcdodds/dom-testing-library/pull/10/commits/1252d0824d454b27ff8ed92390af681984069668

src/__tests__/waitForElement.js
sompylasar: 👉 https://github.com/kentcdodds/dom-testing-library/pull/10/commits/414a3dca4f75abb64033d7e614255b66e42bf694

src/__tests__/waitForElement.js
sompylasar: You're welcome 😉

src/__tests__/waitForElement.js
sompylasar: 👉 https://github.com/kentcdodds/dom-testing-library/pull/10/commits/edfb24144f1bc1a34dfa4d3e6e91ff1b45a5d81d

sompylasar:
@kentcdodds Thanks for merging! I'm currently more interested in having `cypress-testing-library` updated to use `waitForElement` and published. I'm already using the updated Cypress code with `waitForElement` in my project (and it works!), but it would be nice to install this piece of code via `cypress-testing-library`.

gnapse:
Wow `waitForElement` looks awesome! I did not know something like `MutationObserver` existed.

kentcdodds:
> I'm currently more interested in having cypress-testing-library updated to use waitForElement and published.

Sounds great! I'll be happy to set that project up with auto-releases and such as soon as it's thoroughly tested :+1: I've got the tooling all set up so that should be pretty seamless :+1:

sompylasar:
Can't promise that, now I only have time to make the product app I'm working on thoroughly tested with Cypress.

kentcdodds:
Well, if you can't get cypress-testing-library well tested, then feel free to just get it in a working state and I'd happily manually publish a `0.0.0` so you can at least install and use it until we can get a working version :+1:

sompylasar:
@kentcdodds Deal!

PR #13, opened 2018-04-09
added fireEvent util

jomaxx:
**What**: added fireEvent utility

**Why**: https://github.com/kentcdodds/dom-testing-library/issues/5

**How**: src/events.js

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 


src/__tests__/events.js
```diff
@@ -0,0 +1,287 @@
+import React from 'react'
+import {mount, unmount} from './helpers/react'
+import {fireEvent} from '../'
+
+describe('Clipboard Events', () => {
+  ;['copy', 'paste'].forEach(eventName => {
+    it(`fires ${eventName}`, () => {
+      const node = document.createElement('input')
```
kentcdodds: This code is very similar throughout the test file. Could we extract it to a "testEvent" function?

```javascript
function testEvent({eventName, elementType}) {
  const node = document.createElement(elementType)
  const spy = jest.fn()
  node.addEventListener(eventName.toLowerCase(), spy)
  fireEvent[eventName](node)
  expect(spy).toHaveBeenCalledTimes(1)
}
```

Then most of the tests could be dramatically simplified to something like:

```javascript
describe('Clipboard Events', () => {
  ;['copy', 'paste'].forEach(
    eventName => {
      it(`fires ${eventName}`, () => {
        testEvent({eventName, elementType: 'input'})
      })
    },
  )
})
```

**Alternatively** it may even not be a bad idea to abstract it further:

```javascript
const eventTypes = [
  {type: 'Clipboard', events: ['copy', 'paste'], elementType: 'input'},
  {
    type: 'Composition',
    events: ['compositionEnd', 'compositionStart', 'compositionUpdate'],
    elementType: 'input',
  },
  // etc...
]

eventTypes.forEach(({type, events, elementType}) => {
  describe(`${type} Events`, () => {
    events.forEach(eventName => {
      it(`fires ${eventName}`, () => {
        const node = document.createElement(elementType)
        const spy = jest.fn()
        node.addEventListener(eventName.toLowerCase(), spy)
        fireEvent[eventName](node)
        expect(spy).toHaveBeenCalledTimes(1)
      })
    })
  })
})
```

I think doing this will make it easier to add cases and know which tests are for one-off situations :+1: It will also make this file shorter probably :)

src/__tests__/helpers/react.js
```diff
@@ -0,0 +1,13 @@
+import ReactDOM from 'react-dom'
+
+const node = document.body.appendChild(document.createElement('div'))
```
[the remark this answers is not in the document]
jomaxx: i'll move the `document.body.appendChild` call to `mount`

src/__tests__/events.js
```diff
@@ -0,0 +1,287 @@
...
+  it(`fires doubleClick`, () => {
+    const node = document.createElement('button')
+    const spy = jest.fn()
+    node.addEventListener('dblclick', spy)
+    fireEvent.doubleClick(node)
+    expect(spy).toHaveBeenCalledTimes(1)
+  })
+})
+
+describe('React Events', () => {
```
antoaravinth: May be we need to have these test cases in `react-testing-library`? I'm just wondering because, anyway people are gonna use `dom-testing-library` for several frameworks (`angular`,`preact`,`vue` etc), do we have to add those frameworks testing here in future too?

antoaravinth: Also by moving it to `react-testing-library`, I guess people have good chance of looking at these test cases and understanding how to test events with `react`.

src/__tests__/helpers/react.js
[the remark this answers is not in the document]
sompylasar: And also add the mirroring removeChild to unmount then?

src/__tests__/events.js
sompylasar: Also moving it to react-testing-library will move the respective package dependencies that don't make sense in just the dom-testing-library

antoaravinth: Yes, also what we have created in `src/__tests__/helpers/react.js` is already part of `react-testing-library` (`render`,`unmount`).

kentcdodds: Personally I think it'd be better if we could avoid react in here and have a simple render method that creates DOM nodes manually. It doesn't need to be complicated.

src/events.js
```diff
@@ -0,0 +1,352 @@
+// fallback to Event
```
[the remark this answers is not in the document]
kentcdodds: Also, we should probably do:

```
const g = typeof window === 'undefined' ? global : window
```

And then reference `g` instead. That way this runs in the browser properly as well.

src/__tests__/events.js
jomaxx: makes sense, i can open a PR to `react-testing-library` as well. there are a couple events we'll need to fallback to triggering synthetic events for (`change`, `select`, `mouseenter`, `mouseleave`) so I'll add that logic there

kentcdodds: Perfect. Thanks a bunch @jomaxx. This is so great!

sompylasar:
> there are a couple events we'll need to fallback to triggering synthetic events for (change, select, mouseenter, mouseleave) so I'll add that logic there

@jomaxx Could you please clarify about synthetic events? How does Cypress handles these events from within the browser?

jomaxx:
@sompylasar let's move discussion about synthetic events to https://github.com/kentcdodds/react-testing-library/pull/48

kentcdodds:

PR #21, opened 2018-04-11
Ignore extra whitespace in node texts

gnapse:
As discussed in https://github.com/gnapse/jest-dom/issues/9#issuecomment-380500371


src/queries.js
```diff
@@ -173,6 +171,7 @@ export {
   queryByPlaceholderText,
   getByPlaceholderText,
   queryByText,
+  getText,
```
kentcdodds: This file should _only_ expose queries. Other utilities should live in other files. Could you move `getText` to its own file?

gnapse: Got it. Also, I wonder if we should publicly call it `getNodeText` instead?

kentcdodds: Agreed. We can rename it to `getNodeText` :+1:

README.md
```diff
@@ -395,6 +395,25 @@ fireEvent.click(getElementByText('Submit'), rightClick)
...
+whitespace. The intention is to treat text in nodes exactly as how it is
+perceived by users in a browser, where any extra whitespace within words in the
+html code is not meaningful when the text is rendered.
+
+```javascript
+// <div>
+//   Hello
+//     World  !
+// </div>
+const text = getNodeText(container.querySelector('div')) // "Hello World!"
```
kentcdodds: Shouldn't this be: "Hello World !" ?

gnapse:
This looks unrelated, and perhaps there's a brittle test in the wait-for-element part. I just ran `npm run test:update` locally and it all passes with full coverage. Plus a change to the README breaking a test makes no sense.

gnapse:
BTW (unrelated but) have you considered adding `npm run test:update` to be run in a pre-push hook in your projects?

kentcdodds:
I've restarted the build. Yes, that's odd. Hmmm...

> BTW (unrelated but) have you considered adding npm run test:update to be run in a pre-push hook in your projects?

I have not! Hmmm... I wonder if I could make it interactive. If there are snapshot differences, ask the user if they'd like to cancel the commit or update them... Not sure that'd be possible with lint-staged though...

gnapse:
The restarted build passed! 😱

kentcdodds:
Uh oh... We should probably look into that.

Issue #24, opened 2018-04-19
Add a function for printing a simple DOM representation for debugging


jlongster:
Would you be interested in a function like this? The only annoying thing I've found with using jsdom is that it's hard to debug if you don't know what's going on or are confused. In the browser you can just look at it, but you can't with jsdom. This gives me most of what I want:

```js
function prettyDOM(node, indent = 0) {
  let str = '';
  if (node) {
    str += ' '.repeat(indent) + (node.tagName || node.textContent) + '\n';
    for (let child of node.childNodes) {
      str += debugDOM(child, indent + 2);
    }
  }
  return str;
}
```

You can now do `console.log(prettyDOM(node))` and get an output like this in the terminal:

<img width="695" alt="screen shot 2018-04-18 at 9 36 57 pm" src="https://user-images.githubusercontent.com/17031/38966517-ae48e320-4350-11e8-917a-dca4feb9c7cb.png">

jlongster:
Even better, add the attached testid to the output (and maybe allow the user to pass in specific attributes they want printed?):

```js
function debugDOM(node, indent = 0) {
  let str = '';
  if (node) {
    str += ' '.repeat(indent);
    if (node.tagName) {
      str += node.tagName + ' ' + (node.dataset['testid'] || '') + '\n';
    } else {
      str += node.textContent + '\n';
    }

    for (let child of node.childNodes) {
      str += debugDOM(child, indent + 2);
    }
  }
  return str;
}
```

<img width="587" alt="screen shot 2018-04-18 at 9 40 42 pm" src="https://user-images.githubusercontent.com/17031/38966614-2fce5614-4351-11e8-8a65-ccdedcdacadd.png">

kentcdodds:
Perhaps we could expose this:

https://github.com/kentcdodds/dom-testing-library/blob/131a20bcc7cca4c0853acc49365a6c4746165471/src/queries.js#L150-L160

![dab5me1v4aaehlr](https://user-images.githubusercontent.com/1500684/38967293-f9e5a610-434b-11e8-835a-b2d81b0fd831.jpg)

What do you think?

jlongster:
Yeah, that's probably good! I like seeing a more concise form, is there a way to customize it at all? I basically don't care about most of my attributes (and there is quite a bit in my dom, which makes it harder to read). It's not a big deal though, I'd be fine just using that function.

gnapse:
[...]

PR #25, opened 2018-04-19
Expose pretty-dom utility

gnapse:
**What**:

This exposes the already existing internal utility function to print out a DOM in a human-friendly way.

**Why**:

As it was discussed in #24, this can be useful for users to print out the DOM under a certain html element, to debug more easily when writing tests.

**How**:

Simply moved the already existing utility to its own module.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Code coverage

Code coverage is not full now because of the `logDOM` utility function. Now in hindsight I'm not sure if it's a good idea to provide it. This is because there's no eslint rule now warning the users that they left this command in use in their tests, when I think the intent is that this is only to be used while creating and changing tests, but not meant to be left in actual committed code.

If you still think we should provide it, how do you think we should resolve the code coverage issue? Should we create a new test that uses it? Wouldn't that pollute the tests output? Is there a way to ignore some lines in the coverage report?


src/pretty-dom.js
```diff
@@ -0,0 +1,22 @@
...
+
+const {DOMElement, DOMCollection} = prettyFormat.plugins
+
+function prettyDOM(htmlElement) {
+  const debugContent = prettyFormat(htmlElement, {
+    plugins: [DOMElement, DOMCollection],
+    printFunctionName: false,
+    highlight: true,
+  })
+  const maxLength = process.env.DEBUG_PRINT_LIMIT || 7000
```
kentcdodds: Perhaps we should accept this as an argument? I'm thinking that folks who are logging things themselves might want to control this more easily and in general probably want to see the whole thing. So perhaps we should only have this default for where it was already in use and the real default should be Infinity?

kentcdodds:
As for the coverage thing:

```javascript
jest.spyOn(console, 'log').mockImplementation(() => {})
logDOM(document.createElement('div'))
expect(console.log.mock.calls).toMatchSnapshot()
console.log.mockRestore()
```

:smile:

gnapse:
Yes, I was aware that documentation is pending but wanted to kick off the discussion of code here. I'll change the approach to the limit. I also wanted to discuss the code coverage issue plus the potential linting issue with `logDOM` that I mentioned.

gnapse:
@kentcdodds what do you think about not providing `logDOM`? It's not about the coverage, but about the fact that it'd be effectively hiding a `console.log` from the user's eslint. A `console.log` that can potentially end up being committed into their vcs.

As a reference, I think I recall that enzyme's pretty print helper also did not do the `console.log` itself, it just gives you the output to for the user to log it out or whatever. It's also yet another API to document and maintain, when all it does it very simple.

src/pretty-dom.js
antoaravinth: Yes I agree with @kentcdodds. Currently when a test case is failing, its chipping of the container and I need to increase the `DEBUG_PRINT_LIMIT` to see the whole dom, which is causing me to re-run the case again.

gnapse: I changed it, and now the env var is only used internally when pretty printing out DOM inside thrown exceptions by the element getters. The exposed `prettyDOM` does not limit the output at all, and allows a second optional argument for the user to set a limit per call.

kentcdodds:
I'd be fine with not including `logDOM` :+1:

src/pretty-dom.js
```diff
@@ -0,0 +1,16 @@
+import prettyFormat from 'pretty-format'
+
+const {DOMElement, DOMCollection} = prettyFormat.plugins
+
+function prettyDOM(htmlElement, maxLength = undefined) {
```
kentcdodds: Don't need the `= undefined` because that's functionally equivalent to what the value of `maxLength` would be anyway :)

gnapse: I know, I just like to be more explicit in the declaration that the argument is not expected, to anyone that passes by and read the code. Guess flow or typescript would be clearer for that purpose but I have yet to learn to use one of them, it's on my list ;)

jlongster:
Wow, that was fast! Thanks all!

Issue #27, opened 2018-04-20
getAllBy* matchers?


alexkrolick:
* `dom-testing-library` version: 2.1.1

It looks like all the matchers are designed to return the first element matching a query. Would it make sense to also export matchers that return all of the matching nodes as well (e.g., `document.querySelectorAll` vs `document.querySelector`)? Some use cases are testing the length or sort order of lists, handling pagination, etc.

kentcdodds:
Yes, I can see the value of this and done have asked me about it, so sure, I'd accept a pull request for it 👍

PR #28, opened 2018-04-22
Add queryAll/getAll methods

alexkrolick:
**What**:

Adds "all" methods for:
- byAltText
- byTestId

**Why**:

#27 

**How**:

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 

---

**Sidenote**

If these `getAll` methods are added, the `queryBy` methods seem kind of redundant - instead of checking if `queryBy` returns `null`, you can check if `getAll` is `[]`.


alexkrolick:
> I'm thinking that it would be good to actually have an *All* version for every query/get method for consistency.

Yep, just wanted to start with a few cases for review.

> And I'm thinking that we could probably have the *All* methods be the base and the existing methods just use those and return the first element to simplify the implementation.

Yeah. Would the `queryBy` methods continue to be supported (ie `if (result.length === 0) return null`) or phased out?

> Also, we'll want to make sure to document the existence of the *All* methods.

Yep.

kentcdodds:
Yes, let's do this with no breaking changes.

Yes, please maintain 100% code coverage 👍

alexkrolick:
OK, 🍏 

- Added `queryAll` methods for everything
- Reimplemented `query` APIs using `queryAll`
- Aliased `getAll` to `queryAll` (they are the same)
- Added tests for the array methods; most of the code is actually executed in tests for the `get` methods.

alexkrolick:
✅ Changed `getAll` to throw the same errors as the `get` methods

src/queries.js
```diff
@@ -10,156 +10,261 @@ function debugDOM(htmlElement) {
...
 
 function getByAltText(container, alt) {
   const el = queryByAltText(container, alt)
   if (!el) {
-    throw new Error(
-      `Unable to find an element with the alt text: ${alt} \n\n${debugDOM(
-        container,
-      )}`,
-    )
+    throw noMatchingAltText(container, alt)
```
kentcdodds: What if instead of creating these error functions we had the `getBy*` functions use the `getAllBy*` functions? Perhaps using `firstResultOrNull`.

Sorry, I still haven't reviewed all the PR but this is just something else I noticed.

alexkrolick: Yep that makes sense!

kentcdodds:

PR #34, opened 2018-05-05
poc for getting/querying by title

lgandecki:
**What**:

Add ability to get/query by title, basic test coverage for now

**Why**:

Here comes my question, I've been seeing this pattern quite a bit:
https://github.com/meteor/todos/blob/react/imports/ui/components/ListHeader.jsx#L162

I wanted to learn your thinking here - are those antipatterns? I'm rewriting above example app by the Meteor guys to Apollo backend/frontend, while using cypress-testing-library and react-testing-library to drive the work.

I'm not sure if I have any other way of accessing elements like that.

I could rewrite places like that to use img with alt instead, but I'm trying to keep the components as similar to the original ones as possible (so it will be easier to compare between meteor and apollo)

**How**:

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ when a real user uses it.
   * [`getByPlaceholderText(container: HTMLElement, text: TextMatch): HTMLElement`](#getbyplaceholdertextcontainer-htmlelement-text-textmatch-htmlelement)
   * [`getByText(container: HTMLElement, text: TextMatch): HTMLElement`](#getbytextcontainer-htmlelement-text-textmatch-htmlelement)
   * [`getByAltText(container: HTMLElement, text: TextMatch): HTMLElement`](#getbyalttextcontainer-htmlelement-text-textmatch-htmlelement)
+  * [`getByTitle(container: HTMLElement, title: ExactTextMatch): HTMLElement`](#getbytitlecontainer-htmlelement-title-exacttextmatch-htmlelement)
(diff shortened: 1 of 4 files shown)


lgandecki:
One more thing! In the react-testing-library I just add custom selectors as I go, some for a high level functionality (like this - getByTitle, or getByValue), some for higher level ( getDeleteTodo ), it works beautifully and seamlessly, but when I go to cypress I can't reuse them. It would be nice to expose this:
https://github.com/kentcdodds/cypress-testing-library/blob/master/src/index.js#L8
so people could add their own custom getters the way you have done there :) Right now I just copied and pasted this code to my codebase, but I'm obviously not very happy with that solution. 

Is this something you would be willing to take a PR for?

lgandecki:
Cool, thanks! Next one I needed to add manually that seems universal was getByValue. What do you think about that one?

In case of that specific rewrite I'm doing - all the listed todos are inputs with values. They get automatically updated when you click and change their value. unfortunately, getByText doesn't work in that case. Using placeholder doesn't make too much sense  - it would have to be generic ("todo") instead of the specific todo text. And, obviously, I'd rather not add testids there, since all I have is already visible from the point of view of the user :)

kentcdodds:

kentcdodds:
Yes, I suppose a getByValue also makes sense. In general, I'm in favor of selectors that resemble how users find elements on the page, or (less optimal) how screen readers find elements on a page.

PR #35, opened 2018-05-06
get/query by value

lgandecki:
**What**:

get/query by value

**Why**:

as per discussion in #34 

Note: setting .value on an element here somehow doesn't work, so i used .setAttribute. I linked it to my react project, and it worked with .value just fine. Not sure what's going on here.
I changed the selector part here to * instead of [${attribute}] to verify that that's not the cause:
```javascript
return Array.from(container.querySelectorAll(`[${attribute}]`)).filter(node =>
```
but it didn't change anything.

Another note, related to that, in my original implementation in my project I used a selector for getting all inputs, I guess if someone set an attribute value on a div, current implementation would get that element as well - not sure how much we want to prevent such edge cases. It's a decision between complicating the code a bit, and making sure we treat values only when they are actually input values.  

**How**:

Following the existing pattern

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -135,18 +135,16 @@ test('query/get element by its title', () => {
 })
 
 test('query/get element by its value', () => {
-  const {getByValue, queryByValue, getByPlaceholderText} = render(`
+  const {getByValue, queryByValue} = render(`
   <div>
-    <input placeholder="name" type="text" id="name"/>
-    <input placeholder="lastname" type="text" id="lastname"/>
-    <input placeholder="email" type="text" id="email"/>
+    <input placeholder="name" type="text"/>
+    <input placeholder="lastname" type="text" value="Norris"/>
+    <input placeholder="email" type="text"/>
   </div>
   `)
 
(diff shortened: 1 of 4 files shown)


src/__tests__/element-queries.js
```diff
@@ -132,6 +134,21 @@ test('query/get element by its title', () => {
...
+test('query/get element by its value', () => {
+  const {getByValue, queryByValue, getByPlaceholderText} = render(`
+  <div>
+    <input placeholder="name" type="text" id="name"/>
+    <input placeholder="lastname" type="text" id="lastname"/>
+    <input placeholder="email" type="text" id="email"/>
+  </div>
+  `)
+
+  getByPlaceholderText('lastname').setAttribute('value', 'Norris')
```
alexkrolick: Might be a bit more straightforward to set the `value` initially in the DOM, since the `setAttribute` part is JSDOM's responsibility. What do you think?

lgandecki: makes sense, I updated the test.

PR #37, opened 2018-05-08
Add typescript definitions

maddijoyce:
**What**: Adding typescript definitions

**Why**: The types for some parts of the dom-testing-library are currently in use in react-testing-library. Really these should be stored alongside the appropriate code and then imported as needed.

**How**: Added d.ts files with definitions to match functions. Some definitions pulled from react-testing-library (thanks @alexkrolick and @InExtremaRes).

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged 

I think adding dtslint is a good first step as far as keeping the definition files tested and avoiding types issues (as per kentcdodds/react-testing-library#82).

The types are only compatible with TypeScript 2.8 and up, because I've used mapped types for the bindElementToQueries function. It's possible to avoid using mapped types, but it means duplication of a lot of the queries code, and I think it's better to avoid that.

If you're happy with these types, I can update the react-testing-library to use them once everything has been merged.


maddijoyce:
Also fixed a small error in events.js.
The copy event was using a CompositionEvent not a ClipboardEvent

kentcdodds:

Issue #53, opened 2018-06-10
Improve the 'byText' queries capability.


Dajust:
### Describe the feature you'd like:

'byText' queires, `getByText()`, for example, works greate in writing tests that closely resembles users actions, but it has a drawback, expecially on integration tests, which usually requires many components to be rendered.

Sometimes a user might be interested in a particular text, but only on a prticular section of the page. Take the bellow screenshot for example.

![chat-screen1](https://user-images.githubusercontent.com/8015514/41205760-8a063f88-6cf1-11e8-9ce1-cd48fd179ba8.png)

So the user sends a 'lol' and  'today' message and wants to make sure the message is sent. She'll have to look for the message "byText" 'lol' to see when/if the message has been markded as sent. It is only the "byText" queries that can help us do this job, but here's the problem: There are many 'lol' and 'today' texts on the page, but our user is only looking at or interested in the 'lol' text INSIDE-THE-MESSAGES-SECTION.

`getByText` in it's current form will not help us in such case, because it will grab the text 'lol' anywhere it founds it on the page, which in turn require our tests to rely on DOM structure to get the text where're interested in.
### Suggested API:
Provide an option to tell the 'byText' queries *where-in-the-document* they should get the text. Something like this:

`getByText('some text', {in: node})`

### Suggested implementation:
I think this could be easily implemented by some minor editting the  [queryAllByText fucntion](https://github.com/kentcdodds/dom-testing-library/blob/master/src/queries.js#L88).

We could simply check to see if the `in` option is present, then use it as the baseElement of the querySelectorAll(). We only need to touch too lines in the function. Something like this:

```
function queryAllByText(
  container,
  text,
  {selector = '*', exact = true, collapseWhitespace = true, trim = true, ...options} = {}, 
) {
  const matcher = exact ? matches : fuzzyMatches
  const matchOpts = {collapseWhitespace, trim}
  // determin what the baseElement should be
  const baseElement = options.in ? options.in : container;
  return Array.from(baseElement.querySelectorAll(selector)).filter(node =>
    matcher(getNodeText(node), node, text, matchOpts),
  )
}
```

### Describe alternatives you've considered:

N/A

### Teachability, Documentation, Adoption, Migration Strategy:

To use this feature, user will need to first grab the `in` node, then use it as the `in` option. Like this:

`const messageArea = getByTestId('message-area');`
`getByText('some text', {in: messageArea})`

sompylasar:
[...]

PR #54, opened 2018-06-12
Add within API #53

Dajust:
**What**: Add a `within` API

**Why**: Kindly see #53 

**How**: Exported `getQueriesForElement` as `within` from the index.js file

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -238,6 +238,17 @@
       "contributions": [
         "example"
       ]
+    },
+    {
+      "login": "Dajust",
+      "name": "Justice Mba ",
+      "avatar_url": "https://avatars3.githubusercontent.com/u/8015514?v=4",
+      "profile": "https://github.com/Dajust",
+      "contributions": [
+        "code",
+        "doc",
+        "ideas"
+      ]
     }
   ]
 }
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 [![downloads][downloads-badge]][npmtrends]
 [![MIT License][license-badge]][license]
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-23-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-24-orange.svg?style=flat-square)](#contributors)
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -642,6 +643,22 @@ expect(submitButtons).toHaveLength(3) // expect 3 elements
...
 ```
 
+## `within` API
+
+Sometimes, there is no garauntee that the text, placeholder, or label you want to query is unique on the page. So you might want to explicity tell react-render-dom to get an element **only within** a particular section of the page. `within` is a helper function for this case.
+
+Example: To get the text 'hello' only within a section called 'messages', you could do:
+
+```javascript
+import {render, within} from 'react-testing-library'
```
kentcdodds: Could you change this example to work better in the `dom-testing-library` docs?

```js
import {within} from 'dom-testing-library'

const {getByText} = within(document.body.getElementById('messages'))
const helloMessage = getByText('hello')
```

Also, could you add a note that says:


```md
> Fun fact, this is actually just an alias to the `getQueriesForElement` function!
```

src/index.js
```diff
@@ -16,4 +16,7 @@ export * from './pretty-dom'
 
 // The original name of bindElementToQueries was weird
 // The new name is better. Remove this in the next major version bump.
```
kentcdodds: Could you move these two comments to right above the `bindElementsToQueries` export? That's where the comment applies. We'll keep the `within` alias indefinitely.

.all-contributorsrc
```diff
@@ -238,6 +238,17 @@
       "contributions": [
         "example"
       ]
+    },
+    {
+      "login": "Dajust",
+      "name": "Justice Mba ",
```
sompylasar: There's an extra space after your name, typo?

Dajust: Not sure, I think that file is auto-generated.

kentcdodds: You can fix it manually.

You may want to fix it on GitHub too though :wink:

screen shot 2018-06-12 at 11 59 51 am

PR #58, opened 2018-06-15
Merge waitForElement into wait

hnordt:
**What**:

We are trying to merge `waitForElement` into `wait`.

See https://github.com/kentcdodds/dom-testing-library/issues/57 for details.

**Why**:

To simplify the API.

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [ ] Ready to be merged 
* [ ] Added myself to contributors table

### DIFF
--- a/src/__tests__/wait-for-element.js
+++ b/src/__tests__/wait-for-element.js
@@ -1,4 +1,4 @@
-import {waitForElement, wait} from '../'
+import {wait} from '../'
 // adds special assertions like toBeInTheDOM
 import 'jest-dom/extend-expect'
 import {render} from './helpers/test-utils'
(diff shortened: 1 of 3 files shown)


hnordt:
@kentcdodds 

Actually it's gonna be more complicated than I thought. We can't rely on `container` cause then the user would need to pass `container` every time. The workaround would be returning `wait` from `render()`.

If we use `isElement()`, `callback(() => null)` would run `wait()` and not `waitForElement()` so we need to refactor `waitForElement`.

lgandecki:
don't know how to solve the problem - but I do love the idea of merging those two together:)

kentcdodds:
> We can't rely on container cause then the user would need to pass container every time.

Actually I don't think so. Having the `container` default to `document` will work 99% of the time. For the rest of the time people can pass a custom `container` and I think that's fine.

hnordt:
@kentcdodds I agree with you.

Unfortunately we don't have a reliable way to detect if the user is trying to wait for an element, we really need that information from the user itself.

PR #67, opened 2018-07-09
feat(queries): allow overriding queries in getQueriesForElement

alexkrolick:
**What**:

**Why**:

**How**:

**Checklist**:

* [ ] Documentation
* [x] Tests
* [x] Ready to be merged 

### DIFF
--- a/src/get-queries-for-element.js
+++ b/src/get-queries-for-element.js
@@ -6,17 +6,11 @@ import * as defaultQueries from './queries'
 
 /**
  * @param {HTMLElement} element container
- * @param {FuncMap|FuncMap[]} queries object of functions, or array of objects
+ * @param {FuncMap} queries object of functions
  * @returns {FuncMap} returns object of functions bound to container
  */
 function getQueriesForElement(element, queries = defaultQueries) {
-  let flattenedQueries
-  if (Array.isArray(queries)) {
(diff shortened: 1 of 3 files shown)


src/get-queries-for-element.js
```diff
@@ -1,7 +1,22 @@
...
+/**
+ * @typedef {{[key: string]: Function}} FuncMap
+ */
+
+/**
+ * @param {HTMLElement} element container
+ * @param {FuncMap|FuncMap[]} queries object of functions, or array of objects
+ * @returns {FuncMap} returns object of functions bound to container
+ */
+function getQueriesForElement(element, queries = defaultQueries) {
```
alexkrolick: Note that passing custom queries _overrides_ the default instead of merging. If you want to merge you need to import `queries` and add it to the array, too.

Open to changing this but I think it makes sense; wrappers like react-testing-library's `render` can merge queries by default on top of this API if they prefer.

src/get-queries-for-element.js
```diff
@@ -1,7 +1,22 @@
...
+  let flattenedQueries
+  if (Array.isArray(queries)) {
```
kentcdodds: I'd rather not have this logic here. If someone wants to merge the built in queries they can be responsible for doing that themselves when they call the function.

PR #77, opened 2018-07-19
fix: Reference Node and MutationObserver off of the window object

dadamssg:
fix: Reference Node and MutationObserver off of the window object rather than globally

You have to handle shims yourself if not using jest without this

**What**:
Changes references to `Node` and `MutationObserver` to use `window` rather than globally.

**Why**:
The polyfills/shims do not work in a non-jest environment

**How**:
`Node` -> `window.Node`
`MutationObserver` -> `window.MutationObserver`

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged 
* [ ] Added myself to contributors table

### DIFF
--- a/src/get-node-text.js
+++ b/src/get-node-text.js
@@ -1,7 +1,8 @@
 function getNodeText(node) {
   return Array.from(node.childNodes)
     .filter(
-      child => child.nodeType === Node.TEXT_NODE && Boolean(child.textContent),
+      child =>
+        child.nodeType === window.Node.TEXT_NODE && Boolean(child.textContent),
     )
     .map(c => c.textContent)
     .join(' ')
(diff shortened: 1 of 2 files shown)


kentcdodds:
Would you like to [add yourself](../blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor) to the [contributors table](../#contributors)?

dadamssg:
Sure! I felt awkward doing that with such a small change. Should I make another commit on this branch? or create a new one?

kentcdodds:
I recommend creating a new branch for every PR.

You may also appreciate this: http://makeapullrequest.com/ :)

dadamssg:
I definitely appreciate that :)

PR #85, opened 2018-08-04
feat(fireEvent): helper to assign target properties to node

kentcdodds:
**What**: feat(fireEvent): helper to assign target properties to node

**Why**: Specifically this is so the change event can set the value of the node
in a way that works for React (and all other frameworks and
non-frameworks) nicely.

fixes: https://github.com/kentcdodds/react-testing-library/issues/152

**How**: Add some logic to the `fireEvent` utility, and borrow some code from a thread in react issues.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 
* [ ] Added myself to contributors table N/A

### DIFF
--- a/src/__tests__/events.js
+++ b/src/__tests__/events.js
@@ -1,4 +1,4 @@
-import {fireEvent} from '../'
+import {fireEvent} from '..'
 
 const eventTypes = [
   {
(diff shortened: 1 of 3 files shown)


kentcdodds:
I'd love feedback on this. This change will mean we can remove `change` from [this list](https://github.com/kentcdodds/react-testing-library/blob/3fb77d3f78a8f0c09f6cd60fb14008a90ad1a1b9/src/index.js#L51) which I'm trying to slowly wittle down to nothing.

src/__tests__/events.js
```diff
@@ -149,3 +149,22 @@ describe(`Aliased Events`, () => {
...
+  expect(node.value).toBe(value)
+})
+
+test('assigning a value to a target that cannot have a value throws an error', () => {
+  const node = document.createElement('div')
+  expect(() =>
+    fireEvent.change(node, {target: {value: 'a'}}),
+  ).toThrowErrorMatchingInlineSnapshot(
+    `"The given element does not have a value setter"`,
+  )
```
kentcdodds: I just gotta say, this inline snapshot business is just the best.

kentcdodds:
Ok, I'll go ahead and self-approve this PR. I hope I don't regret this! 😅

kentcdodds:
Thanks @jomaxx!

PR #86, opened 2018-08-09
Add ability to select `title` element of SVG with `getByTitle`

### LINKED ISSUE #83: Add support for `<title>` element in SVG with `getByTitle`

There is a `<title>` element that can be rendered within an SVG element, see [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title), and would be nice to be considered when doing a `getByTitle` assertion. Currently it does work with `getByText`, but that might be incorrect considering it behaves more like a title attribute than text content.

If accepted, I will work up a PR adding support for it.

### Suggested implementation:

Would need to find the `<title>` element and ensure it's parent is a `<svg>` element and allow that to be included in the assertion.

We currently use `getByText` as a selector.

wyze:
Close #83

**What**: Add ability to select `title` element of SVG with `getByTitle`

**Why**: The `title` element of an SVG behaves the same as a `title` attribute on other elements, so the `getByTitle` query should find that element.

**How**: Update `queryAllByTitle` function to check for `[title]` or `svg > title` selectors and use either one that is returned for the match.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


src/queries.js
```diff
@@ -82,12 +82,28 @@ function queryByText(...args) {
   return firstResultOrNull(queryAllByText, ...args)
 }
 
+function queryAllByTitle(
+  container,
+  text,
+  {exact = true, collapseWhitespace = true, trim = true} = {},
+) {
+  const matcher = exact ? matches : fuzzyMatches
```
kentcdodds: I'm guessing this is where we're missing some coverage. Could you add one more assertion that specifies `exact: false` to cover this?

wyze: Sure thing. Will get that pushed up shortly.

kentcdodds:

PR #87, opened 2018-08-10
feat(getNodeText): Don't join text nodes with whitespace

bdchauvette:
**What**:

Changes the way `getNodeText` joins nodes. Instead of joining them with a space, it now joins them with an empty string.

**Why**:

This PR comes out of https://github.com/kentcdodds/react-testing-library/issues/53

In React, if you have an element like `<span>{volts}V</span>`, then the generated element will have two text nodes, one for the variable `volts`, and one for the normal string.

When browsers render this, there is no space between the text nodes. Likewise, the `textContent` property on the span will return a string without spaces (e.g. `300V`).

However, the current implementation of `getNodeText` joins all of the nodes with a space, so if you try to use e.g. `queryByText('300V')` it will return `null`.

For a live demo, see https://codesandbox.io/s/w718n5ojq7

**How**:

- changing `join(' ')` to `join('')` on [line 8 of `src/get-node-text.js`](https://github.com/kentcdodds/dom-testing-library/compare/master...bdchauvette:pr/get-node-text-whitespace?expand=1#diff-6350d468f7684d134aab9d42d96a77beR8)
- adding [a test](https://github.com/kentcdodds/dom-testing-library/compare/master...bdchauvette:pr/get-node-text-whitespace?expand=1#diff-de31c4d0bed96b2a4211de164bb1b589R59) for querying against a DOM element containing adjacent text nodes without any whitespace

**Checklist**:

- [ ] Documentation - **N/A** (I think?)
- [x] Tests
- [x] Ready to be merged 

Thanks! :smile_cat:


kentcdodds:
Thanks for this! I think that I misunderstood something when I created this in the first place. I've verified that this change doesn't break anything on react-testing-library and react-testing-library-examples :+1: Thank you again!

PR #89, opened 2018-08-21
fixed small typo

shairez:
**What**:
fixed a small typo

**Why**:
I have OCD

**How**:
By the power invested in me via my keyboard

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [x] Ready to be merged 

### DIFF
--- a/README.md
+++ b/README.md
@@ -554,7 +554,7 @@ finding elements in the DOM similarly to how users would do.
 
 When using [jest][], it is convenient to import a set of custom matchers that
 make it easier to check several aspects of the state of a DOM element. For
-exmaple, you can use the ones provided by
+example, you can use the ones provided by
 [jest-dom](https://github.com/gnapse/jest-dom):
 
 ```javascript


kentcdodds:
Awesome! I'm glad! Have you seen this? https://github.com/timdeschryver/ngx-testing-library#readme

shairez:
Haven't heard about it, looks nice! 

There is also [Spectator](https://github.com/NetanelBasal/spectator) which uses your lib.
Both look similar.

I'm aiming to solve the same problem from a different angle.
Also have a different approach on how to test more "smarter" components without paying the cost of rendering and setup but without the need to shallow render (to be able to break my tests on a child component mismatch), and without going full integration.

I'm considering using this lib with the `within` function to decorate my object with its functionality. 

Currently experimenting with it, will update.

kentcdodds:
Interesting! The `within` function is definitely the intended composition mechanism for libraries that build on top of dom-testing-library. I'd love to have a *-testing-library for every framework considering they're generally fairly small and easy to write and because the philosophy is to not care about implementation details almost all frameworks can be tested in almost exactly the same way which I think is a very good thing :)

Good luck!

shairez:
Thanks brother! 
Yeah, as I said, I like the simplicity. 

Keep up the good work!

Issue #93, opened 2018-09-04
Add new getByRole queryselector.


JeffBaumgardt:
### Describe the feature you'd like:

A new querySelector to select elements based on role. I have a dialog that while I can query inside the dialog just fine I cannot first a click to close the dialog because all the current queries target elements inside the dialog so the out of bounds click handlers do not fire.

### Suggested implementation:

const queryAllByRole = queryAllByAttribute.bind(null, 'role')

```js
const queryByRole = queryByAttribute.bind(null, 'role')
const queryAllByRole = queryAllByAttribute.bind(null, 'role')

function getAllByRole(container, id, ...rest) {
  const els = queryAllByRole(container, id, ...rest)
  if (!els.length) {
    throw getElementError(
      `Unable to find an element by role=${id}`,
      container,
    )
  }
  return els
}

function getByRole(...args) {
  return firstResultOrNull(getAllByRole, ...args)
}
```

### Describe alternatives you've considered:

I'm using Material UI where there is a lot of abstraction in DOM that I can't put data-testid on the outer dialog. So targeting it by testid won't work for triggering the onClose event.

The only thing I've considered is writing the
``const dialogConstainer = container.querySelector('[role="dialog"]')``
And then writing assertions from there.

### Teachability, Documentation, Adoption, Migration Strategy:

This is a non breaking change as it is only an addition to the existing query set.

---------

### `getByRole`

```typescript
getByRole(
  container: HTMLElement,
  text: TextMatch,
  options?: {
    exact?: boolean = true,
    collapseWhitespace?: boolean = false,
    trim?: boolean = true,
  }): HTMLElement`
```

A shortcut to `` container.querySelector(`[role="${yourRole}"]`) `` (and it
also accepts a [`TextMatch`](#textmatch)).

```javascript
// <div role="dialog">...</div>
const dialogContainer = getByTestrole(container, 'dialog')
```

gnapse:
Funny, I was thinking something similar now after the recent discussion in gnapse/jest-dom#55. It would be a nice addition.

kentcdodds:
Thanks for the thorough feature request @JeffBaumgardt! I'm in favor of this. Would you like to implement it?

JeffBaumgardt:
Can do I'll have a PR later today.

JeffBaumgardt:
Closed via #94

PR #94, opened 2018-09-04
Feature - Add queries for dom attriburte `role`

JeffBaumgardt:
**What**: Add queries for role dom attribute

**Why**: As evidenced by [gnapse/jest-dom#55](https://github.com/gnapse/jest-dom/issues/55) having a role attribute query would be helpful when attempting to get elements that don't fit the other queries.

My example was a dialog who did not have any other matching queries and due to the associated library. I was able to add data-testid but that added the attribute too low in the generated dom not to the container root as expected, this is a problem with the library not dom-testing-library.

**How**: I added new query selectors for the node attribute.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 


JeffBaumgardt:
This resolves #93

README.md
```diff
@@ -380,6 +380,27 @@ const usernameInputElement = getByTestId(container, 'username-input')
 > `data-testid`s from the blog post
 > ["Making your UI tests resilient to change"][data-testid-blog-post]
 
+### `getByRole`
```
kentcdodds: Let's put this above `getByTestId`. I want the docs to read from most recommended to least. I would recommend `getByRole` over `getByTestId` when it applies.

src/__tests__/element-queries.js
```diff
@@ -375,6 +383,22 @@ test('using jest helpers to check element class names', () => {
...
 })
 
+test('using jest helpers to check element role', () => {
+  const {getByRole} = render(`
+    <div role="dialog">
+      <span>Contents</span>
+    </div>
+  `)
+
+  expect(getByRole('dialog')).toBeTruthy()
```
kentcdodds: This assertion is unnecessary because the next assertion exists :)

src/__tests__/element-queries.js
```diff
@@ -375,6 +383,22 @@ test('using jest helpers to check element class names', () => {
...
+  expect(getByRole('dialog')).toHaveTextContent('Contents')
+
+  expect(() => expect(getByRole('dialog2')).not.toBeTruthy()).toThrowError()
```
kentcdodds: These other assertions are unnecessary because they're basically how jest-dom works and we don't need to worry about testing jest-dom :)

kentcdodds:
Just noticed that the commit is attributed to `jbaumgardt` (probably due to your local git config). Just thought I'd mention it in case that's a mistake.

JeffBaumgardt:
Yeah my git config is all messed up. I need to wipe it and start over :)

README.md
JeffBaumgardt: Done

src/__tests__/element-queries.js
JeffBaumgardt: Removed assertions

I basically copied from other tests so I wasn't sure how bullet proof you wanted it.

src/__tests__/element-queries.js
JeffBaumgardt: Removed assertions

I basically copied from other tests so I wasn't sure how bullet proof you wanted it.

PR #95, opened 2018-09-10
fix: upgrade wait-for-expect to 1.0.0

SimenB:
**What**:
Dep upgrade

**Why**:
I noticed I got both 0.4.0 and 1.0.0 in my lockfile when installing react-testing-library

**How**:
GitHub UI

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   "dependencies": {
     "pretty-format": "^22.4.3",
     "mutationobserver-shim": "^0.3.2",
-    "wait-for-expect": "^0.4.0"
+    "wait-for-expect": "^1.0.0"
   },
   "devDependencies": {
     "dtslint": "^0.3.0",


SimenB:
If you rerun the build, it's gone, IIRC. You retrigger that specific build, you don't generate a new one

kentcdodds:
Are there breaking changes in the library?

lgandecki:
Good question. If it works it works by passing the tests, it works - the only change is in the way we export the function (so you no longer have to use .default in some cases).
Since we import (not require) it doesn’t matter.
 It is breaking - hence the change of the version, but not in our (dom-testing-library) case.

kentcdodds:
Super 👍 thanks friends!

PR #96, opened 2018-09-12
New: add async findBy queries

alexkrolick:
**What**:

Add async queries. I'm prefixing them with "findBy" for now.

All the `find` queries return a Promise and retry automatically using polling.

```js
const { findByText, findByTitle } = render(<Comp />)

let btn = await findByText('click me')
// something async happens...
expect(await findByTitle('click count')).toHaveTextContent('Clicked 1 time')
```

There is also a utility function called `waitFor` exposed that is similar to wait/waitForElement but curried:

```js
import { waitFor } from 'dom-testing-library'
const { getByText, container } = render(<Comp />)

const waitForLink = waitFor(href => container.querySelector(`[href=${href}]`))

let btn = await waitFor(getByText, 'click me')

expect(await waitForLink('/home')).not.toBeNull()
```

Note: there are *not* any `findAll` queries yet, though could easily be added by wrapping the `getAll` functions. I think the results may be confusing in an async environment if elements are loading at different times.

**Why**:

This simplifies working with async events, which are very common in environments like Cypress or Puppeteer.

**How**:

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,8 +77,10 @@ when a real user uses it.
   - [`getByAltText`](#getbyalttext)
   - [`getByTitle`](#getbytitle)
   - [`getByValue`](#getbyvalue)
+  - [`getByRole`](#getbyrole)
   - [`getByTestId`](#getbytestid)
   - [`wait`](#wait)
+  - [`waitFor`](#waitfor)
   - [`waitForElement`](#waitforelement)
   - [`fireEvent`](#fireevent)
 - [Custom Jest Matchers](#custom-jest-matchers)
@@ -86,6 +88,7 @@ when a real user uses it.
 - [`TextMatch`](#textmatch)
   - [Precision](#precision)
   - [TextMatch Examples](#textmatch-examples)
+- [`find` APIs](#find-apis)
 - [`query` APIs](#query-apis)
 - [`queryAll` and `getAll` APIs](#queryall-and-getall-apis)
 - [`within` and `getQueriesForElement` APIs](#within-and-getqueriesforelement-apis)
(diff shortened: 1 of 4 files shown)


kentcdodds:
Hey Alex, I really think this would be better as a separate package. The puppeteer use case makes sense I suppose, but it's not a common enough use-case to make this baked into the library. Is there any way you could use the currently exposed utilities to build a separate library?

Also, I think that your `waitFor` could actually be the `wait` function if I'm not mistaken.

alexkrolick:
I published it here: https://github.com/alexkrolick/dom-testing-addon-async

kentcdodds:
Thanks! Let's add a reference to this in the README!

kentcdodds:
What if instead of making the queries async we make the events async?

https://github.com/kentcdodds/dom-testing-library-with-anything/blob/master/fire-event-async.js

I think I'd be willing to bake this into dom-testing-library...

kentcdodds:
```javascript
import {fireEventsAsync} from 'dom-testing-library'

// ...
await fireEventsAsync.click()
getByText('stuff')
// ...
```

alexkrolick:
That would help if you are using a lot of mocked APIs with no artificial timing, which definitely covers a lot of use cases (not exactly the one I was thinking of though).

I'm trying to think of all the scenarios where a test needs to handle asynchronicity:

1. Async actions dispatched after mount/render
1. User actions trigger async handlers
1. Scheduled events (setTimeout/setInterval)
1. Deferred/partial loading (e.g. Apollo @defer directive for GraphQL queries)
1. Animations/transitions
1. Framework asynchronicity - e.g. async React and Suspense (this is unexplored territory)

In many tests on-mount callbacks and user interaction handlers are mocked by functions that return resolved promises, so waiting a tick after fireEvent (_or render_) as you propose should be enough to make subsequent DOM queries work as expected.

Some drawbacks:
- APIs that aren't mocked like that put the test writer in the same bucket as the other scenarios above where they have to fall back to the awkward waitForElement API or a hardcoded wait
- As a test writer you have to know whether you'll need an async event dispatcher or render based on the internals of the component, same as using strategically placed waits today

kentcdodds:
Right, for me that's an acceptable trade-off. Also, I don't personally think that `waitForElement` is an awkward API.

99% of my tests will have all async things mocked out so they resolve in the next tick anyway. All other tests will be full e2e and will use Cypress with cypress-testing-library, so I'm pretty ok with this 👍

kentcdodds:
So I'm planning on adding fireEventAsync 👍

I'm open to anyone else adding it first though 😃

PR #98, opened 2018-09-13
Proposal: Monorepo

alexkrolick:
_This is just a proposal_ - browse the branch here to see what it would look like https://github.com/alexkrolick/dom-testing-library/tree/monorepo

**What**:

Combine dom-testing-library with react-testing-library and other libraries in a monorepo

**Why**:

- Synchronizing version bumps and docs across the related repos is no fun
- Makes it easier to add addon packages that aren't part of the core distribution but are maintained by the community, and use a shared CI system

**How**:

Using lerna

(Note usage of `lerna import` to bring in Git history for `react-testing-library`)


kentcdodds:
Is it possible to still have the release process automated with this approach?

kentcdodds:
Hmmmm... Last commit was a year ago. Some of my favorite semantic-release features were worked on in the last year. But let's see...

alexkrolick:
FYI - there are a few things not working yet. 
- `lerna run validate` isn't able to run ESLint correctly in the package folders for some reason. I am not sure if kcd-scripts is making assumptions about the node_modules folder structure or if it's some other issue.
- Need to update all the contribution links, merge the contributor information, etc.

If you are interested in pursuing this we can come up with a plan on how to do it.

kentcdodds:
I honestly feel like we're getting pretty close to this suite of projects being "finished" meaning there wont be much need to make many changes. With that in mind I think making this a monorepo would be more work than it would be worth.

Thanks, but I think I want to pass on this.

PR #105, opened 2018-09-18
feat(events): add the capability to await events

kentcdodds:
**What**: add the capability to await events

```javascript
// currently
fireEvent.click(button)
// wait for next tick:
await wait()
// assert something that happens on the next tick

// with this change:
await fireEvent.click(button)
// assert something that happens on the next tick
```

**Why**: @alexkrolick has brought it up numerous times and some frameworks render on the next-tick rather then synchronously like React does. Allowing users to await events will significantly simplify those scenarios.

**How**: I had considered exposing a new object called `fireEventAsync`, but I think this is even better.

Tradeoffs:

1. `dispatchEvent` is _not_ async. "The return value is false if event is cancelable and at least one of the event handlers which handled this event called Event.preventDefault(). Otherwise it returns true." ([mdn](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent)). With this change, people wont be able to get the return value until the next tick. I do not consider this a very common use-case. If it comes up that people want a synchronous version of this, then perhaps we can discuss something, but I doubt it.
2. The `wait` call happens regardless of whether people are using `await`. I don't consider this a problem because it's completely unobservable and there are no side-effects.

Bottom line: People can continue doing things as they are, but if they want to wait for the next tick of the event loop (to wait for a mocked HTTP call to resolve for example), then they can simply add `await` and it'll work fine.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged 

cc @sompylasar, I'd like to hear your feedback on this.

### DIFF
--- a/src/__tests__/events.js
+++ b/src/__tests__/events.js
@@ -1,4 +1,4 @@
-import {fireEvent} from '..'
+import {fireEvent, wait} from '..'
 
 const eventTypes = [
   {
@@ -160,11 +160,11 @@ test('assigns target properties', () => {
   expect(node.value).toBe(value)
 })
 
-test('assigning a value to a target that cannot have a value throws an error', () => {
+test('assigning a value to a target that cannot have a value throws an error', async () => {
   const node = document.createElement('div')
-  expect(() =>
+  await expect(
     fireEvent.change(node, {target: {value: 'a'}}),
-  ).toThrowErrorMatchingInlineSnapshot(
+  ).rejects.toThrowErrorMatchingInlineSnapshot(
     `"The given element does not have a value setter"`,
   )
 })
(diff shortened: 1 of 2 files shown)


alexkrolick:
Thanks for this :)

If you don't `await` the fireEvent, could it cause an unhandled Promise rejection?

kentcdodds:
Nope, no unhandled promise rejection.

kentcdodds:
Ok, so I've given this a lot of thought and I can't shake off the feeling illustrated in @suchipi's [response](https://twitter.com/suchipi/status/1042142302603313152) to this:

> This kinda feels like it's conflating two different things. It could also confuse users who don't see their changes after waiting a tick.
> 
> Part of what I like about *-testing-library is that it's candid. This feels a bit too much like it's papering over details, in my opinion.

I agree that it's conflating two things. When I see the code:

```javascript
await fireEvent.click...
```

I get a feeling that I'm waiting for the click event to finish, but that's not what's happening. It's the side-effects of the click event handlers that I'm waiting for.

So as cool as this is, I think I'm going to close it for now. Maybe I can be convinced otherwise in the future. But I think for now I'm going to close this and people will have to continue doing `await wait()` which I think is acceptable and altogether better.

alexkrolick:
Yeah... having thought about async React more (not just async mocked handlers), I'm not sure just 1 tick timeout will be enough anyways.

https://twitter.com/alexmkrolick/status/1041512541719941120

PR #106, opened 2018-09-19
 feat(queryBySelectText): add get/query by select text

### LINKED ISSUE #104: Add a way to access <select> elements given their text value
**Accessing `<select>` elements by the selected option's text**
In my experience using dom-testing-library, I've had no issues accessing and interacting with DOM elements in the ways that a user would, except for `<select>` elements. Unless a label is associated with `<select>` elements, they are a headache to access, and even harder to do so in a way the software is used.

I would like to get a conversation started for how we can better support `<select>` elements moving forward.

Here's an example of a select element that would be difficult to access and use:
```html
<select>
	<option value="">State</option>
	<option value="AL">Alabama</option>
	<option value="AK">Alaska</option>
	<option value="AZ">Arizona</option>
	<option value="AR">Arkansas</option>
</select>
```

The first `<option>` is commonly used in `<select>` elements, like a placeholder attribute is used for text inputs. What the user sees is the text value of the first option, but it's not possible to access the `<select>` element given that text value. 

What I have been doing on my projects has been to create a `getSelectByText()` function, though I would hate to have specific, one-use functions or options added to dom-testing-library. 

Maybe a user should be able to select a `<select>` element by using `getByText()` using the *selector* option. I can't think of any really elegant solutions, but would be curious to hear from other users of the project.

For the time being, I've been using this to access `<select>` values based on the innerText of the selected option:
```js
function getSelectByText(text) {
  const selectElems = document.querySelectorAll('select')
  for (let i = 0; i < selectElems.length; i++) {
    const selectInstance = selectElems[i]
    const selectedText = selectInstance.selectedOptions[0].innerText
    if (selectedText === text) return selectInstance
  }
  throw new Error(`Unable to find a select element with the value: ${text}`)
}
```

matchai:
**What**: 
Add methods to access a `<select>` element by its selected option's text value. 

**Why**:
The first `<option>` is commonly used in `<select>` elements, like a placeholder attribute is used for text inputs. What the user sees is the text value of the first option, but it's not possible to access the `<select>` element given that text value. This PR provides methods to access the `<select>` element in that situation. Closes #104.

**How**:
`queryBySelectText()`, `queryAllBySelectText()`, `getBySelectText()`, and `getAllBySelectText()` were implemented.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

Thank you for making an excellent testing library. Happy to be contributing! 😄

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -8,6 +8,7 @@ beforeEach(() => {
 test('query can return null', () => {
   const {
     queryByLabelText,
+    queryBySelectText,
     queryByPlaceholderText,
     queryByText,
     queryByTestId,
@@ -15,6 +16,7 @@ test('query can return null', () => {
   } = render('<div />')
   expect(queryByTestId('LucyRicardo')).toBeNull()
   expect(queryByLabelText('LucyRicardo')).toBeNull()
+  expect(queryBySelectText('LucyRicardo')).toBeNull()
   expect(queryByPlaceholderText('LucyRicardo')).toBeNull()
   expect(queryByText('LucyRicardo')).toBeNull()
   expect(queryByAltText('LucyRicardo')).toBeNull()
@@ -23,6 +25,7 @@ test('query can return null', () => {
 test('get throws a useful error message', () => {
   const {
     getByLabelText,
+    getBySelectText,
     getByPlaceholderText,
     getByText,
     getByTestId,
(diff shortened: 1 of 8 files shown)


src/__tests__/element-queries.js
```diff
@@ -229,13 +276,25 @@ test('getAll* matchers return an array', () => {
...
       <input id="username" placeholder="Dwayne 'The Rock' Johnson" />
+      <select>
+        <option value="">German cars</option>
+        <option value="volvo">BMW</option>
+        <option value="audi">Audi</option>
+      </select>
+      <select>
+        <option value="">Japanese cars</option>
+        <option value="volvo">Toyota</option>
+        <option value="audi">Honda</option>
```
sompylasar: Fun: Choose a BMW or a Toyota, get a Volvo. Choose a Honda, get an Audi. 😁

PR #109, opened 2018-10-02
feat(getByText): add ignore option which defaults to `'script'`

kentcdodds:
**What**: add `ignore` option to `*ByText` which defaults to `'script'`

**Why**:

In our app at PayPal we're testing with Cypress and our localization content is set in a script tag on the page when it's rendered by the server. When using `getByText` with `cypress-testing-library`, we were getting that script tag before the app had a chance to render.

I'm pretty confident that 99.999% of the time people wont want to get a script tag with `getByText`, so the default makes sense to me.

**How**:

Added another layer of filtering to `queryAllByText` to filter nodes which return true for `.matches(ignore)`. Feel pretty flexible. Happy that DOM API exists!

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/src/queries.js
+++ b/src/queries.js
@@ -69,13 +69,19 @@ function queryByLabelText(...args) {
 function queryAllByText(
   container,
   text,
-  {selector = '*', exact = true, collapseWhitespace = true, trim = true} = {},
+  {
+    selector = '*',
+    exact = true,
+    collapseWhitespace = true,
+    trim = true,
+    ignore = 'script',
+  } = {},
 ) {
   const matcher = exact ? matches : fuzzyMatches
   const matchOpts = {collapseWhitespace, trim}
-  return Array.from(container.querySelectorAll(selector)).filter(node =>
-    matcher(getNodeText(node), node, text, matchOpts),
-  )
+  return Array.from(container.querySelectorAll(selector))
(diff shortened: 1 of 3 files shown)


src/queries.js
```diff
@@ -69,13 +69,19 @@ function queryByLabelText(...args) {
 function queryAllByText(
   container,
   text,
-  {selector = '*', exact = true, collapseWhitespace = true, trim = true} = {},
+  {
+    selector = '*',
+    exact = true,
+    collapseWhitespace = true,
+    trim = true,
+    ignore = 'script',
```
sompylasar: I think script _and style_ would be a better default.

kentcdodds:
More explanation here: https://www.youtube.com/watch?v=LYB0SEYaBlE

src/queries.js
kentcdodds: Good idea. I'll add that!

sompylasar: But I'm just a couple minutes late...

kentcdodds: No worries. I just pushed an update directly to master. I figured the overhead would be unnecessary.

sompylasar: Great :+1:

PR #114, opened 2018-10-09
Allow running in node environment

### LINKED ISSUE #100: Doesn't work with custom jsdom
- `dom-testing-library` version: 3.5.1
- `react` version: 16.4.1
- `node` version: 10.4.1
- `npm` (or `yarn`) version: yarn 1.7.0

### Relevant code or config:

Usual installation,

```
"testEnvironment": "node",
```
as jest configuration, then something like following in tests:

```js
import { getByText } from 'dom-testing-library'

it('Suggests cities to search by', async () => {
  const container = createJSDOM(...)
  getByText(container, 'foobar')
})
```

### What you did:

Tried to use dom-testing-library with node environment in Jest + creating own jsdom instance in a helper and passing it to dom-testing-library.

### What happened:

  ● Test suite failed to run

    ReferenceError: window is not defined

      at Object.<anonymous> (node_modules/mutationobserver-shim/dist/mutationobserver.min.js:3:1)
      at Object.<anonymous> (node_modules/dom-testing-library/dist/wait-for-element.js:8:1)
      at Object.<anonymous> (node_modules/dom-testing-library/dist/index.js:32:23)

### Reproduction:

As above, enable node jest environment and try to use dom-testing-library with some custom container

### Suggested solution:

check if shim is loaded on given container and install it only then? generally something to support no global window object in jest environment

sheerun:
Now that jest-dom supports node environment I could add support for it to this package as well.

Fixes #100


package.json
```diff
@@ -38,14 +39,14 @@
     "typings"
   ],
   "dependencies": {
-    "mutationobserver-shim": "^0.3.2",
```
kentcdodds: Why is this file now vendored rather than a regular dep? If it needs a change, I'd prefer that dep be updated first.

sheerun: Because it mutates window directly instead of exporting MutationObserver. It needs to be this way or you need to fork this project and use what I created in this repository (i.e. prevent assigning MutationObserver to window + check `if typeof window !== 'undefined'` before using window

sheerun: This dep is not maintained and not updated for 2 years

sheerun: If you really don't want to put it in this repository I suggest forking it and publishing under @kentcdodds/mutationobserver-shim send PR there and use your version for now. I think vendor is OK personally

kentcdodds: Gotcha. How about this, you publish a fork yourself and we can use that 👍

alexkrolick: The mutation observer polyfill is the only thing that breaks when you import dtl in a non-DOM environment. It would be great to make it not break to allow modules to consume/extend queries without going through dist/.

sheerun:
@kentcdodds Pushed change that points to my work instead of vendor

kentcdodds:
Would you like to [add yourself](../blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor) to the [contributors table](../#contributors)?

sheerun:
it's ok, next time

kentcdodds:
Sorry, this busted some stuff and I don't have time to investigate what's going on, so I'm going to revert this commit.

PR #119, opened 2018-10-10
Allow running in node environment

sheerun:
Fixed version of previous PR, I now automatically shim not existing event types on window

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -319,10 +319,8 @@ Object.entries(eventMap).forEach(
         })
       }
       const window = node.ownerDocument.defaultView
-      if (typeof window[EventType] === 'undefined') {
-        window[EventType] = class extends window.Event {}
-      }
-      const event = new window[EventType](eventName, eventInit)
+      const EventConstructor = window[EventType] || window.Event
+      const event = new EventConstructor(eventName, eventInit)
       return fireEvent(node, event)
     }
   },
--- a/src/events.js
+++ b/src/events.js
@@ -1,218 +1,201 @@
-const {
-  AnimationEvent,
-  ClipboardEvent,
-  CompositionEvent,
-  DragEvent,
-  Event,
-  FocusEvent,
-  InputEvent,
-  KeyboardEvent,
-  MouseEvent,
-  ProgressEvent,
-  TouchEvent,
-  TransitionEvent,
-  UIEvent,
-  WheelEvent,
-} = typeof window === 'undefined' ? /* istanbul ignore next */ global : window
-
 const eventMap = {
   // Clipboard Events
   copy: {
-    EventType: ClipboardEvent,
+    EventType: 'ClipboardEvent',
     defaultInit: {bubbles: true, cancelable: true},
   },
   cut: {
-    EventType: ClipboardEvent,
+    EventType: 'ClipboardEvent',
     defaultInit: {bubbles: true, cancelable: true},
   },
   paste: {
-    EventType: ClipboardEvent,
+    EventType: 'ClipboardEvent',
     defaultInit: {bubbles: true, cancelable: true},
   },
   // Composition Events
   compositionEnd: {
-    EventType: CompositionEvent,
+    EventType: 'CompositionEvent',
     defaultInit: {bubbles: true, cancelable: true},
   },
   compositionStart: {
-    EventType: CompositionEvent,
+    EventType: 'CompositionEvent',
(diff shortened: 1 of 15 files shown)


src/events.js
```diff
@@ -316,28 +299,34 @@ function fireEvent(element, event) {
...
+        // input.files is a read-only property so this is not allowed:
+        // input.files = [file]
+        // so we have to use this workaround to set the property
+        Object.defineProperty(node, 'files', {
+          value: files,
+        })
+      }
+      const window = node.ownerDocument.defaultView
+      if (typeof window[EventType] === 'undefined') {
+        window[EventType] = class extends window.Event {}
```
sheerun: Here's relevant line, btw. Previously this events were shimmed only in document.js helper

src/events.js
```diff
@@ -1,309 +1,292 @@
...
-  TransitionEvent,
-  UIEvent,
-  WheelEvent,
-} = typeof window === 'undefined' ? /* istanbul ignore next */ global : window
-
 const eventMap = {
   // Clipboard Events
   copy: {
-    EventType: ClipboardEvent,
+    EventType: 'ClipboardEvent',
```
sheerun: Probably because of changes like this bundler didn't bother to shim things like ClipboardEvent (not that it should as such shimming won't work on custom jsdom instances, just on global window)

kentcdodds:
I'm not sure why the test is breaking, but I do want to use `window.MutationObserver` if it's defined (for the in-browser use-case).

src/events.js
```diff
@@ -316,28 +299,34 @@ function fireEvent(element, event) {
   return element.dispatchEvent(event)
 }
 
-Object.entries(eventMap).forEach(([key, {EventType = Event, defaultInit}]) => {
```
davidlukerice: They were just defaulting to `Event` before due to `EventType = Event` in the function params above, when `EventType` was `undefined` (ie, `InputEvent` didn't exist before jsdom 12 so was `undefined`). Now that they are all changed to strings, like `'InputEvent'`, it's always defined so never defaulted over to `'Event'`.

Your new fix of
```
if (typeof window[EventType] === 'undefined') {
        window[EventType] = class extends window.Event {}
}
const event = new window[EventType](eventName, eventInit);
```
works, but we could also just do something like
```
const eventType = window[EventType] || window.Event;
const event = new eventType(eventName, eventInit);
```
if we wanted to keep the same behavior of not shimming.

kentcdodds: Yeah, my changes address this.

davidlukerice: oh, cool. Bit behind on the code updates then ^_^
I definitely like name `EventConstructor` over a weird lowercased `eventType`.

sheerun:
The error was because document has no ownerDocument and it can be passed as a container.

I've fixed the code to account for this case.

kentcdodds:
We've got some merge conflicts. Could you look at those please?

src/wait-for-element.js
```diff
@@ -46,7 +46,9 @@ function waitForElement(
       onDone(lastError || new Error('Timed out in waitForElement.'), null)
     }
     timer = setTimeout(onTimeout, timeout)
-    observer = new MutationObserver(onMutation)
+    const window = container.ownerDocument.defaultView
+    const MutationObserverConstructor = window.MutationObserver || MutationObserver
```
sheerun: This code is not necessary as `@sheerun/mutationobserver-shim` already uses global MutationObserver if available

kentcdodds: Oh nice! Yeah, when you work on the merge conflicts, could you fix that?

Thanks for all the work your putting into this 👍

src/wait-for-element.js
```diff
@@ -62,4 +54,11 @@ function waitForElement(
   })
 }
 
+function getDocument() {
+  if (typeof document === 'undefined') {
+    throw new Error('Could not find default container')
```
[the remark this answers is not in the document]
kentcdodds: No, this is a more elegant way of what you were doing before. To avoid running this code all you have to do is provide a `container` to `waitForElement`. If you don't, it'll throw the error (which is what was happening before).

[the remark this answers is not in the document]
sheerun: Hmm ok it seems fine, I didn't realize JS evaluates default values lazily

sheerun:
I'll send another one shortly

PR #121, opened 2018-10-10
feat(labels): add support for aria-labelledby

### LINKED ISSUE #120: Support querying by aria-labelledby for tags that aren't <label>

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

The `aria-labelledby` attribute doesn't require the labeling element to be a `<label>` tag, but dom-testing-lib only searches for `label`s in `queryByLabelText`.

I'd like to add a way to query for anything by ARIA label, which would allow doing things like selecting a card item by its heading text and then doing scoped queries for related elements like an "edit" button.

This would largely replace the need for workarounds like `getByText` -> `.$closest()` parent testid.

alexkrolick:
Fixes #120 

**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

### DIFF
--- a/src/queries.js
+++ b/src/queries.js
@@ -59,7 +59,28 @@ function queryAllByLabelText(
     .filter(label => label !== null)
     .concat(queryAllByAttribute('aria-label', container, text, {exact}))
 
-  return labelledElements
+  const possibleAriaLabelElements = queryAllByText(container, text, {
+    exact,
+    ...matchOpts,
+  }).filter(el => el.tagName !== 'LABEL') // don't reprocess labels
+
+  const ariaLabelledElements = possibleAriaLabelElements.reduce(
+    (allLabelledElements, nextLabelElement) => {
+      const labelId = nextLabelElement.getAttribute('id')
+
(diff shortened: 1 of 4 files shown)


src/queries.js
```diff
@@ -59,7 +60,28 @@ function queryAllByLabelText(
...
+      const labelledNodes = Array.from(
+        container.querySelectorAll(`[aria-labelledby~="${labelId}"]`),
+      )
+
+      return allLabelledElements.concat(labelledNodes)
+    },
+    [],
+  )
+
+  return uniq([...labelledElements, ...ariaLabelledElements])
```
kentcdodds: Actually, why don't we drop `lodash.uniq` and use:

```javascript
return Array.from(new Set([...labelledElements, ...ariaLabeledElements]))
```

alexkrolick: Sure, good trick

PR #125, opened 2018-10-12
Allow for running in node environment

sheerun:
This PR supersedes #119, incorporates all your changes, checks if window.MutationObserver is available and uses it instead of using mock, and takes care of wait-for-dom-change

### DIFF
--- a/src/wait-for-dom-change.js
+++ b/src/wait-for-dom-change.js
@@ -10,13 +10,6 @@ function waitForDomChange({
     characterData: true,
   },
 } = {}) {
-  if (typeof container === 'undefined') {
-    if (typeof window === 'undefined') {
-      throw new Error('Could not find default container')
-    } else {
-      container = window.document
-    }
-  }
   return new Promise((resolve, reject) => {
     const timer = setTimeout(onTimeout, timeout)
     const MutationObserverConstructor =
--- a/src/wait-for-dom-change.js
+++ b/src/wait-for-dom-change.js
@@ -12,6 +12,7 @@ function waitForDomChange({
 } = {}) {
   return new Promise((resolve, reject) => {
     const timer = setTimeout(onTimeout, timeout)
+    /* istanbul ignore next */
     const MutationObserverConstructor =
       typeof window !== 'undefined' &&
       typeof window.MutationObserver !== 'undefined'
@@ -39,6 +40,7 @@ function waitForDomChange({
 }
 
 function getDocument() {
+  /* istanbul ignore if */
   if (typeof window === 'undefined') {
     throw new Error('Could not find default container')
   }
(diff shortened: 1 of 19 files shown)


src/wait-for-dom-change.js
```diff
@@ -10,9 +10,21 @@ function waitForDomChange({
...
+    if (typeof window === 'undefined') {
+      throw new Error('Could not find default container')
+    } else {
+      container = window.document
+    }
+  }
   return new Promise((resolve, reject) => {
     const timer = setTimeout(onTimeout, timeout)
-    const observer = new window.MutationObserver(onMutation)
+    const MutationObserverConstructor =
```
sheerun: I left this code because I realised this can be useful if someone loads custom MutationObserver after `'@sheerun/mutationobserver-shim'` is already loaded

kentcdodds:
Ok, I've made it so the tests run via projects so they run at the same time. I also added back the coverage threshold and we're missing coverage somehow.

kentcdodds:
Ok, I think this should be good now 👍

sheerun:
yes, it'll be better to merge with 99% coverage and fix it later rather than rebase again soon

kentcdodds:
I think we're good. Time is funny. Pretty sure that this is ready to go. Let's just wait for CI.

sheerun:
I'll push fixed code coverage in a sec

kentcdodds:
We're all set :)

just-boris:
Hello, @sheerun. I see that you have added `@sheerun/mutationobserver-shim` dependency. 

I would like to read the source code of this package, but it cannot be found on Github or elsewhere. The metadata in [npm registry](https://www.npmjs.com/package/@sheerun/mutationobserver-shim) doesn't contain any link to source code.

Would you mind publishing source code somewhere to accept possible contributions?

sheerun:
Contributions should be submitted to the original repository but my fork is available at https://github.com/sheerun/MutationObserver.js

sheerun:
Ideally original repository would implement compatibility with commonjs so there's no need for a fork

just-boris:
Do you have an open pull-request, that I can upvote?

sheerun:
no

kentcdodds:
Hey @sheerun, could you please open a pull request on the upstream repository? I think I'd like to move back to that original project rather than rely on a fork long-term.

sheerun:
sorry but preparing this PR will take considerable amount of time as tests are not working for me on original repository and switching main file to commonjs requires changes in compilation pipeline to preserve backward compatibility, simply I have no time for this. maybe someone can handle it

kentcdodds:
Understandable. @just-boris, would you be able to make such a PR?

PR #129, opened 2018-10-15
fix: unsafe attribute reference

alexkrolick:
Fixes an unsafe reference that can throw an error in `debugDOM` when it receives a reference to a `document` in a real browser (ie puppeteer/codepen)

**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 


kentcdodds:
We'll need a test to cover this case :)

alexkrolick:
repro: https://codepen.io/anon/pen/bmYmPd?editors=0011

There is a test but JSDOM seems to implement `document.prototype` differently than Chrome

alexkrolick:
@kentcdodds added test

PR #137, opened 2018-10-24
Export `setNativeValue`

sivkoff:
**What**:
Exported `setNativeValue`

**Why**:
It can be used by dependent libraries, e.g. in the `react-testing-library`

**How**:
Added `setNativeValue` to the export statement in the `src/events.js` file

**Checklist**:

- [X] Documentation N/A
- [X] Tests N/A
- [X] Ready to be merged 
- [X] Added myself to contributors table


sivkoff:
OK, what about https://github.com/kentcdodds/react-testing-library/pull/210#discussion_r227855090 then?

kentcdodds:
I'm thinking that we can rework that implementation to not need `setNativeValue` :)

kentcdodds:
We worked out how to do what we needed without exposing this!

https://github.com/kentcdodds/react-testing-library/pull/210

Issue #143, opened 2018-11-02
getByName(attribute) feature request


yohaydy:
**Describe the feature you'd like:**
Some text `inputs` and especially `select` tags may not have associated label.
To select them, you need to add testid or labels.

**Suggested implementation:**
It would be nice to query those elements by `name` attribute.
I think name attributes are less prone to changes than class names, so it is fine to rely on them.

Any thoughts for/against this ?

alexkrolick:
Closing with the explanation from the PR: 

`name`, `id`, `className`, etc. are not accessible selectors - they are not visible to the user and not selectable by assistive technology like screensreaders or built-in tab navigation. testid is the designated "escape hatch" if you need to do something like this.

You can add these queries to your own code if you want by doing something like this: see #146 

If you only want to query for this infrequently, remember that regular DOM APIs work fine: document.querySelectorAll("[name=${myName}]" `)

PR #145, opened 2018-11-03
Add *byName queries

sivkoff:
**What**:
Implements feature requested in #143

**Why**:
Queries by labels are not always possible, assigning `testid` can be a solution but inconvenient sometimes.

**How**:
Added following methods (using `queryByAttribute` and `queryAllByAttribute`):
- `queryByName`
- `queryAllByName`
- `getByName`
- `getAllByName`

**Checklist**:

- [x] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,30 +1,30 @@
 {
   "dist/dom-testing-library.cjs.js": {
-    "bundled": 137666,
-    "minified": 59811,
-    "gzipped": 16554
+    "bundled": 138727,
+    "minified": 60548,
+    "gzipped": 16644
   },
   "dist/dom-testing-library.umd.min.js": {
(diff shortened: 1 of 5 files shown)


alexkrolick:
Thanks for the PR but `name`, `id`, `className`, etc. are not accessible selectors - they are not visible to the user and not selectable by assistive technology like screensreaders or built-in tab navigation. `testid` is the designated "escape hatch" if you need to do something like this. 

You can add these queries to your own code if you want by doing something like this: https://github.com/alexkrolick/dom-testing-addon-async/blob/master/README.md#with-react-testing-library

If you only want to query for this infrequently, remember that regular DOM APIs work fine: `document.querySelectorAll(`"[name=${myName}]" `)

PR #148, opened 2018-11-05
ci: test Node.js 8, 10 and 11

DanielRuf:
**What**:

All supported Node.js versions should be tested to catch any issues on newer LTS releases.

**Why**:

To ensure that there are no breaking builds on Node.js 6, 10 and 11.

**How**:

Updated the Travis CI config.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ node_js:
   - 'node'
   - '10'
   - '8'
-  - '6'
 install: npm install
 script: npm run validate
 after_success: kcd-scripts travis-after-success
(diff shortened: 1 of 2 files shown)


alexkrolick:
Node 6 is not supported, but other than that this is a good idea

DanielRuf:
> Node 6 is not supported, but other than that this is a good idea

See the engines field which says `node >= 6`.

DanielRuf:
> Node 6 is not supported, but other than that this is a good idea

If so, it should be `node >= 8` in the `engines` field.

DanielRuf:
https://github.com/kentcdodds/dom-testing-library/blob/master/package.json#L24

kentcdodds:
Thanks!

PR #151, opened 2018-11-09
Pr/unofficially support node 6

terrencewwong:
**What**:

*Unofficially* support node 6 by changing `Object.entries` to `Object.keys`

**Why**:

I opened a [PR](https://github.com/facebook/jest/pull/7346) to jest that adds an example for testing with `react-testing-library`.

jest runs it's test suite against node 6 (including the example I added), which is failing because of the call to `Object.entries`.

Switching to `Object.keys` should make the test pass so the PR can be merged 😸 

See tweet from @kentcdodds: https://twitter.com/kentcdodds/status/1060735136172466176

**How**:

With code 😸 

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A (I ran them... although not in node 6 😅 because the tests use `async`)
- [x] Ready to be merged 


kentcdodds:
Thank you!

Issue #153, opened 2018-11-12
fireEvent missing documentation and requires unnecessary element reference


Jakst:
- `dom-testing-library` version: 3.12.0 (RTL 5.1.0)
- `react` version: 16.7.0
- `node` version: -
- `npm` (or `yarn`) version: yarn 1.12.1

### Problem description:

Events like `fireEvent.keyDown` don't have any documentation.

The obvious way (for me at least) to fire these events would be `fireEvent.keyDown('Enter')`, but through looking at the code I found out that you have to do `fireEvent.keyDown(someElement, { key: 'Enter' })`, where `someElement` can be any element on the page, and is therefore very unintuitive. Key events really have no connection to an element.

Could this be, if not changed, at least clarified in documentation here and in react-testing-library?

kentcdodds:
This is a great point @Jakst! It'd be great to get improved documentation on what events are available and the typical way to use them. Would you like to open a pull request to add that to the README? Thanks!

Jakst:
I could, but I'm not really sure how to handle documentation of the random element. Just be honest about the randomness until a better fix is finished?

kentcdodds:
@Jakst, I think you misunderstand. Every event on the page must have some target. Even if that target is `document.body`.

PR #154, opened 2018-11-16
Document firing keyboard events

Jakst:
Closes issue #153

**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

### DIFF
--- a/README.md
+++ b/README.md
@@ -665,6 +665,13 @@ fireEvent.change(getByLabelText(/picture/i), {
 })
 ```
 
+**Keyboard events**: There are three event types related to keyboard input - `keyPress`, `keyDown`, and `keyUp`. When firing these you need to reference an element in the DOM. If the event you want to trigger is not bound to a certain element, you can reference the `container` returned from `render`.
+
+```javascript
+const { container } = render(<MyComponent />)
+fireEvent.keyDown(container, { key: 'Enter' })
+```
+
 ### `getNodeText`
 
 ```typescript


alexkrolick:
> If the event you want to trigger is not bound to a certain element

> I can't think of a scenario where this makes sense...

To elaborate: DOM events "bubble up"  from the target element to their parent and all the way up to the root element. They can be handled at any depth, even multiple times, unless a handler calls the event's stopPropagation() method. If you're thinking of a keypress as a global event, that means that the only event listener is at the root - the target element is whatever has focus when the event was triggered, and then it propagates up to the top.

README.md
```diff
@@ -665,6 +665,13 @@ fireEvent.change(getByLabelText(/picture/i), {
 })
 ```
 
+**Keyboard events**: There are three event types related to keyboard input - `keyPress`, `keyDown`, and `keyUp`. When firing these you need to reference an element in the DOM. If the event you want to trigger is not bound to a certain element, you can reference the `container` returned from `render`.
```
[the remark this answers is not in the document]
Jakst: Right, mental slip. This should probably be documented in `react-testing-library` as well.

Could there be any unintended consequences of leaving out the code? Maybe there should be a link to where one could find these codes, like https://keycode.info/

Jakst:
@alexkrolick Thanks for the clarification. I actually never considered that keyboard events could be bound to other elements as I've used them very lightly and have always bound them to document.

README.md
```diff
@@ -1100,13 +1108,13 @@ Thanks goes to these people ([emoji key][emojis]):
...
...tent.com/u/12592677?v=4" width="100px;"/><br /><sub><b>Paweł Mikołajczyk</b></sub>](https://github.com/Miklet)<br />[💻](https://github.com/kentcdodds/dom-testing-library/commits?author=Miklet "Code") | [<img src="https://avatars3.githubusercontent.com/u/464978?v=4" width="100px;"/><br /><sub><b>Alejandro Ñáñez Ortiz</b></sub>](http://co.linkedin.com/in/alejandronanez/)<br />[📖](https://github.com/kentcdodds/dom-testing-library/commits?author=alejandronanez "Documentation") | [<img src="https://avatars0.githubusercontent.com/u/1402095?v=4" width="100px;"/><br /><sub><b>Matt Parrish</b></sub>](https://github.com/pbomb)<br />[🐛](https://github.com/kentcdodds/dom-testing-library/issues?q=author%3Apbomb "Bug reports") [💻](https://github.com/kentcdodds/dom-testing-library/commits?author=pbomb "Code") [📖](https://github.com/kentcdodds/dom-testing-library/commits?author=pbomb "Documentation") [⚠️](https://github.com/kentcdodds/dom-testing-library/commits?author=pbomb "Tests") | [<img src="https://avatars1.githubusercontent.com/u/1288694?v=4" width="100px;"/><br /><sub><b>Justin Hall</b></sub>](https://github.com/wKovacs64)<br />[📦](#platform-wKovacs64 "Packaging/porting to new platform") |
```
kentcdodds: Sorry! But please don't edit this table manually! It's generated. Please see the CONTRIBUTING.md file for instructions on how to generate and add yourself.

Jakst: Completely unintentional, sorry. Thought I saved without formatting...

PR #157, opened 2018-11-21
Remove dom node in error messages in cypress environment

### LINKED ISSUE #156: Error message in cypress shows the whole DOM when an element is not found
- `dom-testing-library` version: 3.12.3
- `cypress-testing-library` version: 2.3.3
- `cypress` version: 3.1.2

### What you did:
Made a test fail to find an element with `cy.getByText`

### What happened:
Error message is showing the whole DOM which clutters the log

### Problem description:
It seems like there is a boolean check for cypress environment in the source code [here](https://github.com/kentcdodds/dom-testing-library/blob/81299da3e28f46badceb61b572596e502657e3ed/src/query-helpers.js#L14) but it's not working.

### Suggested solution:
Changing [this line](https://github.com/kentcdodds/dom-testing-library/blob/81299da3e28f46badceb61b572596e502657e3ed/src/query-helpers.js#L14) to
`const inCypress = global.Cypress` works on my local.

I'm not sure if this line worked before but `window` is undefined in cypress 3.1.2. Maybe cypress doesn't expose the window variable anymore.

I could change it to `const inCypress = (typeof window !== 'undefined' && window.Cypress) || global.Cypress` to keep it backward compatible.

I'll have a new PR addressing this issue with the above code changes if it was intended to hide the DOM for cypress environment.

misoguy:
Resolves #156 and also resolves https://github.com/kentcdodds/cypress-testing-library/pull/25

**What**: This PR removes the dom nodes being printed from the error message shown in cypress when a test fails to locate the specified dom node.

**Why**: Showing the whole dom in cypress clutters up the error log.

**How**: There was already a condition present to prevent printing out dom node in cypress but it is not valid in latest version of cypress which is why I added another condition to make this work.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 


kentcdodds:
Also, thanks!

misoguy:
@kentcdodds added the `typeof` check

Issue #162, opened 2018-11-27
Overriding `data-testid` requires a lot of work, and can't be done completely


RoystonS:
### Describe the feature you'd like:

An easier way to override the `data-testid` string.

I'm really sorry to raise this: I'm aware of #76, #128 and #136, and the current proposed solution.  But they're _really_ hard to implement in practice, fully, and a very small amount of configuration in `dom-testing-library` would make that significantly easier for everybody who needs to override the test id.

I definitely agree with @kentcdodds about the composability of JS, wanting to keep things simple, and the fact that it _is_ currently _possible_ to override _most_ of the layers that ordinarily expose the original `xxxByTestId` functions, but the problem is that `data-testid` is baked in at a rather low level, and overriding all the things layered on top is actually quite hard/impossible.  As more things are added to the framework (e.g. `within`), this gets harder, and consumers of the library will need to track and override more things. 

Here's what you have to do right now if you need to override `data-testid`:

1. Implement your own top-level wrapper for `react-testing-library`.  (I've already done this, and would expect to do so, just as I did for Enzyme.  That's not an issue.)
1. Do a fair bit of shenanigans around overriding `render` (and `rerender`) to return a set of queries where the implementation of `xxxByTestId` is changed to your own implementation.  This is not trivial, as shown by the number of times that sample code has been tweaked in the `react-testing-library` docs. These all then need maintaining as APIs change.
1. (Not documented anywhere else afaik) Override `within`/`getQueriesForElement` because that'll return the original `xxxByTestId` functions too.
1. Tell all your developers that they mustn't use any functions imported directly from `dom-testing-library` because _those_ `xxxByTestId` functions won't work properly.  (This rules out being able to use any other helper libraries that make use of `dom-testing-library`.)

### Suggested implementation:

A simple design such as the one proposed by @davidgilbertson in #136 would solve all of this at a stroke and guarantee that whatever functions are layered on top of the library, whatever changes are made to higher level APIs, they'll all work, because we've not had to touch them, and have merely configured the low-level _primitive_ operations directly.

### Describe alternatives you've considered:

Doing it all by hand. I've not just considered this - I've been doing it.  I've almost got it all working, but it's taken a good few hours, and I still haven't sorted out how to override `within` in a good way.

[...]

PR #163, opened 2018-11-28
feat(configureTestIdAttribute): add ability to override data-testid [specific config]

RoystonS:
N.B. This is one of two alternative PRs with slightly differing configuration APIs.  This one contains a specific configuration API exposing one function targeted at changing the test id attribute.

**What**:

Introduce an API to allow `data-testid` to be overridden without consumers needing
to construct and maintain a set of higher level decorator functions to override the various 'xxxByTestId' functions as exposed in various places.

**Why**:

The existing API, where `data-testid` cannot be changed at the low level, requires a significant amount of effort to customise the higher-level APIs. See issue #162 for more discussion.

**How**:

Ultimately we needed to parameterise what was previously some hardcoded strings in `queries.js`. I would have preferred to keep the changes local to `queries.js`, but we assume elsewhere that all exports from `queries.js` are queries, and I didn't want to alter that.  So I added a simple `config.js` module to store the actual configuration, and `queries.js` references that and `index.js` exposes the tiny configuration API.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -419,6 +419,18 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "RoystonS",
+      "name": "Royston Shufflebotham",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/19773?v=4",
+      "profile": "https://github.com/RoystonS",
+      "contributions": [
+        "bug",
+        "code",
+        "doc",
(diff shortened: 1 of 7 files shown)


kentcdodds:
If we do this, we'll go with the other PR 👍

PR #164, opened 2018-11-28
feat(configureTestIdAttribute): add ability to override data-testid [generic config]

RoystonS:
N.B. This is one of two alternative PRs with slightly differing configuration APIs.  This one contains a generic configuration API exposing one configuration function taking an object which could be extended for future use.  That may desirable or not, hence the two PRs.

**What**:

Introduce an API to allow `data-testid` to be overridden without consumers needing
to construct and maintain a set of higher level decorator functions to override the various 'xxxByTestId' functions as exposed in various places.

**Why**:

The existing API, where `data-testid` cannot be changed at the low level, requires a significant amount of effort to customise the higher-level APIs. See issue #162 for more discussion.

**How**:

Ultimately we needed to parameterise what was previously some hardcoded strings in `queries.js`. I would have preferred to keep the changes local to `queries.js`, but we assume elsewhere that all exports from `queries.js` are queries, and I didn't want to alter that.  So I added a simple `config.js` module to store the actual configuration, and `queries.js` references that and `index.js` exposes the tiny configuration API.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 


src/config.js
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export function configure(configDelta) {
+  config = {
+    ...config,
+    ...configDelta,
+  }
+}
+export function getTestIdAttribute() {
+  return config.testIdAttribute
```
alexkrolick: This looks good to me. My only nit is that I don't think there needs to be a special method to grab just this attribute; export the whole config object or a getConfig function instead.

alexkrolick: Configure could also take an update function:

```js
configure(currentConfig => { return {...newConfig} })
```

This would be nice if you want to read and extend subobjects or arrays (like the current set of queries) without importing the default value into the host file.

davidgilbertson:
This is great, nice one @RoystonS .

Just a note (to self, you or @kentcdodds), the "What if my project already uses data-test-id?" section in `react-testing-library` can be replaced with a link to this, once merged.

RoystonS:
> This is great, nice one @RoystonS .
> 
> Just a note (to self, you or @kentcdodds), the "What if my project already uses data-test-id?" section in `react-testing-library` can be replaced with a link to this, once merged.

Absolutely. Very much looking forward to being able to do that!

RoystonS:
I've added another commit for @alexkrolick's extended configuration use case. That required a new set of tests and a new documentation section.

RoystonS:
I've made the usual sacrifices to eslint to make it shut up; this is now ready for review.

README.md
```diff
@@ -928,6 +942,19 @@ console.log(prettyDOM(div))
...
+The library can be configured via the `configure` function, which accepts:
+
+- a plain JS object; this will be merged into the existing configuration. e.g. `configure({testIdAttribute: 'not-data-testid'})`
+- a function; the function will be given the existing configuration, and should return a plain JS object which will be merged as above, e.g.
+  `configure(existingConfig => ({something: [...existingConfig.something, 'extra value for the something array']}))`
+
+Configuration options:
+
+`testIdAttribute`
+: The attribute used by `getByTestId` and related queries. Defaults to `data-testid`. See [`getByTestId`](#getbytestid).
```
kentcdodds: Can this be placed on the same line so there's no space in the rendered output?

screen shot 2018-11-29 at 8 40 37 am

src/__tests__/config.js
```diff
@@ -0,0 +1,60 @@
...
+
+  describe('getConfig', () => {
+    test('returns existing configuration', () => {
+      const conf = getConfig()
+      expect(conf.testIdAttribute).toEqual('data-testid')
+    })
+  })
+
+  describe('configure', () => {
+    test('merges a delta rather than replacing the whole config', () => {
```
kentcdodds: Hmmm... I think the fact that you can _add_ random properties to the configuration is an implementation detail and not necessarily a feature we care to support. By that I mean, if we change the implementation in the future to only allow you to configure keys of things we support then I don't want to have to bother with fixing these tests.

So if we could make only assertions for the features we care to support that would be great.

More specifically, this test is fine, so long as we change the assertion to:

```javascript
expect(conf).toMatchObject({testIdAttribute: 'data-testid'})
```

The fact that `newKey` exists as well is a coincidence and not necessarily something we need to assert in our tests.

src/__tests__/config.js
```diff
@@ -0,0 +1,60 @@
...
+    test('overrides existing values', () => {
+      configure({testIdAttribute: 'new-id'})
+      const conf = getConfig()
+      expect(conf.testIdAttribute).toEqual('new-id')
+    })
+
+    test('passes existing config out to config function', () => {
+      // Create a new config key based on the value of an existing one
+      configure(existingConfig => ({
+        newKey: `${existingConfig.testIdAttribute}-derived`,
```
kentcdodds: Rather than `newKey` in this test, we could set this to the `testIdAttribute` and get the same value out of the test without asserting on a non-feature.

README.md
RoystonS: Hmm. Might be another `prettier` funsie. I'm sure I put them on the same line originally. I'll have a go...

src/__tests__/config.js
RoystonS: Yup, will update tests so they don't create random properties.  Would have been easier to test had there been _two_ supported properties from the start.

RoystonS:
Very much appreciated!

PR #165, opened 2018-12-04
Pr/fireevent on window

VadimBrodsky:
**What**:

Change the typings for `fireEvent` to accept Window in addition to DOM Elements.

**Why**:

Some events such `postMessage` are not fired on a DOM element but rather on Window.

**How**:

- Updated typings
- Add test
- Expose window to the test environment

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged 


dbismut:
Hi @VadimBrodsky, I don't know if this PR was supposed to fix this, but at the moment I'm having an error when doing this:

```js
 fireEvent.mouseMove(window, { clientX: 20, clientY: 50 })
```

Console throws:
**TypeError: Cannot read property 'MouseEvent' of undefined**

It works fine when using `document` instead of `window`.

This is using `react-testing-library@6.1.2` which itself should be using `dom-testing-library@^3.19.0`

kentcdodds:
Hmmm... Could you open a PR with a failing test? Thanks!

dbismut:
Hey @kentcdodds here is a csb link with a failing test:
https://codesandbox.io/s/3kxw3q8z8m

Does that help?

kentcdodds:
Got it. I'll have a fix shortly.

kentcdodds:
A fix will be released in a minute.

Issue #158, opened 2018-11-22
getByValue cannot get element by dynamically assigned value


eunjae-lee:
- `dom-testing-library` version: 3.12.4
- `react` version: not using with react
- `node` version: 8.11.1
- `npm` (or `yarn`) version: npm 5.6.0

### Relevant code or config:

```js
test('get element by its dynamically assigned value', () => {
  const {getByValue, queryByValue, getByTestId} = renderIntoDocument(`
  <div>
    <input placeholder="name" type="text" data-testid="name" />
  </div>
  `)
  getByTestId('name').value = 'Norris'
  expect(getByValue('Norris').placeholder).toEqual('name')
  expect(queryByValue('Norris').placeholder).toEqual('name')
})
```

### What you did:

There's an `input` element (without initial `value` attribute). I put some value in it with JavaScript. 

### What happened:

`getByValue` or `queryByValue` cannot get the element.

```
  ● get element by its dynamically assigned value

    Unable to find an element with the value: Norris.
```

### Problem description:

According to `queryAllByAttribute` at `src/query-helpers.js`, it's executing some code like

```
container.querySelectorAll("[value]")
```

Since the `input` element didn't have any `value` attribute at the first place, this query cannot get it.

### Suggested solution:

It should still get `input` element unless this behaviour is intended.
Querying it with `[value]` didn't work. Getting value with `node.getAttribute("value")` didn't work either. So I guess the following code could solve the problem.

```
Array.from(container.querySelectorAll("input")).filter(node => node.value === value)
```

Let me know what you think!

alexkrolick:
The .value setter/getter and the DOM attribute aren't necessarily in sync (see [Codepen](https://codepen.io/alexkrolick/pen/NELEKE?editors=1011
)), so you'd probably need to either concat the results of both the current querySelectorAll and the proposed filtered DOM traversal, and maybe dedupe the results with some documented hierarchy (probably preferring the dynamic `.value`).

PR #166, opened 2018-12-05
fix(queryAllByAttribute): query `input` elements with dynamically assigned value

eunjae-lee:
**What**: The implementation of `queryAllByAttribute` has been changed so that it can successfully query `input` elements with dynamically assigned `value`.

**Why**: When `value` is dynamically assigned to an `input`, not by an attribute on HTML, `queryAllByAttribute` won't query that element at the moment.

**How**: `queryAllByAttribute` behaves differently when its `attribute` is equal to `"value"`. I'm not sure if this is the best way to implement this. Feedbacks are welcomed.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged 


eunjae-lee:
closes #158

eunjae-lee:
Thoughts? @alexkrolick

kentcdodds:
I don't like this. I might be inclined to add a new query called `queryByProperty` which compares DOM node properties directly rather than `getAttribute`. I honestly don't really like getting by attributes or properties because those are not typically visible to the end user. Getting by the `value` property of an `input` is acceptable though because the value of an input _is_ visible to the user.... Hmmm...

What do you think @alexkrolick?

kentcdodds:
True. We do have [`getBySelectText`](https://github.com/kentcdodds/dom-testing-library#getbyselecttext) perhaps we should do `getByInputText` and `getByTextAreaText` as well?

sompylasar:
The only reason for querying by attribute is because the attributes are proxies to what the browser displays to the user (there's no good way to query the browser render tree itself and get anmotated positioned semantic items out of it).

alexkrolick:
Maybe the queries for select value and textarea could be combined with the input query - `getByCurrentValue`? This would smooth over the select value in the same way React does it.

I think this would work for  most use cases because the value property and the DOM attribute are initially in sync when an element is created, and when they are different, the property is what is visually displayed. 

https://codepen.io/alexkrolick/pen/NELEKE?editors=1011

kentcdodds:
Ok, I'm in favor of that. Let's go ahead and add it and we can add a deprecation message to the README for the `getBySelectText`. Then we can officially remove it in a future breaking change (when we have sufficient justification to break stuff).

Thanks for the PR anyway @eunjae-lee! Would you be interested in making a PR to add a `getByCurrentValue` query?

eunjae-lee:
Thanks all for the feedbacks. I really love how this conclusion has been made!

@kentcdodds I'm still interested in contributing on this. To make sure, `getByCurrentValue` will refer to value property or (if not exists) DOM attribute, right?

kentcdodds:
I think the property should be sufficient.

eunjae-lee:
Hmm. Although value property hasn't been assigned to the element, if it has DOM attribute, end-user will see it on screen. Shouldn't we get elements like this?

alexkrolick:
The value property is initially set to the same thing as the attribute, unless I'm interpreting this wrong (check console) https://codepen.io/alexkrolick/pen/NELEKE?editors=1011

eunjae-lee:
@alexkrolick oh you're right.

PR #169, opened 2018-12-07
feat(queryByCurrentValue)

### LINKED ISSUE #158: getByValue cannot get element by dynamically assigned value

eunjae-lee:
**What**: It adds `queryByCurrentValue`, `getByCurrentValue`, `queryAllByCurrentValue` and `getAllByCurrentValue`.

**Why**: `getByValue` cannot get elements with `value` property (it only checks DOM `value` attributes)
(Discussed at #166)

**How**: Added a set of `(get|query)(All)?ByCurrentValue` methods.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged 

Feel free to give any feedbacks.

---

Closes #158


src/queries.js
```diff
@@ -356,8 +387,12 @@ export {
   getAllByTitle,
   queryByValue,
   queryAllByValue,
+  queryByCurrentValue,
+  queryAllByCurrentValue,
```
kentcdodds: Could you move both of these down two lines. I kinda like to group things together.

src/queries.js
```diff
@@ -180,6 +180,22 @@ function queryByAltText(...args) {
...
 
+function queryAllByCurrentValue(
+  container,
+  value,
+  {exact = true, collapseWhitespace = true, trim = true} = {},
+) {
+  const matcher = exact ? matches : fuzzyMatches
+  const matchOpts = {collapseWhitespace, trim}
+  return Array.from(container.querySelectorAll(`input,textarea,select`)).filter(
+    node => matcher(node.value, node, value, matchOpts),
```
kentcdodds: This wont work for selects. Please look at (and maybe just use) the logic in `queryAllBySelectText` for selects.

src/__tests__/element-queries.js
```diff
@@ -565,4 +565,19 @@ test('getByText ignores script tags by default', () => {
   expect(getAllByText(/hello/i, {ignore: false})).toHaveLength(3)
 })
 
+test('get element by its dynamically assigned value', () => {
```
kentcdodds: Could we also have a test for a textarea and a select. Can be in this test or in a new one I don't care which.

eunjae-lee:
Thanks for the review. I made changes regarding your three comments.

alexkrolick:
Looks good! Just need docs and a test to execute line 358 (the error state for queryAllBy...) to keep coverage at 100%.

Also it might be nice to add an integration test for `fireEvent.change()` with `getCurrentValue`.

src/__tests__/element-queries.js
```diff
@@ -565,4 +565,63 @@ test('getByText ignores script tags by default', () => {
...
+      <option selected value="AK" >Alaska</option>
+      <option value="AZ">Arizona</option>
+    </select>
+  `)
+
+  expect(getByCurrentValue('Alaska').id).toEqual('state-select')
+  expect(queryByCurrentValue('Alaska').id).toEqual('state-select')
+
+  getByTestId('state').value = 'AL'
+  expect(getByCurrentValue('Alabama').id).toEqual('state-select')
```
kentcdodds: I wonder if it's at all confusing that we are expected to set the `value` to `AL` but when we use the query we search for `Alabama`.

I realize that's just how things work (the value for a select has a display value), but I feel like this could be a point of confusion for people. My guess is they'd prefer to select by `AL` instead which isn't what is shown to the user but feels more intuitive here. I'm not sure of the best solution to this... Maybe allow for either the display value or the actual value?

eunjae-lee: That makes sense. To end-users `Alabama` is a value, but `AL` is the actual value.

eunjae-lee: However, `selectElem.value` could be either `1`, `AL`, or whatever user cannot know. In a perspective of writing tests as user would use it, I'd rather choose to restrict it to display value.
What do you think?

kentcdodds: Yeah, I think `displayValue` is probably a better mental model, though it may be a tiny confusing for `input` and `textarea` where the _display_ value is actually the _actual_ value but I think that's alright. Besides, people should normally be getting these elements by their label anyway, so I'm not too concerned about this slightly less-awesome API.

Let's go with `getByDisplayValue` 👍

eunjae-lee:
@alexkrolick I've added a test to check if it throws.

@kentcdodds I've renamed and updates the README.md. I didn't know how to edit it, so it's kind of a draft. Where do you want to put the section in the document? And how do you want to explain it to users? Take a look at the doc and let me know. Besides, I'm not a native English speaker, so pretty much worried about it 😅

README.md
```diff
@@ -86,6 +86,7 @@ when a real user uses it.
   - [`getByAltText`](#getbyalttext)
   - [`getByTitle`](#getbytitle)
   - [`getByValue`](#getbyvalue)
```
kentcdodds: Let's remove the docs for `getByValue` and `getBySelectText` because we want to encourage people to use the `getByDisplayValue` instead.

README.md
```diff
@@ -409,6 +410,55 @@ Returns the element that has the matching value.
...
+  container: HTMLElement,
+  value: TextMatch,
+  options?: {
+    exact?: boolean = true,
+    collapseWhitespace?: boolean = false,
+    trim?: boolean = true,
+  }): HTMLElement
+```
+
+Returns the element that has the matching display value.
```
kentcdodds: > Returns the `input`, `textarea`, or `select` element that has the matching display value.

README.md
```diff
@@ -409,6 +410,55 @@ Returns the element that has the matching value.
...
+
+#### `input`
+
+```javascript
+// <input type="text" id="lastName" />
+// document.getElementById('lastName').value = 'Norris'
+
+const lastNameInput = getByDisplayValue('Norris')
```
kentcdodds: > const lastNameInput = getByDisplayValue(document.body, 'Norris')

README.md
```diff
@@ -409,6 +410,55 @@ Returns the element that has the matching value.
...
+```
+
+#### `textarea`
+
+```javascript
+// <textarea id="messageTextArea"></textarea>
+// document.getElementById('messageTextArea').value = 'Hello World'
+
+const messageTextArea = getByDisplayValue('Hello World')
```
kentcdodds: > const messageTextArea = getByDisplayValue(document.body, 'Hello World')

README.md
```diff
@@ -409,6 +410,55 @@ Returns the element that has the matching value.
...
+```javascript
+// <select id="state-select" data-testid="state">
+//   <option value="">State</option>
+//   <option value="AL">Alabama</option>
+//   <option selected value="AK" >Alaska</option>
+//   <option value="AZ">Arizona</option>
+// </select>
+
+const selectElement = getByDisplayName('Alaska')
+console.log(selectElement.id)
```
kentcdodds: Remove the `console.log` here (it's not necessary and it's nice to be consistent with the others). Then fix it to be:

> const selectElement = getByDisplayName(document.body, 'Alaska')

eunjae-lee:
@kentcdodds done! I've used `container` instead of `document.body` since I found out other parts of the doc used `container`.

kentcdodds:
Looks like we're missing some coverage. You can open the coverage report in the browser in `coverage/lcov-report` to see what's missing.

eunjae-lee:
@kentcdodds coverage 100% done

README.md
```diff
@@ -337,77 +335,54 @@ content is in an inline script file, then the script tag could be returned.
 
 If you'd rather disable this behavior, set `ignore` to `false`.
 
-### `getByAltText`
```
alexkrolick: Why are these doc sections being deleted?

eunjae-lee: @alexkrolick refer to https://github.com/kentcdodds/dom-testing-library/pull/169#discussion_r240464612

eunjae-lee:
Updated the doc. Thanks for pointing it out @alexkrolick

eunjae-lee:
Awesome! Thank you :) This merge made my day!!!

Issue #171, opened 2018-12-07
Allow custom normalization for matchers


RoystonS:
### Describe the feature you'd like:

It would be really helpful to be able to provide custom normalizer functions to the get/query functions. In our applications we have a lot of special RTL-related [UCC characters such as LRE, RLE, RLM, PDF, etc.](https://www.w3.org/International/questions/qa-bidi-unicode-controls) and looking up elements by text is made very hard if tests have to specify those characters precisely. So we'd really like the ability to match on text ignoring those characters.  Having to specify them precisely in the tests makes them very fragile and hard to read.

There are already a couple of normalizer options such as `trim` and `collapseWhitespace`. I _could_ suggest adding another specific one such as `removeUCC`, but it would be more flexible to allow a custom normalizer function to be specified, as that would obviate future requests for more options.

### Suggested implementation:

It's pretty straightforward: another `option` that gets passed through to the low-level `normalize` function in `matches.js`.

There would be a decision to be made about what would happen if either `trim` or `collapseWhitespace` were passed _as well as_ a custom normalizer function.  There are three obvious options:

1. only run the custom normalizer
1. run the trim + whitespace handling before the custom normalizer
1. run the trim + whitespace handling after the custom normalizer
1. (bonus weird option) run the trim + whitespace handling both before and after the custom normalizer

### Describe alternatives you've considered:

There are some alternatives, but they're not lovely:

- Use a regex to provide a 'loose' pattern.  Not a great option: without knowing where UCC characters might appear, we'd have to write something like `.getByText(/this.*is.*a.*series.*of.*words/)` (which would be much too hard work and much too loose) or replacing the `.*` with a regex character class containing the precise UCC characters we'd like to ignore, but that would be unreadable.
- Use custom matcher functions instead.  We've tried doing this: implementing normalization in those is rather a lot of work as we end up duplicating most of what's in `matches.js` as a result (to cope with fuzzy matching, regex matching, etc) which we still want on top of being able to remove UCC characters.  Being able to write `.getByText(/this is a (big)? series of words/, { normalizer: removeUCC })` would be expressive and flexible.

### Teachability, Documentation, Adoption, Migration Strategy:

As above, I'm guessing it could be another `option` passed to the get/query functions. The biggest confusion I could think of is the interoperability between the existing normalization options and a custom normalization function.

No backwards compatibility issues: it would be an additional API.

I'm happy to provide a PR if you're interested in progressing this.

kentcdodds:
[...]

PR #172, opened 2018-12-08
feat(custom normalizer): allow custom control of normalization

RoystonS:
**What**:

(PR for issue #171)

Adds an optional {normalizer} option to query functions; this
is a transformation function run over candidate match text
after it has had `trim` or `collapseWhitespace` run on it,
but before any matching text/function/regexp is tested
against it.

The use case is for tidying up DOM text (which may
contain, for instance, invisible Unicode control characters)
before running matching logic, keeping the matching logic
and normalization logic separate.

**Why**:

It's otherwise not easy to apply normalization logic to DOM text without having to replace large portions of the matching logic.

**How**:

Added a new optional `normalizer` function which applies after `trim` + `collapseWhitespace` but before matching logic.

I had to make the same changes in several places in the code as there's a lot of similar option-handling code around, unfortunately.  That also meant lots of test cases to ensure all those paths are tested.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 


src/__tests__/text-matchers.js
```diff
@@ -194,3 +194,115 @@ cases(
     },
   },
 )
+
+// A good use case for a custom normalizer is stripping
+// out UCC codes such as LRM before matching
```
alexkrolick: Can we expand out acronyms like this for those who might be unfamiliar with Unicode (UCC -> Unicode Control Character)? 😄

RoystonS: Yep, I did expand it out in the user-facing docs, but I'll expand it here too.

src/matches.js
```diff
@@ -2,12 +2,16 @@ function fuzzyMatches(
   textToMatch,
   node,
   matcher,
-  {collapseWhitespace = true, trim = true} = {},
+  {collapseWhitespace = true, trim = true, normalizer} = {},
```
kentcdodds: I would prefer it if we could accomplish `collapseWhitespace` and `trim` from within a built-in `normalizer`. That normalizer can take these options into account:

```js
{collapseWhitespace = true, trim = true, normalizer = getDefaultNormalizer({collapseWhitespace, trim})} = {}
```

It may also be a good idea to soft-deprecate `collapseWhitespace` and `trim` in favor of providing a custom normalizer. This would amount to removing the docs for those options and replacing it with docs for `normalizer`. Then in a future version we can remove the options and have it built-into our default normalizer.

I just don't like having multiple ways to do the same thing, and I like the normalizer as it's more powerful.

RoystonS: Yep, I definitely like the idea of removing `collapseWhitespace` and `trim`, but I didn't want to be the one to suggest it! :)

Just to check my understanding, let me rephrase what I think you're proposing:

a) publicly expose a new thing called `getDefaultNormalizer` that takes `collapseWhitespace` and `trim` options
b) soft-deprecate `collapseWhitespace` and `trim` by not documenting them any more
c) the default value of the `normalizer` option becomes the output of `getDefaultNormalizer`, called with whatever `trim` and `collapseWhitespace` options are provided (until we deprecate + remove them)

I'm not a huge fan of the idea of an array of normalizers: functions are nicely composable already, and with an array you get issues about who owns the array, how shared is it, do I have to clone it before pushing my normalizer, etc?  If you just have one function, composition is straightforward to comprehend, and you don't need to reimplement anything: you just compose over the default `getDefaultNormalizer` if you want it.

(In fact, my team have already doubled up all the queries, producing ...WithoutUCC versions, such as `getByTextWithoutUCC`, and they still accept a `normalizer` but chain it with a `withoutUCC` normalizer, which they choose to run first.)

I'll update the PR.

RoystonS: Hmm. One little thought:  let's say we have a user who writes
`queryBySomething(..., { trim: true, collapseWhitespace: false, normalizer: customNormalizer })`.

Long term we're obviously suggesting that `trim` and `collapseWhitespace` would be removed, but in the short term what should that call do where they specify values for `trim` and `collapseWhitespace` _and_ `normalizer` (which is supposed to replace both)?

Should we error, saying that they can't specify a custom `normalizer` _and_ values for `trim`/`collapseWhitespace`?  I suspect we _should_ error, otherwise we're temporarily giving them an API that will stop working later when we remove the properties, whereas if we error, we force them to do the right thing - composing with `getDefaultNormalizer` from the get go - and that'll continue to work even when `trim` + `collapseWhitespace` are removed as first-class parameters...?

alexkrolick: Hmm, one more idea: With an _object_ you could turn off a default normalizer by passing false or null instead of a function, so the API is almost the same as the current top-level options, and makes it easier to not accidentally opt out of new defaults. Depending on implementation, you could probably also accept an array of (optionally anonymous) functions, and may as well accept a single function as well.

```
normalizers/
  trim.js
  collapseWhitespace.js
  index.js (exports all)
```

```js
import * as normalizers from 'dtl/(dist)/normalizers'
// Or
import normalizers from 'dtl'

getByText(document.body, 'abc', {
  normalizers,
})

// Accept a single function as well, that doesn't require any option handling
getByText(document.body, 'abc', someCustomFn)

getByText(document.body, 'abc', {
  normalizers: {
    // Possibly merge by default if receiving an object?
    ...normalizers,
    // Falsy values don't get called
    trim: false,
    collapseWhitespace: null,
    // A custom function
    replaceUnicode,
  }
})```

Comment (RoystonS) [reply src/matches.js]: I think discovery would be a little harder with that API: firstly you've got two different things called `normalizers`: an import, which I'm guessing is a _default_ set of normalizers, and a parameter specifying a chosen bunch of normalizers.

Ordering also isn't very obvious: if I were to specify `trim: true` and a custom normalizer, which would run first? Or two custom normalizers? I can't tell or control it.  Having an artbitrary-extensible object also mixed in with well-known named parameters is a bit tricky: you wouldn't ever be able to add more well-known named parameters later for fear of confusion with user-specified ones, and mixing booleans and functions in a single object is a little unusual.

I still like @kentcdodds's suggestion of specifying a single `normalizer` function with a `getDefaultNormalizer` function exposed. It feels really obvious what's going on, and both user and library extensibility is simple and separate.

Comment (alexkrolick) [reply src/matches.js]: Sorry, this comment is a bit long!

tl;dr:
- probably need a config option to set the default
- per-test overrides given you have a custom setup on top of the default are a bit complex to do right

Assumptions:

- Usually you want a suite-wide default plus a way of turning _off_ very specific things for certain tests to check for edge cases. 
- You generally don't want to "assume full control of normalization"; you want to extend the default behavior in a forwards-compatible way so you get new features and bug fixes, while adding in support for some specific additional cases.

Playing out the use cases using the proposed API:

1. If you're writing something you're going to use across the whole app, you'll probably want some way to set the default normalizer function so you don't have to add normalizer options to every single query. You'll also probably want some way of customizing the options, so you'll need an internal version of `getDefaultNormalizer` which may even wrap that function and all its options (doing something to avoid namespace collisions).

---

2. If you're doing a one-off normalizer override:

```js
// ignore defaults completely
// 🚀boldly taking normalization into our own hands
getByText(elem, 'text', {
  normalizer: customNormalizer,
})

// take options for custom factory, but ignore forwards compat completely
// (don't call default normalizer even if options match)
// 🤔 what if we want to track upstream functionality?
getByText(elem, 'text', {
  normalizer: getCustomNormalizer({
    trim: false, // don't care if this is a default option
    removeFoo: false,
  })
})

// with default handler and default options
// assuming customNormalizer takes and returns a string
// 👍 composition ftw
getByText(elem, 'text', {
  normalizer: str => 
    customNormalizer(defaultNormalizer(str)),
})

// with default handler and default options
// assuming customNormalizer takes and returns a string
// a realistic use case if you have any conditional logic
// or multiple custom normalization functions
// still a 👍, but you probably wouldn't want to inline this
getByText(elem, 'text', {
  normalizer: str => {
	const defaultResult = defaultNormalizer(str)
    // ...some custom logic here...
    const evenMoreNormalized = customNormalizer(defaultResult)
    return evenMoreNormalized
  },
})

// using the getter to customize default normalizer
// assuming the customNormalizer takes and returns a string
// 🤔 a bit complicated now
getByText(elem, 'text', {
  normalizer: str => 
    customNormalizer(
      getDefaultNormalizer({ trim: false})(str)
    ),
})

// using factories for both default and custom
// custom has no options but takes a function and returns a string
// 🤔 potentially fragile, what if we want options later?
getByText(elem, 'text', {
  normalizer: getCustomNormalizer(getDefaultNormalizer({ trim: false})),
})

// using factories for both default and custom
// custom has options as the second argument
// 🤔 not really leveraging composition at this point
getByText(elem, 'text', {
  normalizer: getCustomNormalizer(
    getDefaultNormalizer({ trim: false}),
    {removeUCC: false}
  ),
})

// hide default normalizer inside custom normalizer
// add default options to custom normalizer factory
// 🤔 potential name collisions
getByText(elem, 'text', {
  normalizer: getCustomNormalizer({
    removeUCC: false,
    trim: false
  }),
})

// hide default normalizer inside custom normalizer
// take default options as separate argument
// 🤔not that composable any more
getByText(elem, 'text', {
  normalizer: getCustomNormalizer(
    { removeUCC: false },
    {trim: false}
  ),
})

// add default options to custom normalizer factory under a namespace
// 🤔 not that composable any more
getByText(elem, 'text', {
  normalizer: getCustomNormalizer({
    removeUCC: false,
    defaultNormalizerOptions: {
      trim: false,
    },
  }),
})
```
---

3. Defining a one-off handler assuming you have set a global custom handler:

(All patterns from above apply except now the default behavior is split between the library default and the custom default)

kentcdodds: That feels... complicated. I'd love to try and keep the library simple...

RoystonS: Yes, that's rather a complex API, plus I (as somebody who actually needs to use a custom normalizer) don't want to apply the same normalizer everywhere.

I'm just finishing up the docs changes for the API I posted at https://github.com/kentcdodds/dom-testing-library/pull/172#discussion_r240053817. That API feels pretty good to use in practice.

alexkrolick: The API in the code samples I posted is the one in your comment, just fleshed out with more scenarios. I think it looks OK for one-offs except some complexity when inlining the getter that returns a function. If you want to use this project-wide it does seem a bit difficult especially if you want to add options to the custom function or compose several conditionally.

alexkrolick: I'm ok with going with the proposed API and dealing with any global config or alternate APIs later if someone asks for them, unless someone has a better idea to address those right now.

README.md
```diff
@@ -802,9 +794,47 @@ affect the precision of string matching:
...
+trimming whitespace from the start and end of text, and collapsing multiple
+adjacent whitespace characters into a single space.
+
+If you want to prevent that normalization, or provide alternative
+normalization (e.g. to remove Unicode control characters), you can provide a
+`normalizer` function in the options object. This function will be given
+a string and is expected to return a normalized version of that string.
+
+Note: Specifying a value for `normalizer` _replaces_ the built-in normalization, but
+you can call `getDefaultNormalizer` to obtain a built-in normalizer, either
```
kentcdodds: I'd prefer to not allow people to call this function and instead simply export it:

```
export const normalizer = getDefaultNormalizer({trim: true, collapseWhitespace: true})
```

RoystonS: Ok, easy to do, but then somebody wouldn't be able to get hold of, say, a default normalizer that only trims but doesn't collapse whitespace.

If you're happy with that, that's fine, and I'll hide `getDefaultNormalizer` and take all mention of `trim` and `collapseWhitespace` out of the docs.

kentcdodds: > then somebody wouldn't be able to get hold of, say, a default normalizer that only trims but doesn't collapse whitespace.

I'm thinking that's a pretty minimal use-case, and if it is, they just need to copy/paste this:

```js
function getDefaultNormalizer({trim = true, collapseWhitespace = true} = {}) {
  return text => {
    let normalizedText = text
    normalizedText = trim ? normalizedText.trim() : normalizedText
    normalizedText = collapseWhitespace
      ? normalizedText.replace(/\s+/g, ' ')
      : normalizedText
    return normalizedText
  }
}
```

Considering we're getting simplicity in exchange for a tiny bit of flexibility for 1% of use cases I think that's a win :)

RoystonS: I'll also have to remove the example for how to match text without trimming, for instance:
`getByText(node, 'text', {normalizer: getDefaultNormalizer({trim: false})})`

There'd be no easy way to do that without `getDefaultNormalizer`: they'd have to implement their own whitespace-collapsing normalizer.  I don't have much of an opinion on that, but presumably the existing `trim` and `collapseWhitespace` options were added because somebody wanted to be able to turn them off?

kentcdodds: Ah, that makes sense, considering I want to deprecate `trim` as an option. Ok, let's stick with what you have. I'd like to have @alexkrolick give this a look over too, but I'm 👍

alexkrolick: I think some of the queries initially had different defaults for trim and collapseWhitespace, and they had different enough effects to keep separate.

I'm on the fence about exporting the factory function, but I definitely don't want to encourage copy-pasting the default implementation - that would be opting-out of future changes.

I think this is good as-is and leaves room to come up with future more streamlined APIs for other use cases.

src/matches.js
```diff
@@ -17,16 +13,12 @@ function fuzzyMatches(
   }
 }
 
-function matches(
-  textToMatch,
-  node,
-  matcher,
-  {collapseWhitespace = true, trim = true} = {},
-) {
+function matches(textToMatch, node, matcher, normalizer) {
```
alexkrolick: Is there a way we can add this option without making a breaking change to the matches APIs?

src/index.js
```diff
@@ -6,7 +6,7 @@ export * from './queries'
 export * from './wait'
 export * from './wait-for-element'
 export * from './wait-for-dom-change'
-export * from './matches'
+export {getDefaultNormalizer} from './matches'
```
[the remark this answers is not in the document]
RoystonS: Yes, as I commented in the commit, I removed `matches` and `fuzzyMatches`: they looked like they were exported accidentally, and aren't documented. Are they officially supported?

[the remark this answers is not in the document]
alexkrolick: They were used in jest-dom but it looks like that dependency was removed: 

https://github.com/gnapse/jest-dom/issues/9

https://github.com/gnapse/jest-dom/commit/4c6b572b6ea4e27af4d1173c3b262da2ee466bbd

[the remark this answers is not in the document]
RoystonS: (Actually, looking closer at those links, where were `matches` and `fuzzyMatches` used in `jest-dom`? I can see `getNodeText`, but at a quick skim I can't see `matches` and `fuzzyMatches`. It looks like they had their own implementation of `matches`.  Am I missing something?)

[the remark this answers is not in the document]
alexkrolick: Was referring to this comment https://github.com/gnapse/jest-dom/issues/9#issuecomment-380496381

But you're right it doesn't look like it was used directly.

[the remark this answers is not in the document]
RoystonS: @kentcdodds can you confirm whether this is a breaking change? I'm happy to change the PR so `matches` and `fuzzyMatches` aren't changed, if necessary, but it'd be _preferable_ not to continue to expose those internals or to have to keep the same APIs for them.

kentcdodds:
I hope that you don't mind if we merge this one first: https://github.com/kentcdodds/dom-testing-library/pull/169 😬

src/index.js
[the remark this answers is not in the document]
kentcdodds: I'm fine with this change.

src/matches.js
alexkrolick: I guess we can ignore this if removing the matchers as exports per https://github.com/kentcdodds/dom-testing-library/pull/172#discussion_r240465361

kentcdodds:
Let's wait for the new `getByDisplayValue` to get merged, then update this to account for that one as well, then we can get it merged. Thanks for your help and patience @RoystonS!

RoystonS:
Sure. Will keep an eye on the `getByDisplayValue` and then give that the same `normalizer` treatment.

RoystonS:
I'm merging in the `getByDisplayValue` changes now. I'll push a rebase + squash to keep this PR tidy against a newer master.

PR #181, opened 2018-12-12
fireEvent keyDown not properly documented. 'code' doesn't fireEvent

SeanRParker:
**What**: Documentation for KeyDown fireEvent was using a code key but it needs keyCode

**Why**: Tests using the documented way won't fire the key

**How**: N/A

**Checklist**:

- [ x] Documentation
- [ N/A] Tests
- [x ] Ready to be merged 
- [ x] Added myself to contributors table


README.md
```diff
@@ -674,7 +674,7 @@ fireEvent.change(getByLabelText(/picture/i), {
 **Keyboard events**: There are three event types related to keyboard input - `keyPress`, `keyDown`, and `keyUp`. When firing these you need to reference an element in the DOM and the key you want to fire.
 
 ```javascript
-fireEvent.keyDown(domNode, {key: 'Enter', code: 13})
+fireEvent.keyDown(domNode, {key: 'Enter', keyCode: 13})
```
kentcdodds: Thanks, but `keyCode` is [a deprecated standard](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode).

It should work to pass it if there's code that relies on `keyCode`, but our docs will use the [standard `code` property](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code).

Thanks anyway!

SeanRParker: Fixed my use case to avoid deprecated standard. The tests work now.
Thank you!

SeanRParker: Also, (if it's useful) we didn't have this library at the most current version.

PR #190, opened 2019-01-09
Use innerText rather than textContent for label queries.

ismail-codar:
**What**:
Using label.innerText rather than label.textContent

**Why**:
An example dom: 
```html
<label>
Options
<select>
    <option>Option 1</option>
    <option>Option 2</option>
</select>
</label>
```
textContent returns incorrect result for label text
```text
Options

    Option 1
    Option 2
```

but innerText returns correct result "Options".

But I can not be simulate with jest tests also `label.innerText || label.textContent` used because jsdom does not support innerText https://github.com/jsdom/jsdom/issues/1245

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 

### DIFF
--- a/src/__tests__/elements-inside-label.js
+++ b/src/__tests__/elements-inside-label.js
@@ -1,0 +1,37 @@
+// query utilities:
+import {getByLabelText} from '../'
+import document from './helpers/document'
+
+function getExampleDOM() {
+  const div = document.createElement('div')
+  div.innerHTML = `
+    <label>Cars
+        <select>
+            <option>Volvo</option>
+            <option>Mercedes</option>
+        </select>
+    </label>
+    <label><span>States</span>
+        <select>
+            <option>Damaged</option>
+            <option>Good</option>
+        </select>
+    </label>
+    <label><input type="checkbox" value="1"> I agree to terms </label>
+  `
+  return div
+}
+
+test('elements inside label', () => {
+  const container = getExampleDOM()
+
+  const element1 = getByLabelText(container, 'Cars')
+  expect(element1.tagName).toBe('SELECT')
+
+  // this fails because getNodeText function returns incorrect result on JSDOM but it good for browser
+  // const element2 = getByLabelText(container, 'States');
+  // expect(element2.tagName).toBe('SELECT');
+
+  const element3 = getByLabelText(container, 'I agree to terms')
+  expect(element3.value).toBe('1')
+})
--- a/src/queries.js
+++ b/src/queries.js
@@ -20,7 +20,7 @@ function queryAllLabelsByText(
   const matcher = exact ? matches : fuzzyMatches
   const matchNormalizer = makeNormalizer({collapseWhitespace, trim, normalizer})
   return Array.from(container.querySelectorAll('label')).filter(label =>
-    matcher(label.textContent, label, text, matchNormalizer),
+    matcher(label.innerText || label.textContent, label, text, matchNormalizer),
   )
 }
 
--- a/src/queries.js
+++ b/src/queries.js
@@ -20,7 +20,7 @@ function queryAllLabelsByText(
   const matcher = exact ? matches : fuzzyMatches
   const matchNormalizer = makeNormalizer({collapseWhitespace, trim, normalizer})
   return Array.from(container.querySelectorAll('label')).filter(label =>
-    matcher(label.innerText || label.textContent, label, text, matchNormalizer),
+    matcher(getNodeText(label), label, text, matchNormalizer),
   )
 }
 
--- a/src/get-node-text.js
+++ b/src/get-node-text.js
@@ -2,16 +2,20 @@ function getNodeText(node) {
   const window = node.ownerDocument.defaultView
 
   if (node.matches('input[type=submit], input[type=button]')) {
-    return node.value;
+    return node.value
   }
 
-  return Array.from(node.childNodes)
-    .filter(
-      child =>
-        child.nodeType === window.Node.TEXT_NODE && Boolean(child.textContent),
-    )
-    .map(c => c.textContent)
-    .join('')
+  return (
+    node.innerText ||
+    Array.from(node.childNodes)
+      .filter(
+        child =>
+          child.nodeType === window.Node.TEXT_NODE &&
+          Boolean(child.textContent),
+      )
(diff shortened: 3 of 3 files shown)


kentcdodds:
Interesting. I've never seen a select inside a label like that... 🤔 

I remember writing this code and I remember researching and making the decision to use `textContent` rather than `innerText`. In retrospect I wish I had written a comment about that decision, but I didn't. I also didn't write a test to illustrate the issue that I saw with using `innerText`. I worry that if we merge this we'll break a bunch of tests 😬

kentcdodds:
Ah, the fact that jsdom doesn't support innerText anyway may be the reason I landed with textContent actually 😅 

I'm good with this PR 👍  Thanks for explaining!

src/queries.js
```diff
@@ -20,7 +20,7 @@ function queryAllLabelsByText(
   const matcher = exact ? matches : fuzzyMatches
   const matchNormalizer = makeNormalizer({collapseWhitespace, trim, normalizer})
   return Array.from(container.querySelectorAll('label')).filter(label =>
-    matcher(label.textContent, label, text, matchNormalizer),
+    matcher(label.innerText || label.textContent, label, text, matchNormalizer),
```
kentcdodds: Actually, I think it may be better to:

1. Change this to: `matcher(getNodeText(label), label, text, matchNormalizer),`
2. Update [getNodeText](https://github.com/ismail-codar/dom-testing-library/blob/fab6064602b2ab5bb35d971bf8a200aa57c32f61/src/get-node-text.js) to try `innerText` before `textContent`

This way we get the benefit of using `innerText` for all text selectors.

Could you make that adjustment here please?

alexkrolick:
Does this still work if you have a nested span or something inside the label?

kentcdodds:
> Does this still work if you have a nested span or something inside the label?

image

Yep :+1:

> I think we need to minimally add tests in JSDOM that mock out and test the different behavior.

I think I agree with you, though I'm not sure what you're suggesting.

alexkrolick:
> I think I agree with you, though I'm not sure what you're suggesting.

Something like a stubbing out the .innerText method in the environment or at least for one test: https://github.com/jsdom/jsdom/issues/1245#issuecomment-445848341

Then verifying that form elements (or whatever else is excluded from innerTextification) are omitted. 

**?** Due to the stubbing, this is probably more like a documentation case than something that is actually testing functionality. I guess you could just assert .innerText is called if present.

kentcdodds:
> I guess you could just assert .innerText is called if present.

That's probably the best we could do by way of a test and I'm not sure how valuable that test would be but I'd definitely appreciate having it if @ismail-codar wants to add it.

src/queries.js
```diff
@@ -20,7 +20,7 @@ function queryAllLabelsByText(
   const matcher = exact ? matches : fuzzyMatches
   const matchNormalizer = makeNormalizer({collapseWhitespace, trim, normalizer})
   return Array.from(container.querySelectorAll('label')).filter(label =>
-    matcher(label.textContent, label, text, matchNormalizer),
+    matcher(getNodeText(label), label, text, matchNormalizer),
```
kentcdodds: Don't forget that you'll need to update the `getNodeText` function to use `innerText` :+1:

src/queries.js
ismail-codar: I will sent it with a test.

src/__tests__/elements-inside-label.js
```diff
@@ -0,0 +1,37 @@
...
+
+test('elements inside label', () => {
+  const container = getExampleDOM()
+
+  const element1 = getByLabelText(container, 'Cars')
+  expect(element1.tagName).toBe('SELECT')
+
+  // this fails because getNodeText function returns incorrect result on JSDOM but it good for browser
+  // const element2 = getByLabelText(container, 'States');
+  // expect(element2.tagName).toBe('SELECT');
```
ismail-codar: This lines added temporarily added until switching jest to karma.

kentcdodds: > This lines added temporarily added until switching jest to karma.

Wait.... Are you attempting to migrate our tests to karma? If so then please don't take the time, it will not be merged.

If you want to add a few tests in karma and leave the existing tests as they are that's fine. The bill of the tests should be in jest, but I'd be happy to have a few tests that run in a real browser with karma.

Just want to make sure we're clear on expectations here.

ismail-codar: Ok I understand.

itsravenous:
What needs to happen to get this merged? I would love to be able to use `getByText` to select an element whose text is split within several child nodes.

alexkrolick:
> What needs to happen to get this merged? I would love to be able to use `getByText` to select an element whose text is split within several child nodes.

Not sure. It sounds like there a ton of layout edge cases causing innerText not to be practical in JSDOM. I don't know if working around/semi-polyfilling the behavior is a good idea if it deviates from the spec.

itsravenous-sky:
I think I misunderstood the purpose of this PR. I'm after being able to query for an element with text split over multiple children, e.g:

```
<div>Hi <span>there</span></div>

getByText('Hi there')
```

I'll look into it and raise an issue/PR

kentcdodds:
Ok, so in investigating this, I realized there's a flaw in the original assumption:

> but innerText returns correct result "Options".

That is not correct. This is what I've learned:

```
labelNode.innerText
"Options 
Option 1
Option 2"


labelNode.textContent
"
Options

    Option 1
    Option 2
"
```

So, our current implementation is great because when running this code with `getNodeText` (without changes), we have:

```
getNodeText(labelNode)
"
Options

"
```

Which is correct I believe. So I'll go ahead and make that fix and we should be good to go.

mcmillion:
Is this change why I suddenly have failing tests with `getByLabelText` for labels with nested `spans`? Just started happening today on a new machine with a fresh `yarn install`.

kentcdodds:
Ah yes, it might be. I think there's a solution to this. I'll have to work on it tomorrow. Could you make a new issue?

NehrDani:
> Ah yes, it might be. I think there's a solution to this. I'll have to work on it tomorrow. Could you make a new issue?

I did: #230

Issue #207, opened 2019-02-08
Enhance getByText query


tylerthehaas:
### Describe the feature you'd like:

Currently the getByText takes an HTML element to scope the search to and then a text matcher. This will return an HTML element. The issue I've run into with this is that say I provide the following.

```
const emptyHeader = container.querySelector('.emptyHeader');

expect(getByText(emptyHeader, /createOneOnOne/).toBeInTheDocument();
```

I would get the following error:

```
 Unable to find an element with the text: createOneOnOneTemplate.oneOnOne. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.

    <h3
      class="emptyHeader"
    >
      createOneOnOneTemplate.oneOnOne
    </h3>
```

This is because I've scoped it to the element that I expect the text to appear in so it will search for elements within that h3 that contain the text matching my text matcher. However there is no other elements in this case just a text node. So I can fix this by just getting the parent of the h3 and everything works. There is an issue however if my text node is at the root of the document. Then I can't get a parent. I actually haven't come across that case where I have a text node at the root of the document but it is possible to happen. However I do think that it is a more intuitive api to scope it to the element you expect the text to be be contained in whether that has children that contain the text or not and it would return a boolean value.

### Suggested implementation:

as mentioned above we can scope it to any html element and then look if the text is contained anywhere within the html sub tree that has now been attached to the body of the document. We would then just return a boolean value rather than an html element | null.

### Teachability, Documentation, Adoption, Migration Strategy:

I'm not sure how `.toBeInTheDocument()` is implemented but assuming that it checks for truthiness then the migration should be fairly straight forward. If not then we can think of alternate implementations to enhance how this query works.

kentcdodds:
We should definitely consider the scoped element in the query along with all its children. I'd call that a bug. Anyone want to fix that?

tylerthehaas:
Sorry for the delayed response. I'd love to work on this. Any objection to my proposed implementation?

kentcdodds:
Yeah, so I think the implementation will basically be: "Does this match the scoped element? Yes? Return the scoped element. No? Do whatever we're currently doing"

An elegant way to do this may be to change [this](https://github.com/kentcdodds/dom-testing-library/blob/7e9a72656a080622b107fdc286a11d6a5f9dacfd/src/queries.js#L108):

```js
return Array.from(container.querySelectorAll(selector))
```

To this:

```js
[...]

PR #209, opened 2019-02-13
allow queryAllByText to select text nodes text content

tylerthehaas:
closes issue #207

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -404,6 +404,7 @@ test('queryAllByText can query dom nodes', () => {
   const {queryAllByText} = render('hi')
   expect(queryAllByText('hi')).toHaveLength(1)
   expect(queryAllByText('not here')).toHaveLength(0)
+  expect(queryAllByText('hi', {selector: 'span'})).toHaveLength(0)
 })
 
 test('using jest helpers to assert element states', () => {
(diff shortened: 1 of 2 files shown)


kentcdodds:
@all-contributors please add @tylerthehaas for code and tests

gnapse:
After this was merged, I'm getting `container.matches is not a function` using `cypress-testing-library`. Any ideas why this could be?

kentcdodds:
Hmmmm.. could you figure out what's going on? Maybe we need to just do a typeof check?

gnapse:
I'll try to carve out time to attempt to create a minimal reproduction repo, if no one has a brilliant idea before I can do that.

I initially suspected this `element.matches` thing was not supported by Cypress browser environment (as it happens, for instance, with `EventTarget`) but that seems unlikely, since that element method [seems to have full support across all browsers](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches).

tylerthehaas:
Its interesting that its breaking for the `getByLabelText` query since all that was changed was `queryAllByText` and as far as I can tell that isn't called anywhere within `getByLabelText`.

tylerthehaas:
The issue here is that cypress testing library is passing the document as the container. We should be able to do a check to make sure that `container.matches` is a function and that should solve this.

kentcdodds:
Agreed 👍

cslinmiso:
` TypeError: container.matches is not a function` is happening to me too, while I was passing `window.document` to `getByText` in jest.

valscion:
I've got a working fix for this, with a new test case that would fail. I'll create a PR shortly

PR #213, opened 2019-02-14
Fix regression "X.matches is not a function" in v3.16.6

valscion:
**What**:

I fixed the regression caused by #209 and described here: https://github.com/kentcdodds/dom-testing-library/pull/209#issuecomment-463217859

> After this was merged, I'm getting `container.matches is not a function` using `cypress-testing-library`. Any ideas why this could be?
>
>**Update:** I can confirm it was this change. I edited my package json to force it to resolve `dom-testing-library` to `v3.16.5` (the version prior to this being released) and my tests passed.
>
> My cypress test's use of this library is a mere `cy.getByLabelText('User name').type('john')` (it's a new project and I'm just starting to write tests to it).

**Why**:

I want my Cypress tests to pass with the latest `dom-testing-library` :smile:

**How**:

I first added a failing test case, running in CI, to demonstrate the existence of a regression.

Then the next commit will fix that by testing first that we can use `.matches` function as described here: https://github.com/kentcdodds/dom-testing-library/pull/209#issuecomment-463309607

**Checklist**:

-  Documentation added to the [docs site](https://github.com/alexkrolick/testing-library-docs) N/A
-  Typescript definitions updated N/A
- [x] Tests
- [x] Ready to be merged 


valscion:
Actually #215 seems to highlight yet another issue with this same code branch. I'll see if I can fix that, too

kentcdodds:
Thanks for working on this @valscion! I'll wait until you give me the :+1: before merging :)

valscion:
I think that I'm too confused on #215 currently that I'd feel better if we'd merge this PR to fix the regression raised in #209 and then if this won't resolve #215, I'll create another PR to tackle that.

So let's merge this as it currently stands, just to get the fix for Cypress out the door at least.

kentcdodds:
Thanks!

PR #216, opened 2019-02-15
Fix node.matches is not a function

valscion:
Fixes https://github.com/kentcdodds/dom-testing-library/issues/215

**What**:

Fixes the error caused by loose mode transpilation as spotted by @lavigneer:

> No, there aren't any elements in that `NodeList` that are also a `NodeList`
> 
> I just tested wrapping `container.querySelectorAll(selector)` explicitly in an `Array.from` call in the built `.esm` bundle and that appeared to resolve the issue.
> 
> The transpilation of
> `Array.from([...baseArray, ...container.querySelectorAll(selector)])`
> results in
> `Array.from([].concat(baseArray, container.querySelectorAll(selector)))`
> so it's not converting the `NodeList` to an array and is just trying to `concat` it directly

**Why**:

To fix regression

**How**:

Explicitly use `Array.from` to force the `NodeList` to be converted to an array and spread correctly.

**Checklist**:

- Documentation added to the [docs site](https://github.com/alexkrolick/testing-library-docs) N/A
-  Typescript definitions updated N/A
- Tests N/A can't be tested as bug is in transpilation
- [x] Ready to be merged 

### DIFF
--- a/src/queries.js
+++ b/src/queries.js
@@ -109,7 +109,7 @@ function queryAllByText(
   if (typeof container.matches === 'function' && container.matches(selector)) {
     baseArray = [container]
   }
-  return Array.from([...baseArray, ...Array.from(container.querySelectorAll(selector))])
+  return [...baseArray, ...Array.from(container.querySelectorAll(selector))]
     .filter(node => !ignore || !node.matches(ignore))
     .filter(node => matcher(getNodeText(node), node, text, matchNormalizer))
 }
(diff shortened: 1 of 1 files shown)


src/queries.js
```diff
@@ -109,7 +109,7 @@ function queryAllByText(
   if (typeof container.matches === 'function' && container.matches(selector)) {
     baseArray = [container]
   }
-  return Array.from([...baseArray, ...container.querySelectorAll(selector)])
+  return Array.from([...baseArray, ...Array.from(container.querySelectorAll(selector))])
```
kentcdodds: We don't need the `return Array.from([` that can simply be: `return [`

valscion: True!

PR #217, opened 2019-02-20
Disabled click events

ken-kenware:
**What**: Disabling click events for certain disabled HTML elements

**Why**: This change resolves a bug that was noticed in react-testing-library where a disabled checkbox was still firing the change / input events, even though the click event was not dispatched: https://github.com/kentcdodds/react-testing-library/issues/275

In a browser environment, when a disabled checkbox is manually clicked, the click event is actually thrown away, so the remaining events never get fired. However, when programmatically dispatching a click event, the click event is disables (as per the spec), but the remaining events fire.

So this change brings the mock event behavior closer to what the browser is actually doing.

**How**: Override our click event handler, check if the element is disabled and if so do not fire the event.

**Checklist**:

- [x] Tests
- [ ] Ready to be merged 

I'm not convinced that my approach for overloading the click event dispatch is right. But I wanted to get something out to figure out the right way to do it. The only constraint is that this should only impact the click event, other events are not changed.


sompylasar:
> I'm thinking that it's probably best if people's tests just don't try to fire a click event on something that's disabled. Why is that happening?

What if the test does not know if the element is disabled? The user wouldn't know, they might click anywhere at any time, and the app should behave appropriately.

alexkrolick:
> What impact will this have for something like this:
> 
> `<button disabled="true"><span>Hello</span></button>`

Seems like this needs to be in JSDOM. They definitely have code related to this: https://github.com/jsdom/jsdom/pull/1952/files

ken-kenware:
My gut also pointed me to JSDom as well. But I did some digging, and JSDom is doing exactly what the browser does when you emit a custom event (vs a natural click event) see: https://codepen.io/kdeland/full/daLBZm

When you click on the checkbox, nothing happens. But if you dispatch the click event (via clicking the button here), then you see the other events are dispatched as well.

So in this case, I think JSDom is doing the right thing. I think what's happening is the test framework isn't behaving as expected due to this weird quirk...

I'm not super committed to that position though and I could be persuaded otherwise.

To your point @kentcdodds I need to change the detection to make sure we don't block on clicking a valid element and having it nested in a disabled element. But I'll hold off on that change until we can decide if this is a change we want to make!

Thanks for the discussion

kentcdodds:
I'm thinking that maybe this should be implemented in something else (like [`user-event`](https://testing-library.com/docs/ecosystem-user-event)).

ken-kenware:
Sure! will close and investigate over there :)

Gpx:
FYI this has now been fixed in `user-event`

Issue #206, opened 2019-02-07
waitForElementToDisappear


mastertinner:
### Describe the feature you'd like:

I love the `waitForElement` function. However, there are cases where I don't add an element but rather remove existing ones. Is there a similar `waitForElementToDisappear` function or is there an easy way to achieve that using the existing functions?

### Suggested implementation:

```javascript
await waitForElementToDisappear(() => getByText('my button'))
...
```

Currently I'm using `process.nextTick` but that seems like a hack to me:

```
process.nextTick(() => {
    ...
    done()
})
```

kentcdodds:
Forgot about that! Done :)

Now let's discuss.

So you can do this today with:

```js
await wait(() => expect(getByText('my button')).not.toBeInTheDocument())
```

But that's kinda awkward and having a `waitForElementToBeRemoved` or something similar would be useful. Just not sure of the best name.

mastertinner:
In the meantime, I found out that `waitForDomChange` actually covers most of my use cases. For explicitly wait for the removal of an element, I think the name you suggested (`waitForElementToBeRemoved`) is quite a good one. It's a little verbose, but I think that's worth having the name be explicit about what it does.

kentcdodds:
I'm into it. I say let's do it.

lgandecki:
@kentcdodds I think it would have to be queryByText? Otherwise the getByText itself will fail when the element goes away, and the expectation won't be checked.
Also, I wonder if the awkwardness here is something you feel very strongly? I don't personally see a problem with that wait (that's +/- how I've seen this getting done in "selenium era"). Maybe a documentation change would be a way to go instead of increasing the API surface?

kentcdodds:
No worries @lgandecki, thanks for providing your perspective. You make good points. I wonder though if there could be several document changes before that thing is actually removed. Like a progress bar for example.

It's true that this is really not a very typical thing and the workaround isn't that challenging (you're right, would need to be a `queryBy*`). Hmmm...

gabriel403:
@kentcdodds @lgandecki I stumbled upon this issue and myself today, I need to check a progress loader has finished, as the main list element to check is always in the dom, I could check individual items within the list element but wanted to give this a try and then use snapshots for asserting that the items are there.
I tried 
```
      await wait(() => expect(getByTestId('progress-loader')).not.toBeInTheDocument());
```
but as stated that fails as the getByTestId fails when the loader is done.
I also tried 
```
[...]

PR #218, opened 2019-02-22
Wait for element to be removed

Tolsee:
**What**: Adding `waitForElementToBeRemoved` feature.

**Why**: As discussed in #206, it would be easier to use `waitForElementToBeRemoved` to check whether the element is removed or not. 

**How**: It is similar to `waitForElement`.  Can be used as `waitForElementToBeRemoved(() => getByTestId('data-testid'))`, actually just like `waitForElement` with any other query functions.

**Checklist**:

- [x] Tests
- [x] Ready to be merged 


src/__tests__/wait-for-element-to-be-removed.js
```diff
@@ -0,0 +1,278 @@
...
+import 'jest-dom/extend-expect'
+import {render} from './helpers/test-utils'
+import document from './helpers/document'
+
+const skipSomeTime = delayMs =>
+  new Promise(resolve => setTimeout(resolve, delayMs))
+
+// Using `setTimeout` >30ms instead of `wait` here because `mutationobserver-shim` uses `setTimeout` ~30ms.
+const skipSomeTimeForMutationObserver = (delayMs = 50) =>
+  skipSomeTime(delayMs, 50)
```
kentcdodds: Do you think we could use jest fake timers for this? I've found that tests that rely on setTimeout are pretty unreliable in CI environments especially.

kentcdodds:
I think that's a good idea as well :+1:

src/__tests__/wait-for-element-to-be-removed.js
Tolsee: Sounds good, I will give a shot.

Tolsee: I tried to use `fakeTimer`, but it does not seem to be an ideal place to use timeFakers.

As per I know, `setTimeout` are unreliable because of time-outs in CI right? How important do you think it is for this PR. If it's pretty important, I will try to look at it in more detail.

Tolsee:
> Should it succeed or fail if the element did not exist in the first place? It may be confusing, "ToBeRemoved" imho assumes just the change from "is present" to "is not present".

I have changed it to throw `Error` when there is no element at starting.

src/__tests__/wait-for-element-to-be-removed.js
kentcdodds: I just don't want tests failing randomly on CI. And that happens when relying on setTimeout quite a bit.

Tolsee: I have changed the tests. Can you please take a look.

src/__tests__/wait-for-element-to-be-removed.js
```diff
@@ -73,9 +74,11 @@ test('it waits for the callback to throw error or a falsy value and only reacts
   for (const [mutationImpl, callbackImpl] of mutationsAndCallbacks) {
     callback.mockImplementation(callbackImpl)
     mutationImpl()
-    await skipSomeTimeForMutationObserver() // eslint-disable-line no-await-in-loop
+    skipSomeTimeForMutationObserver() // eslint-disable-line no-await-in-loop
```
alexkrolick: Doesn't this still need the await?

src/wait-for-element-to-be-removed.js
```diff
@@ -0,0 +1,66 @@
...
+      childList: true,
+      attributes: true,
+      characterData: true,
+    },
+  } = {},
+) {
+  return new Promise((resolve, reject) => {
+    if (typeof callback !== 'function') {
+      reject(
+        'waitForElementToBeRemoved requires a callback as the first parameter',
```
alexkrolick: Nit: requires a function. Callback is jargon.

src/__tests__/wait-for-element-to-be-removed.js
kentcdodds: I don't think so. Now the test runs synchronously because we're using jest fake timers.

src/wait-for-element-to-be-removed.js
```diff
@@ -0,0 +1,66 @@
...
+      )
+    }
+
+    // Check if the element is not present synchronously,
+    // As the name waitForElementToBeRemoved should check `present` --> `removed`
+    ;(function checkElementPresent() {
```
kentcdodds: Is there a reason this code needs to be in an IIFE?

Tolsee: Not much, I was just hoping to separate functionality.

Tolsee: Will it have a bad impact?

kentcdodds: I guess not. Just not a coding style we have anywhere else in the codebase. Feels unnecessary to me.

Tolsee: Ok, will get rid of it.

src/wait-for-element-to-be-removed.js
```diff
@@ -0,0 +1,64 @@
...
+      childList: true,
+      attributes: true,
+      characterData: true,
+    },
+  } = {},
+) {
+  return new Promise((resolve, reject) => {
+    if (typeof callback !== 'function') {
+      reject(
+        'waitForElementToBeRemoved requires a function as the first parameter',
```
kentcdodds: I think this should be an error object rather than just a string so we get a stacktrace.

src/wait-for-element-to-be-removed.js
```diff
@@ -0,0 +1,64 @@
...
+      )
+    }
+
+    // Check if the element is not present synchronously,
+    // As the name waitForElementToBeRemoved should check `present` --> `removed`
+    try {
+      const result = callback()
+      if (!result) {
```
kentcdodds: Should we be more specific in checking that the result is a single element (or non-empty array of elements)? If it's not then the error can be more specific as well: "The callback function which was passed did not return an element or non-empty array of elements. waitForElementToBeRemoved requires that the element(s) exist before waiting for removal."

src/wait-for-element-to-be-removed.js
```diff
@@ -0,0 +1,64 @@
...
+        onDone(new Error('Element is not present in the DOM.'), true)
+      }
+    } catch (error) {
+      onDone(new Error('Element is not present in the DOM.'), true)
```
kentcdodds: I think this should be `onDone(error)` so we don't swallow the error. Also, what's the purpose of `true` as the second argument here?

src/wait-for-element-to-be-removed.js
```diff
@@ -0,0 +1,64 @@
...
+        'waitForElementToBeRemoved requires a function as the first parameter',
+      )
+    }
+
+    // Check if the element is not present synchronously,
+    // As the name waitForElementToBeRemoved should check `present` --> `removed`
+    try {
+      const result = callback()
+      if (!result) {
+        onDone(new Error('Element is not present in the DOM.'), true)
```
kentcdodds: What's the purpose of the second argument of `true` here?

src/wait-for-element-to-be-removed.js
Tolsee: @ken Won't it be confusing for users? by doing `onDone(error)`? Because then they would not know the error is telling them that `waitForElementToBeRemoved` requires that the element(s) exist before waiting for removal.

Tolsee: The perfect solution would have been to `instanceof` check for the `GetElementError`. But we do not have our own error class right?

I will just send back the error for now. But, making our own error class and checking for that, if the error is `GetElementError` send the above message else sending back the caught error itself seems a good solution to me.

src/__tests__/__snapshots__/wait-for-element-to-be-removed.js.snap
```diff
@@ -0,0 +1,51 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
kentcdodds: I don't think these snapshots are at all necessary and just make it harder to know the intent of the tests. Could we change these to be regular assertions instead?

src/__tests__/__snapshots__/wait-for-element-to-be-removed.js.snap
```diff
@@ -0,0 +1,51 @@
...
+
+exports[`it returns error immediately if there callback returns falsy value or error before any mutations 1`] = `
+Array [
+  [ReferenceError: timer is not defined],
+]
+`;
+
+exports[`it returns error immediately if there callback returns falsy value or error before any mutations 2`] = `
+Array [
+  [ReferenceError: timer is not defined],
```
kentcdodds: This error is actually being thrown because we call `onDone` before the `timer` is defined. Almost missed that because we were using snapshots and the tests are separated from the assertions...

src/wait-for-element-to-be-removed.js
```diff
@@ -0,0 +1,70 @@
...
+    } catch (error) {
+      onDone(error)
+    }
+
+    const timer = setTimeout(onTimeout, timeout)
+    const observer = newMutationObserver(onMutation)
+    observer.observe(container, mutationObserverOptions)
+    function onDone(error, result) {
+      const setImmediate = getSetImmediate()
+      clearTimeout(timer)
```
kentcdodds: `onDone` can be called before `timer` is even defined, which results in an error. Make sure to define `timer` earlier in the function.

src/__tests__/__snapshots__/wait-for-element-to-be-removed.js.snap
Tolsee: @kentcdodds I have almost removed the snapshot to inline snapshots(Which should be working as regular assertions). But, because we are also passing the error we got from the `getByTestId` etc, in that case, a snapshot will be a good thing I guess.

Tolsee:
@kentcdodds I am not sure why the test is failing(for `node`, `8.0`).

I tried to simulate it with same version of `nvm`, `node` and `npm` in my machine and it's working fine.

src/__tests__/wait-for-element-to-be-removed.js
```diff
@@ -0,0 +1,344 @@
...
+  expect(callback).toHaveBeenCalledTimes(1)
+  expect(successHandler).toHaveBeenCalledTimes(0)
+  expect(errorHandler).toHaveBeenCalledTimes(0)
+
+  container.querySelector('div').innerHTML = 'new text'
+  skipSomeTimeForMutationObserver()
+  await promise
+
+  expect(successHandler).toHaveBeenCalledTimes(1)
+  expect(successHandler.mock.calls[0]).toMatchInlineSnapshot(`
```
kentcdodds: Any reason this couldn't be simplified to:

```javascript
expect(successHandler).toHaveBeenCalledWith(true)
```

Or you could also replace this line and the one above with:

```javascript
expect(successHandler.mock.calls).toEqual([true])
```

I don't really care either way, but using snapshots for this doesn't make any sense to me personally.

src/__tests__/wait-for-element-to-be-removed.js
```diff
@@ -0,0 +1,344 @@
...
+  expect(errorHandler).toHaveBeenCalledTimes(3)
+
+  container.setAttribute('data-test-attribute', 'something changed once')
+  skipSomeTimeForMutationObserver(50)
+
+  expect(callbackForError).toHaveBeenCalledTimes(1)
+  expect(callbackForFalsy).toHaveBeenCalledTimes(1)
+  expect(successHandler).toHaveBeenCalledTimes(0)
+  expect(errorHandler).toHaveBeenCalledTimes(3)
+  expect(errorHandler.mock.calls[0]).toMatchSnapshot()
```
kentcdodds: I guess I'm fine with this being a snapshot, but please make it inline.

src/__tests__/wait-for-element-to-be-removed.js
Tolsee: Yeah, make sense. I am a little  🆕 to `jest` that is the only reason.

Tolsee:
> Thanks! We're getting closer!

One of my first contribution to this library as well as open source. Really enjoying the journey.

src/__tests__/wait-for-element-to-be-removed.js
```diff
@@ -0,0 +1,334 @@
+import {waitForElementToBeRemoved} from '../'
+// adds special assertions like toBeTruthy
```
kentcdodds: toBeTruthy is built-into jest. This adds things like toBeInTheDocument. But I don't think we need this comment anyway. Please remove it.

Tolsee:
> After looking closer at the tests I can't help but feel like they could be simplified. I hope you don't mind me working on that a little on my own.

Of course not. But, I would like to look at the implementation after you have done it.

Tolsee:
@kentcdodds Did you get a chance to review the tests? If you do not have enough time, you can give me a direction and I will work on the test. I want this to be merged 😄

kentcdodds:
Sorry, been busy. I'll try to get this done this weekend.

kentcdodds:
Alrighty. I'm happy with these tests now. Once the build passes then I'll merge/release it :D

jest.config.js
```diff
@@ -1,20 +1,6 @@
...
-  projects: [
-    {
-      ...baseConfig,
-      displayName: 'jsdom',
-      testEnvironment: 'jest-environment-jsdom',
-    },
-    {
-      ...baseConfig,
-      displayName: 'node',
-      testEnvironment: 'jest-environment-node',
```
kentcdodds: Honestly, running all the exact same tests in `node` led to some poor testing practices and I didn't like the over-testing either. If someone really wants to have tests like this, then we can add a `other/node/__tests__/` directory with a `other/node/jest.config.js` file. I want the tests to be different.

Also, I don't think it's worthwhile to have tests for node. I'm pretty sure there's a very small number of people who benefit from this.

.travis.yml
```diff
@@ -6,7 +6,6 @@ cache:
 notifications:
   email: false
 node_js:
-  - 'node'
```
kentcdodds: Doing this opens us up to stuff like this: https://stackoverflow.com/questions/55059748/travis-jest-typeerror-cannot-assign-to-read-only-property-symbolsymbol-tostr

I think the benefit of testing an unstable version of node isn't worth stuff like this.

PR #222, opened 2019-03-12
fix(events): fix firing events in document

### LINKED ISSUE #159: Unable to fire event on document
- `react-testing-library` version: 5.3.0
- `react` version: 1.6.6.3
- `node` version: 8.11.3
- `yarn` version: 1.12.3

### Relevant code or config:

```js
fireEvent.keyDown(document.body, {
  key: "Escape",
  keyCode: 27,
  which: 27
});
```

### What happened:

TypeError: Cannot read property 'defaultView' of null

### Suggested solution:

https://github.com/kentcdodds/dom-testing-library/blob/master/src/events.js#L324
Should handle document and window elements (possibly in more places)

tnunes:
**What**:

Fix firing events in the document element.

Fixes #159, as per @kentcdodds suggestion in https://github.com/kentcdodds/dom-testing-library/issues/159#issuecomment-453369641.

**Why**:

Currently firing events in the document element throws a `TypeError: Cannot read property 'defaultView' of null` when trying to use `fireEvent` "shortcurt API" (not sure this is its name :)):

```fireEvent.keyDown(document, { key: 'Escape' })```

It was possible to fire events in the document by using `fireEvent` directly, like so:

```fireEvent(document, new KeyboardEvent('keydown', { key: 'Escape' }))```

**How**:

Falling back to `node` to get `defaultView` from, when `node.ownerDocument` is `null`.

**Checklist**:

- [x] Tests
- [x] Ready to be merged 


tnunes:
Thanks! When can we expect a new release of `react-testing-library` with this dependency bumped?

alexkrolick:
You should be able to do `yarn upgrade react-testing-library` to immediately get the new version because of the `^` dependency requirement.

tnunes:
@alexkrolick you're right! Thanks again for the quick merge!

PR #224, opened 2019-03-19
feat(findBy*): adds findBy* query type allowing for async element queries

### LINKED ISSUE #203: Add async/retryable `findBy*` and `findAllBy*` version of the queries

@alexkrolick has brought this up before and we ultimately decided to not go for it at the time, but I've been thinking about it more and I think it's time to revisit the decision.

There are some UI frameworks that are completely async. In my TestingJavaScript.com course when I show how to use dom-testing-library with these, I had to create a [`fire-event-async.js`](https://github.com/kentcdodds/dom-testing-library-with-anything/blob/9361a120bc52334968e94a10363bab9724d5dbd3/fire-event-async.js) file to deal with those, but that wasn't enough for [mithril's tests](https://github.com/kentcdodds/dom-testing-library-with-anything/blob/master/mithril.test.js) because mithril does not re-render on the next tick all the time (for who knows what reason 🤷‍♂️).

The `findBy*` and `findAllBy*` variants of the queries would return promises that resolve when the query is satisfied, or timeout after 3500ms or so.

### Suggested implementation:

I'm thinking it'd be something like this:

```javascript
const findByText = (...args) => waitForElement(() => getByText(...args))
const findAllByText = (...args) => waitForElement(() => getAllByText(...args))
```

So far the alternative is to have people write this themselves:

```javascript
const successMessage = await waitForElement(() => getByText(/.*success.*/i))
expect(successMessage).toHaveTextContent('Welcome back Nancy')
```

That's worked pretty well most of the time, but I think it would be nicer to write that like this:

```javascript
const successMessage = await findByText(/.*success.*/i)
expect(successMessage).toHaveTextContent('Welcome back Nancy')
```

We do have query explosion going on and I think the docs could be improved slightly around this to document first the types of queries (label text, text, placeholder, etc), and then the variants (`getBy*`, `queryAllBy*`, etc.).

This would be a new feature, so no breaking change/migration strategy necessary.

kentcdodds:
**What**: add findBy* query type

**Why**: Closes #203

**How**: Created a simple `makeFinder` function and made a finder for every query type based on the getBy queries

**Checklist**:

- [x] Tests
- [x] Ready to be merged  (we can add TS and docs later I think).

### DIFF
--- a/src/__tests__/queries.find.js
+++ b/src/__tests__/queries.find.js
@@ -1,0 +1,119 @@
+import {render} from './helpers/test-utils'
+
+test('find asynchronously finds elements', async () => {
+  const {
+    findByLabelText,
+    findAllByLabelText,
+
+    findByPlaceholderText,
+    findAllByPlaceholderText,
+
+    findByText,
+    findAllByText,
+
+    findByAltText,
+    findAllByAltText,
+
+    findByTitle,
+    findAllByTitle,
+
+    findByDisplayValue,
+    findAllByDisplayValue,
+
+    findByRole,
+    findAllByRole,
+
+    findByTestId,
+    findAllByTestId,
+  } = render(`
+    <div>
+      <div data-testid="test-id" aria-label="test-label">test text content</div>
+      <select><option>display value</option></select>
+      <input placeholder="placeholder" />
+      <img alt="test alt text" src="/lucy-ricardo.png" />
+      <span title="test title" />
+      <div role="dialog"></div>
+    </div>
+  `)
+  await expect(findByLabelText('test-label')).resolves.toBeTruthy()
+  await expect(findAllByLabelText('test-label')).resolves.toHaveLength(1)
+
+  await expect(findByPlaceholderText('placeholder')).resolves.toBeTruthy()
+  await expect(findAllByPlaceholderText('placeholder')).resolves.toHaveLength(1)
+
+  await expect(findByText('test text content')).resolves.toBeTruthy()
+  await expect(findAllByText('test text content')).resolves.toHaveLength(1)
+
+  await expect(findByAltText('test alt text')).resolves.toBeTruthy()
+  await expect(findAllByAltText('test alt text')).resolves.toHaveLength(1)
+
+  await expect(findByTitle('test title')).resolves.toBeTruthy()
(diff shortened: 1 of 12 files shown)


src/queries.js
```diff
@@ -375,6 +376,38 @@ function getByDisplayValue(...args) {
   return firstResultOrNull(getAllByDisplayValue, ...args)
 }
 
+function makeFinder(getter) {
```
alexkrolick: how about a comment here like

```suggestion
// wraps the get query with an awaitable
// version that retries until a timeout
function makeFinder(getter) {
```

src/query-helpers.js
```diff
@@ -32,6 +32,7 @@ function getElementError(message, container) {
 function firstResultOrNull(queryFunction, ...args) {
```
alexkrolick: should this be renamed to something like "onlyResultOrNull" since it is more strict that the length has to be 1 now?

src/queries.js
kentcdodds: Excellent 👍

src/query-helpers.js
kentcdodds: Whoops, I didn't mean to change that yet 😬

src/__tests__/queries.find.js
```diff
@@ -0,0 +1,119 @@
+import {render} from './helpers/test-utils'
+
+test('find asynchronously finds elements', async () => {
+  const {
```
alexkrolick: would it be helpful to add some asynchronicity to the tests, or is that covered by unit tests for waitForElement?

```js
const {/* queries * /, rerender} = render(`<div />`)
setTimeout(() => rerender(`<div>correct dom</div>`), 100)
//  wait IRL if not using fake timers, or something like:
const findPromises = Promise.all([expect().toBeTruthy()])
jest.runAllTimers()
await findPromises
```

kentcdodds: I think I want to rely on `waitForElement` for these things. It's a pretty simple abstraction on top of `waitForElement` so I think we're probably good as they are.

That said, I'll go ahead and add one test just to be sure. Can't hurt.

kentcdodds:
@all-contributors please add @alecook for review and doc

alexkrolick:
Sweet

kentcdodds:
I'm working on docs now.

PR #229, opened 2019-03-19
fix(getBy*): throw an error if more than one element is found

### LINKED ISSUE #202: Throw an error if `getBy*` queries return anything other than one result.

As a random thought, what if we throw an error if you use `getBy*` query and there's more than 1 element that could be returned for the given query? We already do that if there are no elements returned.

I'm not certain how I feel about it, but it seems like a good idea. I had the idea when responding to #201 

As it occurs to me, one very common mistake when using CodeSandbox is people render the thing they're testing in the document as well as run tests with it. The problem is that CodeSandbox does not isolate the tests' DOM from the app's DOM, so you often wind up interacting with the element that's rendered in the app rather than the one you're testing. Sometimes it happens to work, but much of the time it doesn't. If we implemented this, then people would get an error and it may help them know why.

### Suggested implementation:

Take this:

https://github.com/kentcdodds/dom-testing-library/blob/369d17e169e5c7f47688482b8f0ad8ecda2ee35f/src/query-helpers.js#L32-L36

Change it to:

```javascript
function resultOrNull(queryFunction, ...args) {
  const result = queryFunction(...args)
  if (result.length === 1) return result[0]
  return null
}
```

Currently we allow this:

```javascript
// Given: <div>Hello</div><div>Hello2</div>
getByText(/.*hello.*/i) // what should this return?
```

Right now we'll get the first, but that could be undesirable and even surprising.

This would require a breaking change, some slight changes to the docs, and the error message could help with teachability. As for migration, hopefully most people only have one result already, if they don't it may be a little surprise but should hopefully be pretty easy to fix by either improving the TextMatch to only match one element, or switching to the `getAllBy*` version of the query and grabbing the first result.

kentcdodds:
**What**: fix(getBy*): throw an error if more than one element is found

**Why**: Closes #202

TL;DR: more explicit, more better.

**How**: Added some logic for the `getBy` queries to throw if the returned elements has length > 1

**Checklist**:

- [x] Tests
- [x] Ready to be merged 

BREAKING CHANGE: All `getBy` and `findBy` query variants now will throw an error if more than one element is returned. If this is expected, then use `getAllBy` (or `findAllBy`) instead.


kentcdodds:
Note: this PR is to the `next` branch because it's a breaking change.

I can't think of any other breaking changes I want to make at this time, so once this is merged, we'll go ahead and merge the `next` branch as well.

src/queries.js
```diff
@@ -136,7 +137,6 @@ function queryAllByTitle(
 function queryByTitle(...args) {
   return firstResultOrNull(queryAllByTitle, ...args)
```
alexkrolick: So the query APIs would still return the first one even if there are multiple?

src/query-helpers.js
```diff
@@ -29,6 +29,13 @@ function getElementError(message, container) {
   return new Error([message, debugDOM(container)].filter(Boolean).join('\n\n'))
 }
 
+function getGetByElementError(message, container) {
+  return getElementError(
+    `${message}\n\n(If this is intentional, then use the \`*AllBy*\` variant of the query (like \`getAllByText\` or \`findAllByText\`)).`,
```
alexkrolick: This error message is nice and clear and makes the breaking change pretty safe

src/queries.js
kentcdodds: Correct. Hmmm... Should we throw for this as well? Probably huh...

alexkrolick: Not sure... querySelector vs querySelectorAll would just return the first item

alexkrolick: I could see it throwing as well. Maybe to clarify it all we could make a table:

|            | One Match | Multiple Matches | No Match | Async? |
| ---------- | --------- | ---------------- | -------- | ------ |
| getBy      | Return    | Throw            | Throw    | No     |
| findBy     | Return    | Throw            | Throw    | Yes     |
| queryBy    | Return    | Throw            | Null    | No     |
| getAllBy   | Array     | Array            | Throw    | No     |
| findAllBy  | Return    | Throw            | Throw    | Yes     |
| queryAllBy | Array     | Array            | Array    | No     |

src/query-helpers.js
wyze: Might be able to enhance it even more with another parameter like `query`, which contains the suffix, `AllByText`. Then this could be updated to:

```
`${message}\n\n(If this is intentional, then use the \`get${query}\` or \`find${query}\` variant of the query).`
```

src/queries.js
bcarroll22: In this model does it make sense to have `query` and `queryAll`? Should there just be:

`query`
`get`
`getAll`
`find`
`findAll`?

To me that seems pretty consistent with like, the type of behavior you'd expect from a REST API for example. like `/api/users` is going to return a list of users even if there's only 1 (or 0). But if i say `/api/users/1` or `/api/users/1,2` i'm asking for specific things and I'll get a 404 if they're not found.

alexkrolick: The current API is based on querySelector/querySelectorAll

https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector

> The Document method querySelector() returns the first Element within the document that matches the specified selector, or group of selectors. If no matches are found, null is returned.

https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll

> A non-live NodeList containing one Element object for each element that matches at least one of the specified selectors or an empty NodeList in case of no matches.

bcarroll22: 👌 makes sense now, thanks for that explanation. Does querySelector throw if more than 1 is found?

alexkrolick: > Does querySelector throw if more than 1 is found?

No, it returns the first element (see above).

bcarroll22: 🙄 duh it was the first sentence, nvm i can't read apparently.

bcarroll22: I feel like I'd like to see query continue to never throw. I think it's nice to have a set of searches that you know aren't going to throw in any circumstance, but then again it might catch some bugs people don't know about currently... I dunno 🤔

Do you see any chance that throwing might make it more confusing for knowing when to use query vs. get since they both have the potential to throw in this model?

kentcdodds: If you think there could be more than one then use queryAll*

alexkrolick: The primary use case for `queryBy` is for when you think/expect something not to exist and you want to check it returns null. The only I can think of it being disruptive to change it would be if you were doing something with `.not`, like `wait(() => expect(queryByText('something')).not.toBeNull()`.

bcarroll22: sounds good to me, i can dig it. we can make this happen in native too 👍

bcarroll22:
I’ve pretty much got this ready in NTL, just waiting to merge/release until I know whether or not queryBy should throw as well 👍

alexkrolick:
@kentcdodds how about releasing a beta release with queryBy throwing, then we can test it out on some codebases and see if it breaks a lot.

kentcdodds:
I can do that 👍

kentcdodds:
Sorry it's taking so long. I finally decided to look at this thing through the lens of abstraction (https://kcd.im/aha) and I think I've figured out a good way to avoid copy/paste issues and hopefully make abstraction worthwhile and make this easier to maintain. We'll see...

kentcdodds:
Ok, feel free to look at the PR. It's kinda a big diff now, but I think it'll be a good thing for the codebase. I'd love to hear what people think.

I've also published an alpha:

```
yarn add --dev dom-testing-library@alpha
```

Let me know what you think!

alexkrolick:
@kentcdodds the makeXQueries are an interesting abstraction - do you think you could simplify it even more to something like this:

```js
const [getByText, getAllByText, findByText, findAllByText, queryByText]
  = buildQueries(queryAllByText, e => 'single error', e => 'multiple error')
```

kentcdodds:
I had considered that as well. Maybe I'll give it a try 👍 I like what you've got there.

kentcdodds:
I've just published an alpha of react-testing-library that upgrades dom-testing-library to the alpha as well:

```
yarn add --dev react-testing-library@alpha
```

kentcdodds:
I just verified that my workshop code all still works with these changes so that's great :)

kentcdodds:
Great idea on the `buildQueries` thing @alecook 👍 I like it.

I'm a tiny bit worried this is over-abstraction. But I _think_ it'll be better this way. I think we've had this copy/pasted code long enough to know what the right abstraction is here.

alexkrolick:
> I'm a tiny bit worried this is over-abstraction. But I think it'll be better this way. I think we've had this copy/pasted code long enough to know what the right abstraction is here.

I think it's fairly safe to assume can can continue abstracting the query types off of a base query- I considered it even back last year when adding the `queryAll` functions for the first time. It'll make it a lot simpler for people to build custom queries that implement all the query types.

kentcdodds:
That's a good point!

bcarroll22:
Man. Oh. Man.

This is a good stuff. I think I need to digest it a bit more, but I believe this even helps us in native be more explicit and make the tests more closely resemble the way the platform works. To me it's a good balance, it abstracts where it makes sense, but it's also a good example of not abstracting when it makes sense not to.

This is awesome, I like it a lot.

bcarroll22:
Question. 

When you were doing this, how did you decide when/when not to use the abstractions vs writing a custom implementation? For example, you limit what elements are queried for `display-value` but not for placeholder. Was there a reason for that?

kentcdodds:
I dunno 🤷‍♂️

kentcdodds:
Ok, I'm ready to make this happen :)

PR #236, opened 2019-04-05
feat: add asyncWrapper config

kentcdodds:
**What**: add asyncWrapper config

**Why**: At first glance you're probably thinking: "Oh no, no no. No way." But think for a moment that react-testing-library can't just export its own `wait*` APIs and make everything work, because the `findBy*` queries all use the `waitForElement` API that's exposed by dom-testing-library (not the one that would be wrapped by react-testing-library), so we have two options:

1. Write wrapping code for every `findBy*` query that exists today and keep that updated in the future if new queries are ever added. This would include wrapping `within`/`getQueriesForElement` so the `findBy*` queries from those are are wrapped. In addition to wrapping all the `wait*` utilities.
2. Allow configuration of the dom-testing-library `wait*` APIs as this PR does

I don't like either of those solutions, but I dislike solution 2 WAY less.

**How**:

Add a config option called `asyncWrapper` and wrap all the `wait*` APIs in that. The default just calls the callback so there's no real cost to doing it this way. I've tested this locally with and without my changes to react-testing-library and it works great (so people using a new dom-testing-library and an old react-testing-library will be fine). When I update react-testing-library I'll make sure to add a version bump to dom-testing-library so we don't wind up with the opposite situation.

**Checklist**:

- [x] Documentation added to the [docs site](https://github.com/alexkrolick/testing-library-docs) N/A
- [x] Typescript definitions updated (I think so?)
- [x] Tests N/A
- [x] Ready to be merged

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -3,6 +3,14 @@
 // './queries' are query functions.
 let config = {
   testIdAttribute: 'data-testid',
+  // this is to support React's async `act` function.
+  // forcing react-testing-library to wrap all async functions would've been
+  // a total nightmare (consider wrapping every findBy* query and then also
+  // updating `within` so those would be wrapped too. Total nightmare).
+  // so we have this config option that's really only intended for
+  // react-testing-library to use. For that reason, this feature will remain
+  // undocumented.
+  asyncWrapper: cb => cb(),
 }
 
 export function configure(newConfig) {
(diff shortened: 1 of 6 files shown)


kentcdodds:
Note, the diff is worse than this really is. Checkout the changes in `wait.js` to see what's really going on in the others.

alexkrolick:
I think this has benefitd for another reason - MutationObserver isn't always available. It can be used to replace the waitForElement based on MutationObserver with a simpler poll or timeout - I did that in the original async wrapper lib.

PR #246, opened 2019-04-17
Fix typings of getQueriesForElement

msimulcik:
**What**:
Fix typings of getQueriesForElement function

**Why**:
There are currently two problems:
1. Type of `queriesToBind` parameter is `BoundFunctions<typeof queries> | BoundFunctions<typeof queries>[]` which is IMO incorrect because the function is doing the binding and thus should receive unbound queries.
1. Type parameter of `BoundFunctions` is hardcoded to `typeof queries` which prevents binding of custom queries

**How**:
1. Parametrised `getQueriesForElement` to enable binding custom queries
1. Changed type of `queriesToBind` to type parameter

**Checklist**:

- [ ] Tests N/A
- [x] Ready to be merged

### DIFF
--- a/typings/get-queries-for-element.d.ts
+++ b/typings/get-queries-for-element.d.ts
@@ -13,9 +13,7 @@ export type BoundFunction<T> = T extends (
   : never
 export type BoundFunctions<T> = {[P in keyof T]: BoundFunction<T[P]>}
 
-export function getQueriesForElement(
+export function getQueriesForElement<T = typeof queries>(
   element: HTMLElement,
-  queriesToBind?:
-    | BoundFunctions<typeof queries>
-    | BoundFunctions<typeof queries>[],
-): BoundFunctions<typeof queries>
+  queriesToBind?: T,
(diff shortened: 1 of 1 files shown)


msimulcik:
@kentcdodds I've just realised that maybe it would be better to make parameter type more specific. See code below. Those are types copied from [react-testing-library typings](https://github.com/kentcdodds/react-testing-library/blob/master/typings/index.d.ts#L5-L17). What do you think?

```
interface Query extends Function {
  (container: HTMLElement, ...args: any[]):
    | Error
    | Promise<HTMLElement[]>
    | Promise<HTMLElement>
    | HTMLElement[]
    | HTMLElement
    | null
}

interface Queries {
  [T: string]: Query
}

export function getQueriesForElement<T extends Queries = typeof queries>(
  element: HTMLElement,
  queriesToBind?: T,
): BoundFunctions<T>
```

kentcdodds:
Ah, sure. Feel free to make another PR :)

Issue #249, opened 2019-04-23
Support `debug` inside `within` API


renerbaffa:
### Describe the feature you'd like:

Usually you want to test components positioned high up in the three to get closer to the real user interaction possible (integrating child components) and it can make the DOM tree very long. Due to that, sometimes is hard to find/debug specific parts of the DOM only using the `debug` method provided by the `render` API.

The `within` API is very helpful to find elements inside a specific container of the DOM but it does not return a `debug` method to help developers to see/debug that specific part of the DOM you are interested on.

### Suggested implementation:

It wold be very nice to extend the `within` API, providing `debug` method so that it helps developers to debug only part of the DOM tree. I thought of adding this method in the `getQueriesForElement` (https://github.com/kentcdodds/dom-testing-library/blob/master/src/get-queries-for-element.js). That was the only solution that came into my mind so far.

### Describe alternatives you've considered:

Currently I am doing that like this:

```javascript
import { react, prettyDOM } from 'react-testing-library'
const renderedElement = render(<MyComponent />)


console.log(prettyDOM(renderedElement.getByTestId('some-test-id')))
```

I could also extend the `within` API in my project to return that kind of function. Both options would work just fine but I thought it would be a nice idea to extend the `within` API to support such feature.

### Teachability, Documentation, Adoption, Migration Strategy:

The sintax I had in mind was something like this:

```javascript
import { within } from 'react-testing-library'
const renderedElement = render(<MyComponent />)



// Here is the suggestion
const { debug, getByText } = within(renderedElement.getByTestId('some-test-id'))
expect(getByText('my-assertion')).toBeInTheDocument() // some assertions
debug() // This would print in the terminal only the container we targeted rather than the hole tree
```

kentcdodds:
I'm good with this 👍 PRs welcome.

kentcdodds:
See the PR comments.

PR #250, opened 2019-04-24
feat: add debug to default queries (#249)

renerbaffa:
**What**:

Addresses #249 adding a debug method to the default list of queries

**Why**:

Helps other developers to debug part of the DOM tree (by having the debug function on `within` API)

**How**:

I thought of exporting this debug method in the queries file but this is not a query and wouldn't make sense to have it there. Then I decided to extend the default queries in `getQueriesForElement`. Thus we don't touch the queries and the developers can still use their own custom queries.

**Checklist**:

- [x] Documentation added to the [docs site](https://github.com/alexkrolick/testing-library-docs)
  - I have created the issue in the project and will work on that soon (https://github.com/alexkrolick/testing-library-docs/issues/82)
- [x] Tests
- [ ] Ready to be merged 

I still have 3 questions:

1 - Do we need to update Typescript definitions? 🤔

2 - Is it fine to have a function that does `console.log` in the `dom-testing-library` given that other libraries use it? (I know it would be fine for `react-testing-library` but I am not 100% sure about the others)

3 - Can we get rid of this line?
https://github.com/kentcdodds/react-testing-library/blob/master/src/index.js#L62
Since `getQueriesForElement` will now return the `debug` function down in the line 81 I think it is safe, no?

### DIFF
--- a/src/__tests__/get-queries-for-element.js
+++ b/src/__tests__/get-queries-for-element.js
@@ -6,6 +6,7 @@ test('uses default queries + debug method that prints out the given DOM element'
   const {debug, ...defaultBoundQueries} = getQueriesForElement(container)
   expect(Object.keys(defaultBoundQueries)).toEqual(Object.keys(queries))
   expect(debug).toBeDefined()
+  debug() // need to invoke the function for coverage
 })
 
 test('accepts custom queries', () => {
--- a/src/__tests__/get-queries-for-element.js
+++ b/src/__tests__/get-queries-for-element.js
@@ -1,10 +1,11 @@
 import {getQueriesForElement} from '../get-queries-for-element'
 import {queries} from '..'
 
-test('uses default queries', () => {
+test('uses default queries + debug method that prints out the given DOM element', () => {
   const container = document.createElement('div')
-  const boundQueries = getQueriesForElement(container)
-  expect(Object.keys(boundQueries)).toEqual(Object.keys(queries))
+  const {debug, ...defaultBoundQueries} = getQueriesForElement(container)
+  expect(Object.keys(defaultBoundQueries)).toEqual(Object.keys(queries))
+  expect(debug).toBeDefined()
 })
 
 test('accepts custom queries', () => {
--- a/src/get-queries-for-element.js
+++ b/src/get-queries-for-element.js
@@ -1,4 +1,10 @@
 import * as defaultQueries from './queries'
+import {prettyDOM} from './pretty-dom'
+
+function debug(...args) {
+  // eslint-disable-next-line no-console
+  console.log(prettyDOM(...args))
+}
 
 /**
  * @typedef {{[key: string]: Function}} FuncMap
@@ -9,7 +15,7 @@ import * as defaultQueries from './queries'
  * @param {FuncMap} queries object of functions
  * @returns {FuncMap} returns object of functions bound to container
  */
-function getQueriesForElement(element, queries = defaultQueries) {
+function getQueriesForElement(element, queries = {debug, ...defaultQueries}) {
   return Object.keys(queries).reduce((helpers, key) => {
     const fn = queries[key]
     helpers[key] = fn.bind(null, element)


kentcdodds:
Hey @renerbaffa, thanks for this, but after thinking about it a bit more I'm starting to think this is something we should not do. The `getQueriesForElement` function should really only return queries and not utilities like this.

I'd like to hear what others have to say about it, but I'm leaning toward a no.

I'm really sorry to backpedal on you especially after inviting you to make a PR.

bcarroll22:
Was tracking this one as well, and was curious: can't this technically already happen? `debug()` accepts an element as an optional argument, so you can already "scope" it to whatever element you want?

bcarroll22:
I know what I said is only relevant to `react-testing-library`, but any extension of `dom-testing-library` could do the same. `debug` provided by this library is sufficiently powerful to handle what it needs to for this to happen, and any wrapper library could implement this type of behavior, yeah?

kentcdodds:
Yeah, that's kinda what I'm thinking as well. And I really want `getQueriesForElement` to only expose queries and nothing else I think.

renerbaffa:
@bcarroll22 the equivalent function would be `prettyDOM` and that would do the work. As I mentioned in the issue, the alternative would be:

```javascript
console.log(prettyDOM(renderedElement.getByTestId('some-test-id')))
```

@kentcdodds I see your point and I kind of agree with you. After digging the code a bit I felt a bit strange to add this utility function there.

When I suggested this one I had exactly the `react-testing-library` as the use case and I still think that extending the `within` there would be very handy since you are intended to use `within` anyways. You don't need to repeat yourself finding the element just for debugging purpose.

Or am I overengineering things here? 🤔

bcarroll22:
@renerbaffa I think I’m confused. This works, no?

```
const el = getElementByText(/hello world/I);
const queries = within(el);

debug(el);
```

I think this is the same functionality that this PR is for?

bcarroll22:
Looking at the testing-library docs, they actually don't mention that you can do this with debug in `react-testing-library`.

Would you be able to make a PR to the docs that mentions this capability exists @renerbaffa?

renerbaffa:
@bcarroll22 I don't think so. This is an example of what `within` returns:

```javascript
{
  getByPlaceholderText: [Function: bound getByPlaceholderText],
  getAllByPlaceholderText: [Function: bound getAllByPlaceholderText],
  queryByText: [Function: bound queryByText],
  queryAllByText: [Function: bound queryAllByText],
  getByText: [Function: bound getByText],
  getAllByText: [Function: bound getAllByText],
  queryByLabelText: [Function: bound queryByLabelText],
  queryAllByLabelText: [Function: bound queryAllByLabelText],
  getByLabelText: [Function: bound getByLabelText],
  getAllByLabelText: [Function: bound getAllByLabelText],
  queryByAltText: [Function: bound queryByAltText],
  queryAllByAltText: [Function: bound queryAllByAltText],
  getByAltText: [Function: bound getByAltText],
  getAllByAltText: [Function: bound getAllByAltText],
  queryBySelectText: [Function: bound queryBySelectText],
  queryAllBySelectText: [Function: bound queryAllBySelectText],
  getBySelectText: [Function: bound getBySelectText],
  getAllBySelectText: [Function: bound getAllBySelectText],
  getByTestId: [Function: bound getByTestId],
  getAllByTestId: [Function: bound getAllByTestId],
  queryByTitle: [Function: bound queryByTitle],
  queryAllByTitle: [Function: bound queryAllByTitle],
  getByTitle: [Function: bound getByTitle],
  getAllByTitle: [Function: bound getAllByTitle],
  getByValue: [Function: bound getByValue],
  getAllByValue: [Function: bound getAllByValue],
  queryByDisplayValue: [Function: bound queryByDisplayValue],
  queryAllByDisplayValue: [Function: bound queryAllByDisplayValue],
  getByDisplayValue: [Function: bound getByDisplayValue],
  getAllByDisplayValue: [Function: bound getAllByDisplayValue],
  getAllByRole: [Function: bound getAllByRole],
  getByRole: [Function: bound getByRole],
  findByLabelText: [Function: bound ],
  findAllByLabelText: [Function: bound ],
  findByPlaceholderText: [Function: bound ],
  findAllByPlaceholderText: [Function: bound ],
  findByText: [Function: bound ],
  findAllByText: [Function: bound ],
  findByAltText: [Function: bound ],
  findAllByAltText: [Function: bound ],
  findByTitle: [Function: bound ],
  findAllByTitle: [Function: bound ],
  findByDisplayValue: [Function: bound ],
  findAllByDisplayValue: [Function: bound ],
  findByRole: [Function: bound ],
  findAllByRole: [Function: bound ],
  findByTestId: [Function: bound ],
  findAllByTestId: [Function: bound ],
  queryByPlaceholderText: [Function: bound bound queryByAttribute],
  queryAllByPlaceholderText: [Function: bound bound queryAllByAttribute],
  queryByTestId: [Function: bound queryByTestId],
  queryAllByTestId: [Function: bound queryAllByTestId],
  queryByValue: [Function: bound bound queryByAttribute],
  queryAllByValue: [Function: bound bound queryAllByAttribute],
  queryByRole: [Function: bound bound queryByAttribute],
  queryAllByRole: [Function: bound bound queryAllByAttribute] }
```

It does not give you access to the element itself. So the only way to debug specific DOM parts is by doing:

```javascript
console.log(prettyDOM(renderedElement.getByTestId('some-test-id')))
```

That is why I suggested to have this `debug` there, to make it simple to debug. The solution there would be to extend the `within` before exporting it. Does that make sense?

If the issue does not fit `react-testing-library` either, I'll open a PR to the doc at least to document that kind of debug :)

kentcdodds:
Sorry @renerbaffa, but I think you misunderstand....

When you call `within`, you're calling it with a specific element. What @bcarroll22 is saying is the correct way to do this. If you could make a codesandbox of what you're trying to do, we can update it to show you what we mean.

bcarroll22:
What within does is irrelevant in my snippet. Pretend it's not there. I could query any element in the DOM, and pass it to debug, and it would print to the console.

```
const Component = () => (
  <div>
    <span>hello world</span>
  </div>
)

const { getByText } = render(<Component />);
debug(getByText(/hello world/i);
```

this will print:

```
<span>
  hello world
</span>
```

renerbaffa:
What I meant in the other comment was that debug in that example would be undefined, unless you deconstruct `debug` from the `render`, no?

```
const { getByText, debug } = render(<Component />);
```

If that was the case I totally got your point. I'll create a codesandbox example soon to show you guys what I was intended from the beginning. Then you can judge if it is worthy implementing or not.

Anyways, I can use this issue (alexkrolick/testing-library-docs#82) to create a PR for at least documenting this way to debug.

bcarroll22:
That's a good catch, I did forget to grab `debug` off of the render result 👍

renerbaffa:
Here is an example on code sandbox:

https://codesandbox.io/embed/1oovlrk4r4

kentcdodds:
Yeah, so here's how I would do that:

```javascript
const {debug, getByTestId} = render(<Component />)
const specificSelection = getByTestId('specific-section')

const {getByText} = within(specificSelection)
debug(specificSelection)

expect(getByText(/content/i)).toBeInTheDocument()
```

I understand that maybe this is not as nice as getting `debug` from `within`, but I really think that this is the best way. If you really want, you could create your own within utility as well:

```javascript
import {within as dtlWithin, prettyDOM} from 'dom-testing-library'

function within(element) {
  const utils = dtlWithin(element)
  utils.debug = (el = element) => console.log(prettyDOM(el))
  return utils
}

export {within}
```

I'd suggest that you could do this as part of [the recommended custom utility](https://testing-library.com/docs/react-testing-library/setup#custom-render) and then it would be pretty seamless. But I don't see this use case as general enough to be applicable to everyone using *-testing-library.

Thanks anyway!

Issue #257, opened 2019-05-11
Could fireEvent return the event created somehow?


dbismut:
### Access the event created by fireEvent

I would like to get access the event created by `fireEvent`. I rely on `event.timeStamp` to do calculations in my component and testing without access to the created event is complex.

### Suggested implementation:

```js
const { canceled, event } = fireEvent[key](node, { ...eventProps })
```

`canceled` would be the result of `element.dispatchEvent` (ie the current return value of `fireEvent`) and `event` would be the event created. 

### Describe alternatives you've considered:

- Base my calculations on `Date.now()` in my component
- Use `timeStamp = Date.now() + 1` right before `fireEvent` (for some reason they equal the `timeStamp` created by `fireEvent`
- [current solution]: create the event manually and fire it with `fireEvent(element, event)`

kentcdodds:
What if instead of that we exposed an easy way to create the event you pass?

```javascript
import {createEvent} from 'dom-testing-library'

const myEvent = createEvent.click()
fireEvent(node, myEvent)
```

dbismut:
Indeed! Way more elegant!

kentcdodds:
Would you be willing to [make a pull request](http://makeapullrequest.com) to do that?

dbismut:
Sure, if I have time, tomorrow!

kentcdodds:
This is done :)

PR #258, opened 2019-05-12
Add createEvent that returns the created Event

dbismut:
**What**:

Exposes a utility to create events that can be thrown by `fireEvent`. This is useful in order to have access to the source event attributes see #257

**Why**:

Some components might rely on event attributes created when the event object is created (i.e. `new Event({...eventAttributes})`).

**How**:

As suggested by @kentcdodds in #257:

```js
import {createEvent} from 'dom-testing-library'

const myEvent = createEvent.click(node, eventAttributes)
fireEvent(node, myEvent)
```

However since the event constructor depends on `window` which itself depends on the node, `createEvent` also requires the node to be passed as an argument (not very elegant I admit).

**Checklist**:

- [ ] Tests: since `fireEvent` now uses `createEvent` I believe successful tests on `fireEvent` cover `createEvent`
- [x] Ready to be merged 

Pre-commit hook threw an error on Typescript definitions but I'm really not sure what's wrong, so I committed with the `--no-verify` flag.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,8 +30,7 @@
     "test:update": "npm test -- --updateSnapshot --coverage",
     "validate": "kcd-scripts validate",
     "setup": "npm install && npm run validate -s",
-    "dtslint": "dtslint typings",
-    "precommit": "kcd-scripts pre-commit"
+    "dtslint": "dtslint typings"
   },
   "husky": {
     "hooks": {
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
     "test:update": "npm test -- --updateSnapshot --coverage",
     "validate": "kcd-scripts validate",
     "setup": "npm install && npm run validate -s",
-    "dtslint": "dtslint typings"
+    "dtslint": "dtslint typings",
+    "precommit": "kcd-scripts pre-commit"
   },
   "husky": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -30,7 +30,8 @@
     "test:update": "npm test -- --updateSnapshot --coverage",
     "validate": "kcd-scripts validate",
     "setup": "npm install && npm run validate -s",
-    "dtslint": "dtslint typings"
+    "dtslint": "dtslint typings",
+    "precommit": "kcd-scripts pre-commit"
```
kentcdodds: Could you revert these changes?

dbismut: Humpf sorry. I was trying to see what was wrong with my commit on typescript definitions...

kentcdodds:
Could you make a pull request to the docs site to add documentation?

kentcdodds:
@allcontributors[bot] please add @dbismut for code

dbismut:
Yes I will PR the docs (although my English might be shaky)! Is it necessary to add createEvent as an export to react-testing-library?

kentcdodds:
react-testing-library re-experts everything from dom-testing-library so we're good there 👍

Issue #282, opened 2019-06-08
Add `getRoles` and `logRoles` utilities


kentcdodds:
### Describe the feature you'd like:

With the addition of `aria-query` which made huge improvements to the `ByRole` queries in #280 (thank you again @szabototo89!) I would like to suggest something which I think would help people discover improved ways to semantically query their output:

```javascript
import {getRoles, logRoles} from '@testing-library/dom'

const container = document.getElementById('some-element')
const roles = getRoles(container)
// roles is an object where the key is the role name and the value is an array of elements which match that role.

logRoles(container)
// logs each element with debugDOM. Something like:
//
// form:
//
//   <form>
//     <input type="radio" />
//     <input type="radio" />
//   </form>
//
// -------------------------
//
// radio:
//
//   <input type="radio" />
//
//   <input type="radio" />
//
// ------------------------
// or something like that...
```

### Suggested implementation:

I'm not sure. Anyone have ideas/want to try to implement this?

### Describe alternatives you've considered:

Alternatively people have to learn all the possible roles which is unlikely to happen. Having this will encourage people to use semantic roles to query around the DOM for what they're looking for which would be really helpful.

Also, it occurs to me that it may be even more helpful to log _every_ element on the page with the queries that could be used to find it. That would be AMAZING, but seems like a pretty hard problem...

### Teachability, Documentation, Adoption, Migration Strategy:

We'd want to feature this in the docs so people get used to using this regularly in their workflow.

I'm in favor of this (obviously) and would love to see a PR for it. I'm going to be away from the keyboard on a family trip for the next week so I will not be responsive for a while, but I'd love to read what people think/come up with when I get back :)

szabototo89:
We could extend `debugDOM` function with an optional second parameter and adding a configuration parameter to enrich printed DOM with explicit roles. Just like in this example:
```html
<form id="container">
    <input type="radio" />
    <input type="radio" />
</form>
```
```js
const container = document.getElementById('container');
debugDOM(container) // simple use-case
// prints this:
// <form id="container">
//    <input type="radio" />
//    <input type="radio" />
// </form>

debugDOM(container, { logRoles: true }) // new configuration parameter to print container with ARIA roles
// prints this:
// <form id="container" role="form">
//    <input type="radio" role="radio" />
//    <input type="radio" role="radio" />
// </form>
```

michaellasky:
Created a draft PR for this.

#285

alexkrolick:
[...]

PR #285, opened 2019-06-14
Getroles logroles

michaellasky:
This is a draft implementation for the functionality requested in #282  

The intention is to give a jumping off point for discussion.  
 

I've added a file named role-helpers.js which implements the two requested functions, and also moved the function buildElementRoleList from src/queries/role.js into role-helpers.js.  Since role-helpers needed similar functionality, and it didn't seem to make sense to export it from query/role with a bunch of query functions.  

I also added a dependency on lodash.merge for a deep merge in getRoles.  Maybe this could be removed because the merge happening there isn't all that complicated.  But using the premade merge() provided by lodash allowed me to keep the initial PR code cleaner for discussion.  

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

I'm still pretty new to this particular code base, so I don't think  it's ready to be merged without some significant review and iteration from more experienced members.


michaellasky:
I have code for a test that doesn't rely on snapshotting if that's easier for the CI.  I'm using a snapshot because the test shouldn't necessarily fail (beyond needing to update the snapshot) if the formatting from debugDOM changes.  If I hardcode the output from debugDOM into a test case, then the code would need to be updated if debugDOM formatting changes.

src/__tests__/role-helpers.js
```diff
@@ -0,0 +1,120 @@
...
+    listitem: [li1, li2],
+    table: [table],
+    row: [tr],
+    cell: [td1, td2, td3],
+    form: [formEl],
+    textbox: [input, input2],
+  })
+})
+
+test('logRoles is a function', () => {
```
kentcdodds: I don't think we need this test as the other test calls this function.

src/__tests__/role-helpers.js
```diff
@@ -0,0 +1,120 @@
...
+    tr,
+    td1,
+    td2,
+    td3,
+    formEl,
+    input,
+    input2,
+  }
+}
+test('getRoles is a function', () => {
```
kentcdodds: I don't think we need this test.

src/__tests__/role-helpers.js
```diff
@@ -0,0 +1,120 @@
+import {getRoles, logRoles} from '../role-helpers'
+
+function setup() {
```
kentcdodds: Maybe we could use this instead:

```javascript
const section = document.createElement('section')
section.innerHTML = `
  the html for everything else...
`
```

The way it is currently is a bit hard to visualize.

src/__tests__/role-helpers.js
michaellasky: I don't disagree, but just for discussion the reason I include a test like that is that it distinguishes between problems of implementation and problems with configuration/plumbing.  If that test passes and the others fail then you know it's a problem with the implementation.  If that one fails along with the others then it lets you know at a glance there's some other problem (it's not being exported or something).  

But in this, and many cases it doesn't matter, and is mostly a way to get past the 'blank page' syndrome, a test to get started, haha.  

I'll pull them out.

src/__tests__/role-helpers.js
michaellasky: Will do.  The only sticking point with doing it that way is that I then have to reach into that dom and pull all those elements out to add them to my expected obj.  But I think the readability of the dom itself is a bigger win, so i agree something like your suggestion should be done.

michaellasky:
> Also, a screenshot would be helpful!


[Quick screenshot from npm test with a console log](https://drive.google.com/file/d/1xpzwVwUqQzQFqqvGKg1U2CB7FXr-OgXL/view?usp=sharing)  I'll spin it up in the browser in a bit when I make the other changes and take a look at / screencap the output there.  

I can of course make any changes to the formatting.  I didn't match the suggested implementation exactly, but can if that's wanted.

kentcdodds:
Looks pretty good to me!

1. It would probably be helpful to indent the DOM output so the sections are more visible
2. It would probably be helpful to have a separator (even just a single line) between each DOM output so people don't think they represent the same node or something.
3. That `<article>` node has a bunch of stuff in it and if it were really big, that output could be VERY extensive. I wonder if we should just clip it to `<article />` instead of `<article>all the contents</article>` Thoughts?

michaellasky:
>     3. That `<article>` node has a bunch of stuff in it and if it were really big, that output could be VERY extensive. I wonder if we should just clip it to `<article />` instead of `<article>all the contents</article>` Thoughts?

I actually had a similar thought.  I didn't pursue it because that would be more of a change to debugDOM I think?  We'd need to add a parameter to tell it not to print all the children (assuming it can't already do that, i haven't looked at it extensively).  The current implementation is just debugDom(node), which in the case of a large node like the article node, prints out all the children.  

In any event, i don't think it'd be very difficult to handle.

kentcdodds:
My thinking is that before we hand it to `debugDOM` we could create a copy of the DOM node (via something similar to [asFragment](https://github.com/testing-library/react-testing-library/blob/3076b119172b8b7188603509bfe576b7d2914de3/src/index.js#L69-L80)) and then remove all the children.

michaellasky:
Oh, I hadn't considered that.  That's not a bad idea, I'll give it a shot and update this (hopefully tonight, time permitting) with that and the other suggested changes.

michaellasky:
So I've made most of the discussed changes.

[This is what the output of logRoles looks like now](https://drive.google.com/file/d/1AXMSkON1U6Xemj8eL2I7RGhGpZmUE6yB/view?usp=sharing)

I also fixed a bug with getRoles, and refactored it to clean it up and simplify things.  

logRoles uses a snapshot test again.  The regex only worked if the html under test is very simple, and it a tad more 'real world'.  So rather than putting an ugly regex in the test i went back to a snapshot.  I'm not sure what is needed to tell travis to save the snapshot so it'll pass the build though.  

Lodash.merge dependency is gone.

michaellasky:
I'm pretty sure the current build failure is unrelated to these code changes.  

wait-for-element-to-be-removed.js is timing out at some point in one of the builds, which I had noticed happening intermittently when i had issues earlier (when my build was also failing for coverage reasons).

The assumption is that it'd (probably) pass if i closed and re-opened the PR to trigger a rebuild, but I'm leaving it so someone who knows more about what's going on can verify.

src/__tests__/__snapshots__/role-helpers.js.snap
```diff
@@ -37,6 +37,28 @@ article ###################################
 [36m/>[39m
 
 
+command ###################################
+
+[36m<menuitem[39m
```
kentcdodds: Maybe we could use https://github.com/joaogranado/jest-serializer-ansi to make these snapshots better.

src/__tests__/__snapshots__/role-helpers.js.snap
```diff
@@ -0,0 +1,166 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`logRoles logs expected roles for various dom nodes 1`] = `
+"region ####################################
```
kentcdodds: Rather than `##################` could we use `----------------------`?

michaellasky: Switched [deb5bf5c](https://github.com/testing-library/dom-testing-library/pull/285/commits/de5bf5c2a184ed999e80e33b7ec3f05af767b22c)

src/__tests__/__snapshots__/role-helpers.js.snap
michaellasky: Added jest-serilizer-ansi  

Also updated a couple other tests ([e452465](https://github.com/testing-library/dom-testing-library/pull/285/commits/e4524653e4b0212063d4d2e715cf99bdb479b7c2)) that had inline snapshots containing the same control characters.

src/__tests__/__snapshots__/role-helpers.js.snap
```diff
@@ -0,0 +1,166 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`logRoles logs expected roles for various dom nodes 1`] = `
```
kentcdodds: I think this looks really good. Would love to hear what others have to say though.

src/index.js
```diff
@@ -12,6 +12,7 @@ export * from './get-node-text'
 export * from './events'
 export * from './get-queries-for-element'
 export * from './query-helpers'
+export * from './role-helpers'
```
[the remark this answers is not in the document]
afontcu: I love this PR! I learned a ton by reading your code and comments. 

Now that you've discussed public API, I'm still not sold on the idea of the current naming. I feel that, as exposed in https://github.com/testing-library/dom-testing-library/issues/282#issuecomment-502298508, `getRoles` could easily be confused with one of the `getBy*` queries from DTL. What do you folks think about that?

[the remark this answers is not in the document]
kentcdodds: I don't think all that many people do this:

```javascript
import {getByRole} from '@testing-library/dom'
```

I'm pretty sure most people do this:

```javascript
import {render} from '@testing-library/some-implementation'

const {getByRole} = render(myThing)
```

Or this:

```javascript
import {getQueriesForElement} from '@testing-library/dom'

const {getByRole} = getQueriesForElement(someContainer)
```

So I'm not very concerned about this.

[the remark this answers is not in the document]
michaellasky: Good catch on the exports, I had completely forgotten to revisit those.  

I've removed the elementRoleList export, that is certainly not needed anymore, now that getImplicitAriaRoles is moved from src/queries/role to src/role-helpers.

We do need to export getImplicitAriaRoles  from role-helpers because src/queries/role uses it.  That said, I don't think it really needs to be part of the public API.  

Should we just export getRoles and logRoles in index?  Will commit that change after hearing back.

src/__tests__/role-helpers.js
```diff
@@ -0,0 +1,182 @@
...
+  const td2 = getByTestId('a-cell-2')
+  const td3 = getByTestId('a-cell-3')
+  const form = getByTestId('a-form')
+  const radio = getByTestId('a-radio-1')
+  const radio2 = getByTestId('a-radio-2')
+  const input = getByTestId('a-input-1')
+  const input2 = getByTestId('a-input-2')
+  const textarea = getByTestId('a-textarea')
+
+  return {
```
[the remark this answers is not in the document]
michaellasky: Great idea, done.  [2ecda86](https://github.com/testing-library/dom-testing-library/pull/285/commits/2ecda860af4cfa0bb5e22332188b1f0e14cc7dbe)

src/index.js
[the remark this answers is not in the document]
kentcdodds: Yeah, I think that's right 👍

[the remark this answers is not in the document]
michaellasky: [Done b9a16e1](https://github.com/testing-library/dom-testing-library/pull/285/commits/b9a16e1f27d4f57e53d16b992563f4b1d4097303)

I'll get started on documentation tonight or tomorrow as well, now that things are starting to come together.

michaellasky:
I've attempted to add the typescript definitions but I seem to be doing something wrong.  I'm not particularly fluent with typescript. 

I'd imagine the definitions, in a file named typings/role-helpers.d.ts, should look something like:
  
    export function logRoles(container: HTMLElement): string
    export function getRoles(
        container: HTMLElement,
     ): {[index: string]: HTMLElement[]}

But unfortunately that fails linting (kcd-scripts) on build with:

2:1  error  Parsing error: Unexpected token, expected "{"

After some experimentation, I can only get it to pass if I add an implementation for some reason.  For debugging purposes, I changed it to:

    export function logRoles(container: HTMLElement): string
    export function getRoles(container: HTMLElement): string

Which fails in the same way.  But:

    export function logRoles(container: HTMLElement): string { return container? '1': '2' }
    export function getRoles(container: HTMLElement): string { return container? '1': '2' }

passes linting fine.  I am confuse.  But it's probably just some assumption I'm making because of lack of TS experience.

kentcdodds:
Yeah, honestly I'm not sure what's up with that 😞 we should probably just disable linting on that file. This always happens to me when I change that file so I just commit that file with --no-verify 😬

I think things look pretty good. I'd love another review though!

kentcdodds:
Oh, and it looks like there's a merge conflict 😬

michaellasky:
Added the typescript definitions.  Whoever else reviews these changes should pay particular attention to them, since I don't have complete confidence they're correct.  

[93f9530](https://github.com/testing-library/dom-testing-library/pull/285/commits/93f9530baf31ecdb848044f6272d143744107cdd

src/role-helpers.js
```diff
@@ -0,0 +1,90 @@
...
+    )
+  }, {})
+}
+
+function logRoles(container) {
+  const roles = getRoles(container)
+
+  return Object.entries(roles)
+    .map(([role, elements]) => {
+      const numDelimeters = 42 - role.length // 42 is arbitrary
```
alexkrolick: As far as I know there aren't any aria-roles greater than 42 chars, but I would really hate for a formatting error to break someone's test. Can we be conservative here and do a Math.min(0, numDelimiters) or move the delimiter characters to a fixed length line?

michaellasky: That's a good point.  I think it's very unlikely there will ever be a role greater than 42 characters long... there's also no real downside to protecting against that.  Maybe I was being too clever with embedding the name in the delimiters.  It read easier (to my eyes) with just a single heading/spacer element, but it probably makes more sense to just implement it like Kent originally had it and avoid any confusion, extra code, or bugs (however unlikely). 

I'll make that change this evening unless I hear an objection.

michaellasky: Done [593c33f](https://github.com/testing-library/dom-testing-library/pull/285/commits/593c33f987bba473a3c0c89f446e8a1af604d7fd)

Check the snapshot for output format.  

I would have like to indent the elements themselves to further distinguish them from the headings, but unfortnately I couldn't find a decent way to do that with the way debugDOM->prettyDOM->prettyFormat works.  They inject their own newlines into the output, so if I try to indent it I can only indent the first line.  I'd need to split the output of debugDOM up by lines and indent them individually... which seemed messy for a minor visual improvement.  

I also purposefully left out the space between the delimeter bar ('-----') and the role name label.  With the space it looks a lot louder visually. 

     navigation:
    <nav
        data-testid="a-nav"
    />

    heading:

    <h1
      data-testid="a-h1"
    />

vs

     navigation:
    <nav
        data-testid="a-nav"
    />

    heading:

    <h1
      data-testid="a-h1"
    />

kentcdodds:
@all-contributors please add @michaellasky for code, tests, and docs

kentcdodds:

michaellasky:
Thanks Kent, you're making open source a lot of fun.  I'll continue to watch and contribute.

PR #288, opened 2019-06-19
fix: support jsdom detached fragments

DylanPiercey:
**What**:
This PR builds on top of https://github.com/testing-library/dom-testing-library/commit/c4b4efffb19233e1cca9799544e7ab016b3bd916 and adds support for using the [`JSDOM.fragment`](https://github.com/jsdom/jsdom#fragment) API for server side tests.

**Why**:

I think using the JSDOM.fragment API makes sense for SSR only tests. In theory if there was client side logic you would instead be running these tests in a browser context.

With this fragment API there is no `window`, which caused the existing jsdom test to break.
My current thinking is that for SSR tests you really shouldn't be testing events and what not, instead use a browser test for that. Using the fragment API outputs a sane error if you try to do so, while still supporting the same API for everything else.

**How**:

I've made a couple small changes that account for the scenario when the DOM is not in a browser context.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -344,7 +344,7 @@ function getWindowFromNode(node) {
   if (node.defaultView) {
     // node is document
     return node.defaultView
-  } else if (node.ownerDocument) {
+  } else if (node.ownerDocument && node.ownerDocument.defaultView) {
     // node is a DOM node
     return node.ownerDocument.defaultView
   } else if (node.window) {
--- a/src/get-node-text.js
+++ b/src/get-node-text.js
@@ -1,4 +1,7 @@
+// Constant node.nodeType for text nodes, see:
+// https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType#Node_type_constants
 const TEXT_NODE = 3
+
 function getNodeText(node) {
   if (node.matches('input[type=submit], input[type=button]')) {
     return node.value
--- a/src/get-node-text.js
+++ b/src/get-node-text.js
@@ -1,15 +1,11 @@
+const TEXT_NODE = 3
 function getNodeText(node) {
-  const window = node.ownerDocument.defaultView
-
   if (node.matches('input[type=submit], input[type=button]')) {
     return node.value
   }
 
   return Array.from(node.childNodes)
-    .filter(
-      child =>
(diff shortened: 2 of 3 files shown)


src/get-node-text.js
```diff
@@ -1,15 +1,11 @@
...
   if (node.matches('input[type=submit], input[type=button]')) {
     return node.value
   }
 
   return Array.from(node.childNodes)
-    .filter(
-      child =>
-        child.nodeType === window.Node.TEXT_NODE && Boolean(child.textContent),
-    )
+    .filter(child => child.nodeType === TEXT_NODE && Boolean(child.textContent))
```
DylanPiercey: Here I opted to just inline the TEXT_NODE constant. Alternatively it looks like this is also available as `document.TEXT_NODE` although I'm not entirely sure what the browser support for that is. Using the constant seemed sane as it is only in one place so far.

src/get-node-text.js
```diff
@@ -1,15 +1,11 @@
+const TEXT_NODE = 3
```
kentcdodds: Could we add a link to MDN docs that explain that this is `window.Node.TEXT_NODE` and the value is `3` (per the spec). Otherwise this is surprising.

src/events.js
```diff
@@ -344,7 +344,7 @@ function getWindowFromNode(node) {
   if (node.defaultView) {
     // node is document
     return node.defaultView
-  } else if (node.ownerDocument) {
+  } else if (node.ownerDocument && node.ownerDocument.defaultView) {
```
DylanPiercey: With JSDOM fragments there is still an `ownerDocument`, but no `defaultView`.

src/get-node-text.js
DylanPiercey: Good idea, updated the pr 😄

kentcdodds:
@all-contributors please add @DylanPiercey for code and tests

DylanPiercey:
@kentcdodds not sure if you use gitter, but I reached out to you on there with some quick questions regarding adding a testing-library/marko module. Is there an appropriate place other than twitter to have a bit of a back and forth? 😛

kentcdodds:
Go ahead and open an issue on dom-testing-library :)

Issue #201, opened 2019-02-04
getByText()/queryByText() (or some new query) should search across nested elements


helixbass:
### Describe the feature you'd like:

In my app, I'm rendering some markup like:
```
<div>
  <span>a. </span>
  <span class="something">Here is the answer text.</span>
</div>
```

It feels like I should be able to test that the right "answer prefix" (here, "a.") is rendered with the corresponding answer text just by doing eg:
```
getByText(/a\. Here is the answer text\./)
```
(which would select the enclosing `<div>`)

So I guess recursively concatenating text node children from nested elements?

It didn't look like there's already a way to do this

### Suggested implementation:

I haven't looked at the implementation, just checked that in `__tests__/element-queries.js` I saw a test for sibling text node concatenation but not for nested child element text nodes

Would possibly be up for implementing this but am not super familiar with low-level DOM APIs

### Describe alternatives you've considered:

I guess adding a `data-testid` to the enclosing `<div>` and then examining its children individually? But in the spirit of "what matters is how the user experiences it", that seems fragile compared to just searching text content across whatever arbitrary nested markup you may be rendering

### Teachability, Documentation, Adoption, Migration Strategy:

I think adding an example with simple nested content eg:
```
<div>Some <span>nested</span> text</div>
```
with
```
getByText(/Some nested text/)
```
would be pretty clear as far as the gist of the supported feature

kentcdodds:
I think that I'm in favor of this if we can make it work. I can't promise that I'll be able to work on it any time soon, but I would be happy to review and merge a PR that implements this kind of text searching for all queries. One potential issue I see is:

```javascript
// Given: <div>Some <span>nested</span> text</div>
getByText(/.*nested.*/) // what should this return?
```

The outer div and the span both match. I think a naive implementation of this would result in returning the `<div>`, but perhaps the `<span>` is what you'd want. This is tricky. I've opened https://github.com/kentcdodds/dom-testing-library/issues/202 with an idea to help deal with this problem.

alexkrolick:
I think this would fix these queries, if someone wants to pick it up: https://github.com/kentcdodds/dom-testing-library/pull/190

alexkrolick:
Turns out #190 probably isn't the right solution since .innerText isn't actually implemented/implementable in JSDOM (missed that part, sorry). 

[...]

PR #303, opened 2019-07-01
feat: Add ByTextContent query

eps1lon:
This is a mix of RFC, feature request and personal experience when writing tests. Maybe I'm missing something obvious about getByText but I do not find this query useful in its current state.

**What**:

Add `*ByTextContent` query

**Why**:

Simpler, more user centric API for finding elements with a given `textContent`

**How**:

Simply iterate over all elements matching the given selector and comparing with `textContent` (exact).

**Checklist**:

- [x] Tests
- [ ] Ready to be merged

I think [getNodeText](https://github.com/testing-library/dom-testing-library/blob/b1644c2b66a2055f4ca3d020e140474d786dbf0b/src/get-node-text.js#L5) is currently overly complex which can lead to surprising behavior in `*ByText` queries.

Consider the following button: `<button><span>Button</span></button>` or a bit more convoluted `<button><span>But</span><span>ton</span></button>`. Both of these have a `textContent` of `Button` but either `getByText` returns the `span` for the former case or null for the latter case.

I suspect that the current implementation tries to avoid including wrapper divs in e.g. `<div class="wrapper"><p>Text</p></div>`. I understood those query helpers as a11y first helpers which means it would be sufficient to match `textContent` and then let the user define meaningful selectors.

This would encourage thinking about using more semantic elements e.g. if `getByTextContent('Text')` returns multiple spans and divs with the same content one of those should be made a paragraph or given an aria role etc.

Intuitively a user centric approach to testing what happens when the user clicks the button  that says "Click me" would just query for getByTextContent('Click me', 'button'). The second argument would be a selector. No normalization config, no exact option etc. All of this can be expressed via `TextMatch`.

It's IMO ok to fail if surrounding whitespaces changes. Whitespace can change how the user perceives the text (see css `whitespace`) and the test should reflect that. In a perfect world `jsdom` could implement `innerText` and we could use that (it includes layout, whitespace trimming etc).


kentcdodds:
Thanks for this @eps1lon, but it's pretty unlikely that this will happen. I'm SUPER resistant to making two ways to do the same thing and while it appears that this may be better than the current implementation of `getByText`, using straight-up `textContent` means that the `container` would match pretty much any `getByTextContent` query. This is why the current implementation of `getByText` is so complex.

If you do happen to have one of the edge cases as you described, then you can actually make it work with `getByText` pretty easily:

```javascript
// <div><button><span>But</span><span>ton</span></button></div>
getByText(div, (text, el) => /button/i.test(el.textContent)), {selector: 'button'})
```

For this reason, I don't think that we'll do this.

eps1lon:
Completely understandable. Would you mind leaving this open a bit longer in case others would like to comment? 

So far I think a `.textContent` matcher would be more appropriate with the recommendation to chain this of other selectors. At least so far I haven't encountered cases where I would want to only match `Hello, ` in `<p>Hello, <em>World!</em>`. It's currently the only API that still feels like testing implementation details i.e. reading `getByText(div, (text, el) => /button/i.test(el.textContent)), {selector: 'button'})` is something I would immediately try to rewrite but so far queries don't naturally chain of of on another (mostly because they don't need to).

kentcdodds:
Yeah, it's understandably less than optimal. The challenge is that we're trying to work with DOM nodes without actually working with DOM nodes because the user knows nothing about DOM nodes. It's definitely one of the trade-offs we're using to make it at all testable 😬 Not sure how to improve things without introducing more problems.

bcarroll22:
@eps1lon another thing to keep in mind too is that even if this doesn't get merged/accepted, it's pretty easy to add custom queries that you and your team find useful. That way if you prefer this over the current matchers, you can encourage the team to use the custom one instead. 

For example, if you're using React Testing Library, you could use [the queries option](https://testing-library.com/docs/react-testing-library/api#queries) along side the [`buildQueries`](https://testing-library.com/docs/dom-testing-library/api-helpers#buildqueries) helper to add this to a [custom render](https://testing-library.com/docs/react-testing-library/setup#custom-render).

alexkrolick:
I think this is essentially a dupe of https://github.com/testing-library/dom-testing-library/issues/201

The big issue is how to prioritize the matches so that you don't get document.body for a case like:

```html
<body>
  <div>
    <button>Foo</button>
  </div>
</body>
```

PR #304, opened 2019-07-02
feat: Add DescriptionOf

eps1lon:
**What**:

Adds `*DescriptionOf` query method. Example usage: `getDescriptionOf(getByRole('button'))`.

**Why**:

Can be a custom query. However, it is currently explicitly warned against using those because they might not be user-centric. This doesn't apply here.

**How**:

It's a simple wrapper for `querySelector('#' + element.getAttribute('aria-describedby'))`.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

Adding docs and types once this is approved.


src/__tests__/element-queries.js
```diff
@@ -801,4 +801,70 @@ test('get/query textarea element by current value', () => {
...
+    </main>
+  `)
+
+  expect(getDescriptionOf(getByRole('button'))).toHaveTextContent(
+    'Hello, Dave!',
+  )
+  expect(queryDescriptionOf(getByRole('button'))).toHaveTextContent(
+    'Hello, Dave!',
+  )
+  expect(findDescriptionOf(getByRole('button'))).resolves.toHaveTextContent(
```
kentcdodds: Make this test async, and await this assertion. Otherwise it resolves/rejects after the test completes.

src/__tests__/element-queries.js
```diff
@@ -801,4 +801,70 @@ test('get/query textarea element by current value', () => {
...
+  expect(() => getDescriptionOf(getByRole('button'))).toThrow(
+    'Found multiple descriptions. An element should be described by a unique element.',
+  )
+  expect(() => getAllDescriptionOf(getByRole('button'))).toThrow(
+    'An element should be described by a unique element. If you want to expect that only a single element exists prefer getDescriptionOf and expect an error thrown.',
+  )
+  expect(() => queryAllDescriptionOf(getByRole('button'))).toThrow(
+    'An element should be described by a unique element. If you want to expect that only a single element exists prefer queryDescriptionOf and expect an error thrown.',
+  )
+  expect(() => findAllDescriptionOf(getByRole('button'))).toThrow(
```
kentcdodds: Same here

src/__tests__/element-queries.js
eps1lon: This would do the same it's doing now, no? Otherwise wouldn't jest fail because it doesn't resolve?

eps1lon: At least it's how I understood https://jestjs.io/docs/en/asynchronous.html#resolves-rejects

kentcdodds: Try this:

```js
expect(Promise.reject()).resolves.toBe(true)
```

You'll see that the test passes, but then there's a promise rejection in the output.

kentcdodds: Yeah, I prefer to not return. Too easy to forget to do if you change the rest later. Just use async/await.

eps1lon: >  Too easy to forget to do if you change the rest later

Already forgot it without anything else...so yeah :smile:

src/__tests__/element-queries.js
eps1lon: It doesn't return a promise but immediately throws. Should it return a promise that rejects instead?

src/__tests__/element-queries.js
eps1lon: Done

src/__tests__/element-queries.js
kentcdodds: Sorry. Was confused. This makes sense.

kentcdodds:
I'd love feedback from other maintainers. I'm in favor of this.

afontcu:
Quick question. What's the reasoning behind calling the query method `*DescriptionOf`? I feel that making all of them look similar (`verb + (all) + by + query`) is really powerful, since you just have to learn things once. Here we kinda break this.

What do you think about naming it `*ByDescription`?

eps1lon:
> What do you think about naming it *ByDescription?

Given 
```html
<button aria-describedby="button-description">Click me</button>
<p id="button-description">Hello, Dave!</p>
```
These imply the reverse for me i.e. `getByDescription` would be used like `getByDescription('Hello, Dave!') => HTMLButtonElement` whereas `getDescriptionOf(getByRole('button')) => HTMLParagraphElement`. Or in math terms: `*Of` is the inverse of `By*`.

So it purposely breaks the current schema. An example for an inverse for a given query would be `labelOf <-> byLabelText`.

But I'm not a native speaker so maybe something is lost in translation.

alexkrolick:
I'm not totally clear why you would need a selector for this, rather than a jest-dom assertion? As far as I know the DOM doesn't actually associate these elements with hover states or anything.

eps1lon:
> As far as I know the DOM doesn't actually associate these elements with hover states or anything.

`aria-describedby` is part of the DOM spec and does associate elements.

`aria-describedby` establishes relationships and therefore the element in question is interesting beyond text matching. So in essence this is more about semantics than interactions.

alexkrolick:
Yes, it associates the elements for screen readers, but there's no interaction to jump from an element to its description. As you change focus, the screenreader automatically reads first the label, then the description for the focused element. 

Can you write an example test using this method to illustrate the use case?

* Also, both aria-labelledby and aria-describedby take a space-separated list of IDs, so the described-describer relationship is 1:many for a given described element

kentcdodds:
That makes a lot of sense Alex. I'd also prefer that if you want to verify the describe by text of an element you do so in an assertion library (like jest-dom). I can't think of any other reason to get the description element for a particular element.

If you want to get an element by its description, then `getByDescriptionText` makes a lot of sense to me as well 👍

eps1lon:
> Yes, it associates the elements for screen readers

The DOM spec makes no claim about the purpose of this attribute. Screen readers is the obvious choice but it may also be used by search engines or 3rd party extensions since it's purpose is a text alternative not just "for screen readers". So no that statement is inaccurate regarding the spec.

>  I can't think of any other reason to get the description element for a particular element.

Because accessible elements may also be described in visual terms e.g. the description has a different font style, size etc. Accessbility is not just for blind people but also visually impaired for which this library doesn't have proper selectors. I'm basically just repeating myself but the point was to write selectors that test your app in a more semantic i.e. accessible way.

I didn't know that they should only be used for interactions not assertions. I can keep them as custom queries but the strong wording against custom queries is an issue. This query is definitely not not user-centric.

> Also, both aria-labelledby and aria-describedby take a space-separated list of IDs, so the described-describer relationship is 1:many for a given described element

Yeah I saw that. So if anybody wants to use this PR locally please adjust for it.

alexkrolick:
@eps1lon how about this:

\* I think the assertions should in some way account for the 1-many relationship so that they're compliant with the spec

- `getByDescription(text, { selector: '*' })` this simulates the way a user would find an element by reading it's description, also using the established `selector` option for extending the query
- One or more of these in `jest-dom`:
  - `expect(element).toHaveDescriptionText(text)` assertion about the text, possibly concating all the descriptions (I think this is what a screen reader does when multiple descriptions or labels are referenced for an element).
  -`expect(element).toBeDescribedBy(anotherElement)` You could query for the description  element separately and assert about it with `.toHaveStyle()`, etc. Users _not_ using assistive technology don't have any indication in the browser of aria-describedby relationships, so asserting about styles of a description should probably treat querying for the description as a separate action.

eps1lon:
Yeah I'm not interested in those queries because of the limitations of `ByText` queries currently and because classNames are part of the API for us.

Going to close this since there's no outside interest.

Issue #300, opened 2019-06-27
jest.useFakeTimers() breaks waitForDomChange, waitForElement, waitForElementToBeRemoved due to setTimeout 


LaurensBosscher:
**Issue:**

When running any tests with:

- waitForDomChange
- waitForElement
- waitForElementToBeRemoved

with jest.useFakeTimers enabled, these functions only run once and then fail.

This is because they use setTimeout internally to set a timeout. 

However wait does work as expected because wait-for-expect has implemented a workaround, see: https://github.com/TheBrainFamily/wait-for-expect/pull/5/files

Would it be possible to implement this workaround for waitForDomChange, waitForElement, waitForElementToBeRemoved as well? 

I'm willing to make a PR if I could talk the changes through with someone (e.g do we want to implement setTimeout as a util function or do we want to add the workaround to these three functions, how do we make sure that this will not break again in the future?).

kentcdodds:
I'm in favor 👍 thanks.

alexkrolick:
> do we want to implement setTimeout as a util function or do we want to add the workaround to these three functions, how do we make sure that this will not break again in the future

How about implementing it as a util function that performs the workaround from `wait-for-expect` as the initial implementation, so that future changes, if any, are isolated? It looks like they have added a few more tweaks to the workaround since that PR.

https://github.com/TheBrainFamily/wait-for-expect/blob/7bf926a9b347ff297bdc3e4099fb123d85935392/src/index.ts#L10-L14

LaurensBosscher:
> > do we want to implement setTimeout as a util function or do we want to add the workaround to these three functions, how do we make sure that this will not break again in the future
> 
> How about implementing it as a util function that performs the workaround from `wait-for-expect` as the initial implementation, so that future changes, if any, are isolated? It looks like they have added a few more tweaks to the workaround since that PR.
> 
> https://github.com/TheBrainFamily/wait-for-expect/blob/7bf926a9b347ff297bdc3e4099fb123d85935392/src/index.ts#L10-L14
> 
> 
> ```ts
> // Used to avoid using Jest's fake timers and Date.now mocks
> // See https://github.com/TheBrainFamily/wait-for-expect/issues/4 and
> // https://github.com/TheBrainFamily/wait-for-expect/issues/12 for more info
> const { setTimeout, Date: { now } } =
>   typeof window !== "undefined" ? window : global;
> ```

[...]

PR #305, opened 2019-07-04
feature/use-original-setTimeout-when-used-in-a-test

LaurensBosscher:
**What**:

Utilizes the original (non-mocked) setTimeout for waitForDomChange, waitForElement and waitForElementToBeRemoved

**Why**:

waitForDomChange, waitForElement and waitForElementToBeRemoved will now work as expected when the user utilizes this library in a test that uses jest.useFakeTimers(). See: https://github.com/testing-library/dom-testing-library/issues/300

**How**:

I've implemented the workaround used by wait-for-expect: https://github.com/TheBrainFamily/wait-for-expect/pull/5/files. I've added a global (global.useFakeTimers) so that we can still utilize fake timers in our test.

getSetTimeout and getClearTimeout utilize dependency injection because it's not possible to set window to undefined in jest. I would love feedback on the tests that I've added, let me know if you think there's something that could be improved.

**Checklist**:

- [X] Tests
- [X] Ready to be merged


src/helpers.js
```diff
@@ -34,4 +34,54 @@ function getSetImmediate() {
...
+
+/*
+ * Return the original setTimeout function so that all functions work as expected by a user
+ * if he/she utilizes dom-testing-library in a test where jest.fakeTimers() is used.
+ *
+ * global.useFakeTimers is used by us to make sure that we can still utilize fakeTimers in our tests.
+ *
+ * see: https://github.com/testing-library/dom-testing-library/issues/300
+ */
+function getSetTimeout(windowObject) {
```
LaurensBosscher: I've injected the windowObject as a dependency here because (as far as I know) there's no way in Jest to set the window object to be undefined. This seemed to be the best way to make this code testable.

kentcdodds: Instead of this, you could add a test for this case to https://github.com/testing-library/dom-testing-library/tree/master/src/__node_tests__

Other than that, this looks pretty good to me.

LaurensBosscher: @kentcdodds Do you mean instead of dependency injection? 

The node tests are still run within Jest and the main issue I faced was that in Jest window === globals (see: https://github.com/facebook/jest/issues/3692#issuecomment-408132529) and I can't set globals (and thus window) to undefined.

kentcdodds: Is the a reason the code can't be:

```js
return typeof window === 'undefined' ? global.setTimeout : window.setTimeout
```

LaurensBosscher: Updated!

kentcdodds: What I was saying is you no longer need the `windowObj` argument with the code I provided.

LaurensBosscher: Hmm, I get that but how would I be able to test that: "returns original getSetTimeout from global if window is undefined" without passing windowObject?

Perhaps we could chat about what the best approach to test that case would be?

kentcdodds: > returns original getSetTimeout from global if window is undefined

You can add another test to the node environment tests, as I said earlier:

> Instead of this, you could add a test for this case to https://github.com/testing-library/dom-testing-library/tree/master/src/__node_tests

LaurensBosscher: Ahhhhh, it finally clicked this morning. The node tests are run without JSDOM and thus without a window object. I'm new to the project, that wasn't clear to me.

I've updated the PR, removed the dependency injection of window and updated the tests for the node environment. Let me know what you think!

src/__tests__/helpers.js
```diff
@@ -1,9 +1,53 @@
...
+  getClearTimeout,
+} from '../helpers'
 
 test('returns global document if exists', () => {
   expect(getDocument()).toBe(document)
 })
 
+describe('getSetTimeout', () => {
+  it('returns mocked setTimeout if global.useFakeTimers is set and jest.useFakeTimers is used', () => {
+    // global.useFakeTimers is set to true for all tests in tests/setup-env.js
```
kentcdodds: If that's the case, then isn't `jest.useFakeTimers()` on the next line unnecessary and `jest.useRealTimers()` at the end of this test breaking test isolation?

src/helpers.js
```diff
@@ -34,4 +34,50 @@ function getSetImmediate() {
...
+
+let originalClearTimeout = global.clearTimeout
+
+if (typeof window !== 'undefined') {
+  originalClearTimeout = window.clearTimeout
+}
+
+/*
+ * Return the original setTimeout function so that all functions work as expected by a user
+ * if he/she utilizes dom-testing-library in a test where jest.fakeTimers() is used.
```
kentcdodds:
```suggestion
 * if they utilize dom-testing-library in a test where jest.fakeTimers() is used.
```

src/helpers.js
```diff
@@ -34,4 +34,50 @@ function getSetImmediate() {
...
+ * if they utilize dom-testing-library in a test where jest.fakeTimers() is used.
+ *
+ * global.useFakeTimers is used by us to make sure that we can still utilize fakeTimers in our tests.
+ *
+ * see: https://github.com/testing-library/dom-testing-library/issues/300
+ */
+function getSetTimeout() {
+  if (global.useFakeTimers) {
+    // eslint-disable-next-line no-negated-condition
+    return typeof window !== 'undefined' ? window.setTimeout : global.setTimeout
```
kentcdodds: You could avoid that eslint issue with:

```suggestion
    return typeof window === 'undefined' ? global.setTimeout : window.setTimeout
```

kentcdodds: Same as below

src/__tests__/helpers.js
kentcdodds: Actually, now that I think of it, why are we setting `useFakeTimers` to true in setup-env? I think it'd be better if most of our tests run without fake timers wouldn't it?

LaurensBosscher: global.useFakeTimers is a boolean that enables fake timers **if** someone uses jest.useFakeTimers(). So most of our tests will be run without fake timers.

I think we can improve upon the name of the boolean. How about global.returnMockedTimers? Or would you prefer that we leave this undefined by default and only set this to true in our tests that use fake timers (https://github.com/testing-library/dom-testing-library/search?q=jest.useFakeTimers&unscoped_q=jest.useFakeTimers)?

kentcdodds: Oh wait. Are you saying that if people want to take advantage of this feature they have to set `global.useFakeTimers`? I'd really prefer to find another way to detect whether fake timers are happening... Maybe I don't understand what's going on/what the purpose of this is 😬

LaurensBosscher: Nop, nop, exactly the opposite! This behaviour is enabled by default for everyone who uses dom-testing-library in their application tests, but it's disabled for dom-testing-library internal tests.

I do think that I'm explaining and documenting this rather poorly. Let me explain it to you first with a bunch of examples, I hope that clears things up! Perhaps than you give me some pointers on how exactly I can communicate this behaviour better? Your suggestions are very much welcome!

If I as a user use dom-testing-library all the dom-testing library methods that I use in my application tests will use a non-mocked setTimeout. So if I write:

```
jest.useFakeTimers();
findByText('My application is awesome!')
```

dom-testing-library will check for 4.5 seconds if the text 'Awesome Text' is in my application.

Without this PR, the above test will most likely fail because findByText uses a mocked fake timer which reduces the timeout to 1 tick.

If I as a contributor for dom-testing-library write a new test for dom-testing-library and if I write the following test:

````
jest.useFakeTimers();
setTimeOut(() => {
   findByText('Dom testing library is awesome!')
, 9999}
````

This test will use the mocked setTimeout and thus succeed within <10 msecs. This is exactly the same behaviour as before this PR.

The global is used to determine if we're running the dom-testing-library internal tests.

I hope this clears things up :)

The [fix ](https://github.com/TheBrainFamily/wait-for-expect/pull/5/files) for wait-for-expect is a lot smaller because they don't have any internal tests that use jest.useFakeTimers() which greatly reduces the complexity.

Comment (kentcdodds) [reply src/__tests__/helpers.js]: Oh, I see what you're saying. So the whole reason that we have the global thing is for internal testing purposes.

Hmm... I'm not sure I understand why it's necessary though. Why can't DOM Testing Library's tests work the same way that users of DTL test?

Comment (kentcdodds) [reply src/__tests__/helpers.js]: Ok, I think I've worked it out. I'm going to make a contribution to this to improve things I think.

**Comment by kentcdodds** [comment]
Here we go. I think this will be better :) Requires some fanciness in the test, but I'd rather put the fanciness in the tests than in the source code :)

File: src/__tests__/fake-timers.js [hunk 5]
```diff
@@ -0,0 +1,85 @@
+import {render} from './helpers/test-utils'
```
kentcdodds: We've moved all the fake timers tests to this file. The only reason we're using fake timers is for testing timeouts and we don't want this test to take forever.

src/helpers.js
```diff
@@ -1,5 +1,17 @@
 import MutationObserver from '@sheerun/mutationobserver-shim'
```
kentcdodds: Simplifying this quite a bit by not using getters and instead just creating variables. Similar to how wait-for-expect does it.

Issue #311, opened 2019-07-09
logRoles doesn't log


kentcdodds:
This is my fault for not reviewing properly, but the new `logRoles` function simply generates a formatted string and returns it. It does not call `console.log` as the name implies.

## Suggested solution

Rename the current `logRoles` function to `prettyRoles` (analogous to `prettyDOM`) and make a new function called `logRoles` which actually uses console.log to log out the result of `prettyRoles`.

Thoughts?

cc @michaellasky

afontcu:
Wow, this is so funny. I was just trying to get my head around `logRoles` - right now! I couldn't get any output, so I assumed I was doing something wrong 😅 

Question: Should the _new_ `logRoles` live in DTL, though? AFAIK, there's no `logDOM` function here, but in RTL/VTL (where `debug()` is just logging `prettyDOM`).

What are your thoughts?

kentcdodds:
Yes, I had a similar thought. I'd actually be more ok with adding a `logDOM` than removing the `logRoles`.

And I can see value in wrappers adding a `debugRoles` method to their return value as well.

michaellasky:
I actually was going to ask if we wanted to actually call console.log in the function or just return the string, but didn't when nobody had a problem with the string, haha.  

Probably should have been clearer with my intentions there, but thought you might want to keep it general so you could log the output to places other than the console.  

Anyway, I'm fine with submitting a PR to fix this in the next 24 hours, at least as a stopgap until other changes are proposed/made?

kentcdodds:
Yes, that sounds good. For now let's just make `logRoles` actually log the roles string rather than return it.

We can discuss the addition of `logDOM` and `prettyRoles` later. I'd love to hear what others have to say about that.

michaellasky:
Added a quick PR that makes logRoles actually log :).  

Note: In this PR it still returns the string, as before.  I can change this so it only logs, but i figured we want to keep the api that same for the moment?  

It's unlikely anyone is using logRoles extensively, but if it stops returning the string it could break suddenly for someone.

michaellasky:
I also have some code started for the other changed discussed.  

Assuming it ends up something like (abbreviated):
   
       logroles = container => console.log(prettyRoles(container)); 

Any idea how to test logRoles if it's just calling console.log?  Can we mock the console easily?  I don't think i've ever tested for a call to console.log before.

kentcdodds:
Looks like the PR hasn't been opened yet. Did you intend to do so?

[...]

PR #312, opened 2019-07-10
logRoles now logs its output to console.

michaellasky:
API is unchanged, logRoles will still return the same string, but now
also logs to the console.

https://github.com/testing-library/dom-testing-library/issues/311

Initial fix for Issue #311 

See discussion in #311 

logRoles still returns a string, but now also logs its output to the console.

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged


michaellasky:
This build is going to fail because of code coverage.  As mentioned in #311, I'm not sure the best way to go about testing a console log side effect with void return value.

src/__tests__/role-helpers.js
```diff
@@ -132,15 +131,6 @@ test('getRoles returns expected roles for various dom nodes', () => {
   })
 })
 
-test('logRoles logs expected roles for various dom nodes', () => {
-  expect.addSnapshotSerializer(jestSerializerAnsi)
```
kentcdodds: I missed this in the previous review, but you could add this to https://github.com/testing-library/dom-testing-library/blob/master/tests/setup-env.js instead and then it'll be applied to all snapshots which is probably a good thing.

src/__tests__/role-helpers.js
```diff
@@ -132,15 +131,6 @@ test('getRoles returns expected roles for various dom nodes', () => {
   })
 })
 
-test('logRoles logs expected roles for various dom nodes', () => {
```
kentcdodds: To test calls to `console.log` you could do something like this:

```javascript
jest.spyOn(console, 'log').mockImplementationOnce(() => {})

const {section} = setup()
logRoles(section)

expect(console.log).toHaveBeenCalledTimes(1)
expect(console.log.mock.calls[0][0]).toMatchSnapshot()

console.log.mockRestore()
```

kentcdodds: Nice! Looks like you figured it out without my help. Two things:

1. Let's move that `addSnapshotSerializer` stuff to the `setup-env.js` file.
2. Make sure to call `mockRestore()`

src/__tests__/role-helpers.js
```diff
@@ -132,13 +132,16 @@ test('getRoles returns expected roles for various dom nodes', () => {
...
+test('logRoles calls console.log with output from prettyRoles', () => {
   const {section} = setup()
-  const output = logRoles(section)
+  expect.addSnapshotSerializer(jestSerializerAnsi)
 
-  expect(output).toMatchSnapshot()
+  jest.spyOn(console, 'log')
+  logRoles(section)
+  // eslint-disable-next-line no-console
+  expect(console.log.mock.calls).toHaveLength(1)
```
kentcdodds: This should probably be swapped with `expect(console.log).toHaveBeenCalledTimes(1)` to improve any error message that may occur.

src/__tests__/role-helpers.js
michaellasky: Will do!  Test is fixed, was overthinking things.. been a long day. 

Adding the serializer to setup-env incoming

src/__tests__/role-helpers.js
michaellasky: Done [5fefa1](https://github.com/testing-library/dom-testing-library/pull/312/commits/5fefa10d167518fa396538a09ef428b81f2ce45d)

src/__tests__/role-helpers.js
michaellasky: Handled [5fefa1](https://github.com/testing-library/dom-testing-library/pull/312/commits/5fefa10d167518fa396538a09ef428b81f2ce45d)

PR #319, opened 2019-07-12
Fix getByRole('link') not returning anchor elements

### LINKED ISSUE #313: Get/Log Roles helpers doesn't return anchors when querying link roles
- `dom-testing-library` version: `5.5.1`
- `node` version: `10.16.0`
- `npm` (or `yarn`) version: `6.9.0`

### What you did:

Tried to get an anchor element (`<a href="http://whatever.com">`).

Since it wasn't working (even though [the implicit ARIA role exists](https://www.w3.org/TR/html-aria/#document-conformance-requirements-for-use-of-aria-attributes-in-html) in anchor elements as long as they have a `href` attribute) I turned to DTL and added `<a href="http://link.com" data-testid="a-link">link</a>` to [role-helpers test](https://github.com/testing-library/dom-testing-library/blob/master/src/__tests__/role-helpers.js#L8).

### What happened:

Test continued to pass.

### What did you expect:

Test to fail, since `logRoles` would've logged a `link` implicit ARIA role with the anchor element within

### Reproduction:

1. Add `<a href="http://link.com" data-testid="a-link">link</a>` to [role-helpers test](https://github.com/testing-library/dom-testing-library/blob/master/src/__tests__/role-helpers.js#L8).
2. Run tests.

### Suggested solution:

I've been able to track it down to `makeElementSelector()`, where we extract `value` and use it to create the node selector expression:

```js
.map(({name: attributeName, value}) => `[${attributeName}=${value}]`)
```

([source](https://github.com/testing-library/dom-testing-library/blob/master/src/role-helpers.js#L19))

the resulting output for an anchor element is `a[href=undefined]`. Thus, if I change my link to `<a href="undefined">`, the elements is found and test starts to fail - as expected :)

However, not sure what the fix is. Why is the "=value" part needed? Should we just check for `value` to be truthy, and otherwise skip it? It'd generate a `a[href]` selector, which would come in handy to find links. Another option is to get the "http://whatever.com" part somehow, and add it to the element selector query.

What are your thoughts? I'm happy to tackle this if we reach an agreement 👍 I'm gonna open a draft PR with the updated test cases, so we can work from there.

Thanks!

afontcu:
**What**:

This PR tries to close #313 😇 

**Why**:

`getByRole('link')` wasn't returning valid anchor elements (read: anchors with `href` attribute).

**How**:

I simply check if `value` is set, and change the return format accordingly. I kept it simple and readable, feel free to suggest some JavaScript calisthenics to improve the resulting function!

**Checklist**:

- [x] Tests
- [x] Ready to be merged
   

I discovered that only `<a>` and `<area>` suffer from "value-less" attributes. Following is an output of elements provided by `aria-query` and their associated attribute array:

imatge

Thanks! 👋

### DIFF
--- a/src/__tests__/__snapshots__/role-helpers.js.snap
+++ b/src/__tests__/__snapshots__/role-helpers.js.snap
@@ -8,6 +8,14 @@ exports[`logRoles calls console.log with output from prettyRoles 1`] = `
 />
 
 --------------------------------------------------
+link:
+
+<a
+  data-testid="a-link"
+  href="http://whatever.com"
+/>
+
+--------------------------------------------------
 navigation:
 
 <nav
--- a/src/__tests__/role-helpers.js
+++ b/src/__tests__/role-helpers.js
@@ -6,6 +6,9 @@ afterEach(cleanup)
 function setup() {
   const {getByTestId} = render(`
 <section data-testid='a-section'>
+  <a href="http://whatever.com" data-testid="a-link">link</a>
+  <a>invalid link</a>
+
   <nav data-testid='a-nav' />
   
   <h1 data-testid='a-h1'>Main Heading</h1>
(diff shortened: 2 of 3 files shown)


kentcdodds:
@all-contributors please add @afontcu for code and tests

Issue #328, opened 2019-07-29
Don't log "here are the available roles" if there are no roles available.


kentcdodds:
Pretty much just need to do a bit of logic for this:

https://github.com/testing-library/dom-testing-library/blob/d719edd658ce6ad616f2632f378b3a7f4bf03897/src/queries/role.js#L33-L38

SavePointSam:
I would like to do this!

SavePointSam:
I believe this can be closed now! 🎉

kentcdodds:
Thanks!

PR #330, opened 2019-07-31
fix: getByRole error message, when there are no available roles

SavePointSam:
**What**:

Added an error message when there are no available roles.

**Why**:

Solves #328.

When unable to locate a role, an error message prints listing the available roles. When there are no available roles, the same message would print without a list of roles. This behavior is confusing/misleading to users.

**How**:

When the length of the string returned by `prettyRoles` is 0, we know there are no roles to be listed. The error message then changes based on that condition.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/queries/role.js
+++ b/src/queries/role.js
@@ -26,16 +26,26 @@ function queryAllByRole(
 
 const getMultipleError = (c, role) =>
   `Found multiple elements with the role "${role}"`
-const getMissingError = (container, role) =>
-  `
-Unable to find an element with the role "${role}"
 
+const getMissingError = (container, role) => {
+  const roles = prettyRoles(container)
+  let roleMessage
+
+  if (roles.length === 0) {
+    roleMessage = 'There are no available roles.'
+  } else {
+    roleMessage = `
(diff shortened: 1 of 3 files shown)


src/queries/role.js
```diff
@@ -26,16 +26,28 @@ function queryAllByRole(
...
+  const roles = prettyRoles(container)
+  let roleMessage
+
+  if (roles.length === 0) {
+    roleMessage = 'There are no available roles.'
+  } else {
+    roleMessage = `
 Here are the available roles:
 
   ${prettyRoles(container)
```
kentcdodds: We can avoid calculating the `prettyRoles` twice by putting `roles` in here instead of calling `prettyRoles(container)` again.

SavePointSam: Oops! I meant to do that. Haha.

SavePointSam: Corrected!

kentcdodds:
@all-contributors please add @SavePointSam for code and tests

PR #333, opened 2019-08-02
Fix outdated links on templates

afontcu:
I also removed the `react` version request from the template.


kentcdodds:
@all-contributors please add @afontcu for docs

Issue #314, opened 2019-07-10
Use debug module for logging instead of console.log


alexkrolick:
### Describe the feature you'd like:

DTL currently provides several debug methods that pretty print the DOM

What if instead of, or in addition to, these methods we added a debug scope using the debug npm module?

https://www.npmjs.com/package/debug

Before:
- write test
- get weird output
- add debug log calls to test

After:
- `DEBUG=dom:query:* yarn test` (on changed files or with a specific test regex)

kentcdodds:
Interesting idea! I wish we'd thought of this before making the `debug` and `log*` APIs. We could've said `debug` will use the `debug` module and `log*` uses `console.log`.

As it is, I don't think this is enough for a breaking change OR for a weird API name. Maybe we can reserve doing this for the next breaking change 🤔

Or maybe it's fine to make a breaking change for this. Thoughts?

pedroapfilho:
I don't see any issues making a breaking change for the debug API, you only use it once or twice and then remove it from the test itself.

timdeschryver:
Do we know when the next major release will be released? 
I'm fine either way, but if we have some changes lined up I think it would be better to add this one to the list.

kentcdodds:
> Do we know when the next major release will be released?

There's not currently any scheduled major release. And I don't think there are any other needs for a major release.

I have mixed feelings about treating this as an acceptable breaking change. I don't feel like it's necessary enough to warrant doing that. I'm not sure though.

kentcdodds:
After considering this further, I think we're too far along to make this kind of change. It would be pretty confusing to people I think 😬

alexkrolick:
🤔 would it be a breaking change to add support for the debug env option while leaving the explicit methods as an option?

kentcdodds:
After working on an exploratory PR for this, I'm starting to think this is going to be ok. It'll be a bit of an adjustment for folks, but I think they'll figure it out.

kentcdodds:
So here's my thinking. In DTL we'll have `debugDOM` and `logDOM` which basically amount to this:

```javascript
const debugDOM = (...args) => debug(prettyDOM(...args))
const logDOM = (...args) => console.log(prettyDOM(...args))
```

Where `debug` is: `const debug = getDebugger('testing-lib')`

In RTL, we'll make `debug` use `debugDOM` and we'll make a new method called `log` which calls `logDOM`. Thoughts?

alexkrolick:
[...]

PR #336, opened 2019-08-09
feat(logDOM): add logDOM export

kentcdodds:
**What**: add `logDOM` export

**Why**: I was working on #314, but I misunderstood it and changed gears. But I liked the refactor I was working on it.

**How**: Moved some logic around and now pretty-dom is responsible for determining default values for limit and options. I think it's a bit cleaner this way.

This is a breaking change because it basically moves `debugDOM` logic to `prettyDOM`. `debugDOM` was pretty useless after this change, so it's been removed.

It also adds an export for a new `logDOM` which does what it says.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/pretty-dom.js
+++ b/src/pretty-dom.js
@@ -1,21 +1,50 @@
 import prettyFormat from 'pretty-format'
+import {getDocument} from './helpers'
+
+function inCypress(dom) {
+  const window =
+    (dom.ownerDocument && dom.ownerDocument.defaultView) || undefined
+  return (
+    (typeof global !== 'undefined' && global.Cypress) ||
+    (typeof window !== 'undefined' && window.Cypress)
+  )
+}
+
+const inNode = () =>
+  typeof process !== 'undefined' &&
+  process.versions !== undefined &&
+  process.versions.node !== undefined
+
+const getMaxLength = dom =>
+  inCypress(dom) ? 0 : process.env.DEBUG_PRINT_LIMIT || 7000
 
 const {DOMElement, DOMCollection} = prettyFormat.plugins
 
-function prettyDOM(htmlElement, maxLength, options) {
-  if (htmlElement.documentElement) {
-    htmlElement = htmlElement.documentElement
+function prettyDOM(
+  dom = getDocument().body,
+  maxLength = getMaxLength(dom),
+  options,
+) {
+  if (maxLength === 0) {
+    return ''
+  }
+  if (dom.documentElement) {
+    dom = dom.documentElement
   }
 
-  const debugContent = prettyFormat(htmlElement, {
+  const debugContent = prettyFormat(dom, {
     plugins: [DOMElement, DOMCollection],
     printFunctionName: false,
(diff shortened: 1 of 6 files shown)


kentcdodds:
Actually, it looks like `debugDOM` wasn't exported at all anymore and now it is. So this is a new feature. I like it :)

src/pretty-dom.js
```diff
@@ -1,21 +1,50 @@
...
+const getMaxLength = dom =>
+  inCypress(dom) ? 0 : process.env.DEBUG_PRINT_LIMIT || 7000
 
 const {DOMElement, DOMCollection} = prettyFormat.plugins
 
-function prettyDOM(htmlElement, maxLength, options) {
-  if (htmlElement.documentElement) {
-    htmlElement = htmlElement.documentElement
+function prettyDOM(
+  dom = getDocument().body,
```
kentcdodds: If someone's not running in an environment where the document is globally available, then they have to provide a DOM node to log, otherwise we'll just default to `document.body` which I think is a good default. I imagine lots of people will just call `prettyDOM()` and that'll be helpful to them!

kentcdodds:
Dang it. I just realized that we _did_ export `debugDOM`, but it only returned a string that was formatted the way that I now have `prettyDOM` doing. So it wasn't that useful...

So here's what I'm thinking. Let's call this a breaking change (because it technically is) and we'll release this one with the RTL change here: https://github.com/testing-library/react-testing-library/pull/430

Thoughts?

typings/query-helpers.d.ts
```diff
@@ -20,7 +20,7 @@ export type AllByAttribute = (
 
 export const queryByAttribute: QueryByAttribute
 export const queryAllByAttribute: AllByAttribute
-export const debugDOM: (htmlElement: HTMLElement) => string
+export const logDOM: (htmlElement: HTMLElement) => void
```
kentcdodds: This is the breaking change.

Impact is minimal because people probably aren't relying on this for their actual tests and it's more of a workflow method (which I doubt many people use).

That said, some libraries built on top of DTL may use this method, and the upgrade strategy for them is to switch from `debugDOM` to `prettyDOM`.

src/query-helpers.js
```diff
@@ -2,33 +2,8 @@ import {prettyDOM} from './pretty-dom'
 import {fuzzyMatches, matches, makeNormalizer} from './matches'
 import {waitForElement} from './wait-for-element'
 
-/* eslint-disable complexity */
-function debugDOM(htmlElement) {
-  const limit = process.env.DEBUG_PRINT_LIMIT || 7000
```
alexkrolick: do we want to support this env var still?

kentcdodds: My changes keep it. Are you suggesting we should remove it?

alexkrolick: nope, I see it's moved into prettyDOM itself now 👍

PR #337, opened 2019-08-11
chore(typescript): move typings to DefinitelyTyped

alexkrolick:
See https://github.com/testing-library/react-testing-library/pull/437, #335 

Remove the TypeScript typedefs from this repo in favor of having the community maintain them at DefinitelyTyped.

**Why**

There are no core teammembers with the TS knowledge needed to maintain the types, and they are not well tested or integrated with the codebase.

---

See also https://github.com/reduxjs/redux/issues/3500


nickmccurdy:
It would be best to wait to merge until the types are published in DefinitelyTyped.

afontcu:
Are types gonna be published under "testing-library" namespace? We had [this PR](https://github.com/testing-library/vue-testing-library/pull/74) opened in Vue Testing Library to provide types locally, but I believe we should follow other libraries' directions and publish types next to them.

alexkrolick:
Yes, I think the way to do it is to mirror the package name directly in DefinitelyTyped

package.json
```diff
@@ -30,28 +29,26 @@
...
   },
   "files": [
-    "dist",
-    "typings"
+    "dist"
   ],
   "dependencies": {
     "@babel/runtime": "^7.5.5",
     "@sheerun/mutationobserver-shim": "^0.3.2",
+    "@types/testing-library__dom": "^6.0.0",
```
kentcdodds: Ok, so I've added this as a dep. Here's what things look like before any of these changes (with the current 6.0.0):

Screen Shot 2019-08-15 at 1 01 14 PM

And here's what it looks like before I added this dependency (with 6.0.1-alpha.0):

Screen Shot 2019-08-15 at 1 00 12 PM

And here's what it looks like after I added this dependency (with 6.0.1-alpha.1):

Screen Shot 2019-08-15 at 1 00 32 PM

So as far as the consumer is concerned, there's absolutely no change. And if someone wants to have more control over the version of the typings, then they can install `@types/testing-library__dom` themselves and `npm` or `yarn` will de-dupe to whatever version they want.

This is getting a 👍 from me. Let's do it!

PR #339, opened 2019-08-16
Add asyncUtilTimeout to config to allow custom default async timeout

andrewhillcode:
**What**:

Add a new config value called `asyncUtilTimeout` to allow users to adjust the default timeout for the async utils. 

**Why**:

There doesn't seem to be a nice way to change all of the async utils to use something other than `4500` timeout by default. This is useful for those who have changed the default jest timeout that the `4500` value was intended to be used with.

**Checklist**:

- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -3,6 +3,7 @@
 // './queries' are query functions.
 let config = {
   testIdAttribute: 'data-testid',
+  asyncUtilTimeout: 4500,
   // this is to support React's async `act` function.
   // forcing react-testing-library to wrap all async functions would've been
   // a total nightmare (consider wrapping every findBy* query and then also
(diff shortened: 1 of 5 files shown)


alexkrolick:
I'm for this

afontcu:
Looks like a nice addition 👍 

Quick question: what would happen if someone set a greater value than Jest's timeout? I guess DTL could warn user about that scenario.

FWIW, default value is 5000 and can be set using `jest.setTimeout()` ([source](https://jestjs.io/docs/en/jest-object#jestsettimeouttimeout)).

andrewhillcode:
Yeah I didn't think about the user increasing the timeout above Jest's. The DTL docs already mention that Jest's timeout is `5000` (which is why the DTL default is `4500`). 

Does this need more warnings than what is already on the docs? I plan to mention Jest's default when updating the config page to include `asyncUtilTimeout`

kentcdodds:
@all-contributors please add @andrewhillcode for code

andrewhillcode:
Cool, thanks Kent :)

PR #342, opened 2019-08-21
Upgrade wait-for-expect package

brapifra:
Related to https://github.com/TheBrainFamily/wait-for-expect/pull/20

### DIFF
--- a/src/__tests__/wait-for-dom-change.js
+++ b/src/__tests__/wait-for-dom-change.js
@@ -1,10 +1,16 @@
-import {waitForDomChange} from '../wait-for-dom-change'
 import {renderIntoDocument} from './helpers/test-utils'
 
 function importModule() {
-  jest.resetModules()
   return require('../').waitForDomChange
 }
+
+let waitForDomChange
+
+beforeEach(() => {
+  jest.useRealTimers()
+  jest.resetModules()
+  waitForDomChange = importModule()
+})
 
 test('waits for the dom to change in the document', async () => {
   const {container} = renderIntoDocument('<div />')
--- a/src/__tests__/wait-for-dom-change.js
+++ b/src/__tests__/wait-for-dom-change.js
@@ -53,7 +53,6 @@ test('uses real timers even if they were set to fake before importing the module
   jest.resetModules()
   jest.useFakeTimers()
   const importedWaitForDomChange = require('../').waitForDomChange
-  jest.useRealTimers()
 
   const {container} = renderIntoDocument('<div />')
 
@@ -62,4 +61,13 @@ test('uses real timers even if they were set to fake before importing the module
   await expect(importedWaitForDomChange({timeout: 200})).rejects.toThrow(
     /timed out/i,
   )
+})
+
+test("doesn't change jest's timers value when importing the module", () => {
+  jest.resetModules()
+  jest.useFakeTimers()
+  // eslint-disable-next-line
+  require('../').waitForDomChange
+
+  expect(window.setTimeout._isMockFunction).toEqual(true)
 })
--- a/src/__tests__/wait-for-dom-change.js
+++ b/src/__tests__/wait-for-dom-change.js
@@ -1,6 +1,11 @@
 import {waitForDomChange} from '../wait-for-dom-change'
 import {renderIntoDocument} from './helpers/test-utils'
 
+function importModule() {
+  jest.resetModules()
+  return require('../').waitForDomChange
+}
+
 test('waits for the dom to change in the document', async () => {
   const {container} = renderIntoDocument('<div />')
   const promise = waitForDomChange()
@@ -49,25 +54,47 @@ Array [
 `)
 })
 
-test('uses real timers even if they were set to fake before importing the module', async () => {
-  jest.resetModules()
+test('always uses real timers', async () => {
+  const expectElementToChange = async () => {
+    const importedWaitForDomChange = importModule()
+    const {container} = renderIntoDocument('<div />')
+
+    setTimeout(() => container.firstChild.setAttribute('id', 'foo'), 100)
+
+    const promise = importedWaitForDomChange({container, timeout: 200})
+
+    if (setTimeout._isMockFunction) {
+      jest.advanceTimersByTime(110)
+    }
+
+    await expect(promise).resolves.toMatchInlineSnapshot(`
+  Array [
+    Object {
+      "addedNodes": Array [],
+      "attributeName": "id",
+      "attributeNamespace": null,
+      "nextSibling": null,
+      "oldValue": null,
+      "previousSibling": null,
+      "removedNodes": Array [],
+      "target": <div
+        id="foo"
+      />,
+      "type": "attributes",
+    },
+  ]
+  `)
+  }
+
   jest.useFakeTimers()
-  const importedWaitForDomChange = require('../').waitForDomChange
-
-  const {container} = renderIntoDocument('<div />')
-
-  setTimeout(() => container.firstChild.setAttribute('id', 'foo'), 1000)
-
-  await expect(importedWaitForDomChange({timeout: 200})).rejects.toThrow(
-    /timed out/i,
-  )
+  await expectElementToChange()
+  jest.useRealTimers()
+  await expectElementToChange()
 })
 
 test("doesn't change jest's timers value when importing the module", () => {
-  jest.resetModules()
   jest.useFakeTimers()
-  // eslint-disable-next-line
-  require('../').waitForDomChange
+  importModule()
 
   expect(window.setTimeout._isMockFunction).toEqual(true)
 })
--- a/src/__tests__/wait-for-dom-change.js
+++ b/src/__tests__/wait-for-dom-change.js
@@ -60,7 +60,7 @@ Array [
 `)
 })
 
-test('always uses real timers', async () => {
+describe('timers', () => {
   const expectElementToChange = async () => {
     const importedWaitForDomChange = importModule()
     const {container} = renderIntoDocument('<div />')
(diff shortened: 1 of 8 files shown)


kentcdodds:
Will this be a breaking change? If so, I'll need to understand the justification for bumping the major version here and in all other consuming libraries for this change.

brapifra:
It's a breaking change for those who are using the lib and are setting jest's timers to `fake`. This scenario used to work because `wait-for-expect` was actually using jest's default `real` timers.
That is obviously an undesired behaviour. The library should always use the current setted timers (e.g if they are set to `fake`, it should use the `fake` timers).

However, as we discussed in https://github.com/TheBrainFamily/wait-for-expect/pull/20, it doesn't really make sense to use `fake` timers with `wait-for-expect`, so bumping the major version here should be fine.

Regarding the benefits, upgrading the package will allow to have `fake` timers as default and set them to `real` in specific test files. @kentcdodds

kentcdodds:
Hmmm... I wonder if our own `wait*` utilities should receive this same treatment at the same time. Care to take a look and see if that's necessary/fix it?

- https://github.com/testing-library/dom-testing-library/blob/master/src/wait-for-element.js
- https://github.com/testing-library/dom-testing-library/blob/master/src/wait-for-element-to-be-removed.js
- https://github.com/testing-library/dom-testing-library/blob/master/src/wait-for-dom-change.js

brapifra:
Sure. I will take a look at it tomorrow

brapifra:
Indeed, the `wait*` utilities needed the same treatment. I've added some changes and unit tests to make sure they have the correct behaviour.

kentcdodds:
I'd like to hear what other maintainers have to say about this breaking change.

To help clarify things, @brapifra, could you please give us a small example of a test that works today, but wont work with these changes, as well as one which does not work today and will work with these changes and why that is desirable?

brapifra:
@alexkrolick That part of the code does take care about Jest timers because is using the globals `setTimeout/clearTimeout/...`. These globals are mocked if jest timers are set to fake.

@kentcdodds This is an example that doesn't work today (when it should):
```typescript
import { waitForElement } from "@testing-library/dom";

// The timers have been defaulted to fake in package.json/jest.config.json

describe("Suite", () => {
  it("should not timeout", async () => {
    jest.useRealTimers();

    const container = document.createElement("div");
    await waitForElement(() => container.querySelector("div"));
  }); // throws timeout error
});
```

It doesn't work because the dom-testing-library doesn't get all these global timeout variables lazily. Therefore, the unit test times out as it's using fake timers when it should use the real ones that have been set inside the `it` block.

On the other hand, this is working now (although it shouldn't) and it will start failing if you accept my changes:
 ```typescript
import { waitForElement } from "@testing-library/dom";

// The timers are real by default if you don't change them in package.json/jest.config.json

describe("Suite", () => {
  it("should timeout", async () => {
    jest.useFakeTimers();

    const container = document.createElement("div");
    await waitForElement(() => container.querySelector("div"));
  }); // it passes
});
```

For the same reason, it's using the default real timers instead of the fake ones set inside the `it` block.

**Comment by kentcdodds** [comment]
Thank you for those examples. They're very helpful.

I think I've never run into this problem because I don't enable fake timers by default. Furthermore I also don't mix fake timers and the wait utilities in a single test (which I think we should recommend in the docs if we don't already).

So what we're doing here is changing the default I guess. When we do something like that, we need to consider if we accept and release this pull request:

1. The negative impact and the likelihood of that impact
2. The positive impact and the likelihood of that impact

Here's my analysis, I'd like to hear from others:

**The negative impact and the likelihood of that impact**: people who are mixing fake timers and wait utilities will have their tests break. I can't tell how many people this is, but it's probably a nontrivial amount of people, so I would say the likelihood of that impact is pretty high. In addition, the likelihood that people will continue to be impacted over time as they bump into this is nontrivial as well and the cost of teaching people to avoid this problem is nontrivial as well.

**The positive impact and the likelihood of that impact**: people who are using fake timers by default and using the wait utilities will be impacted. The number of people in that category today is 0. We know this because doing that is currently impossible. There may be a handful of people who _want_ to do that, but I think that is actually pretty rare because the default configuration in Jest is to use real timers and the vast majority of people do not change the default because they either don't know it's possible, or don't know why they would want to. So the positive impact will be for people who know that they can use fake timers by default and configure jest to do so. I'd say that the impact is relatively low, but the likelihood of a positive impact for that smaller group is high.

Ultimately, mixing fake timers and our wait utilities is pretty problematic. I definitely would like it if we could make enabling fake timers by default work, but I _don't_ want to force people to have to call `jest.useRealTimers()` every time they want to use an async utility. That would just be madness. So if that's what you have to do when you use fake timers by default, then we can't recommend enabling fake timers by default.

The fact is that async stuff (which is tested by our wait utilities) happens in an app _way_ more often than people needing fake timers.

So here's what I want: Can we think of a way to make this work:

```jsx
import { waitForElement } from "@testing-library/dom";

// The timers have been defaulted to fake in package.json/jest.config.json

describe("Suite", () => {
  it("should not timeout", async () => {
    // jest.useRealTimers(); // <-- this is not needed

    const container = document.createElement("div");
    await waitForElement(() => container.querySelector("div"));
  }); // it passes
});
```

Because I don't think that I want to make this change for what is a relatively small group of people and incur the negative impact on what is probably a relatively larger group of people (today and in the future) who will run into issues if we merge this PR.

alexkrolick:
@kentcdodds that's what I meant, too - can we make `wait` never use fake timers at all. Fake timers are meant to control "app time" not so much the constructs that run in the test env.

brapifra:
I agree with you that ideally this lib and others like `wait-for-expect` shouldn't depend on whether the timers have been mocked or not. However, as far as I know, that's not possible.

It's basically the same thing as having a library that is run in the browser: If someone else modifies any of the global variables that your library is using, it will probably stop working as expected.

brapifra:
Any thoughts about my previous comment? @kentcdodds @alexkrolick

brapifra:
> Could you make that happen?

I can, and I will... tomorrow :smile: I agree with you, the last solution is probably the best one.

kentcdodds:
Unfortunately, because the `wait-for-expect` package implemented the change that it did, we will no longer be receiving updates because I don't want to accept the change that was made over there. For that reason, I think we'll leave things at version 1 and if there are changes needed, we'll probably just remove the dependency and write it ourselves for this library (unless the change there is reverted).

brapifra:
I was planning to apply the same logic to `wait-for-expect` and open a PR in their repo. But I think it still makes sense to fix your own wait* upgrade functions and upgrade the `wait-for-expect` package once (if) they accept my change.

kentcdodds:
> But I think it still makes sense to fix your own wait* upgrade functions

Oh? I thought that all we needed to do was your suggestion above and that would result in no breaking change correct?

brapifra:
Oopss... yeah, that's what I was trying to say :sweat_smile: By adding my suggestion, all the wait* functions will work as expected without any breaking change.

kentcdodds:
Sweet! Thanks for talking us through this! I'll look forward to your updates when you get the chance :)

brapifra:
There you go @kentcdodds. I have added a few more tests to make sure that the library restores the jest timers after getting the time functions.
Let me know what you think

brapifra:
Ok, I think everything is properly changed and tested now. I've ended up creating a `runWithRealTimers` helper function that can be used whenever there is some piece of code that needs to be run without being affected by the current jest timers. It could also include some logic for other tests runners if needed in the future. @kentcdodds

src/__tests__/wait-for-dom-change.js
```diff
@@ -1,6 +1,11 @@
 import {waitForDomChange} from '../wait-for-dom-change'
 import {renderIntoDocument} from './helpers/test-utils'
 
+function importModule() {
```
kentcdodds: When `jest.resetModules()` or `jest.useFakeTimers()` is called, then that can result is breaking test isolation. So could we change this a bit:

```javascript
import {renderIntoDocument} from './helpers/test-utils'

let waitForDomChange

beforeEach(() => {
  jest.useRealTimers()
  jest.resetModules()
  waitForDomChange = require('../').waitForDomChange
})
```

This is probably how we should do it in the other tests as well

brapifra: Fair point :+1:

src/__tests__/wait-for-dom-change.js
```diff
@@ -48,3 +53,48 @@ Array [
 ]
 `)
 })
+
+test('always uses real timers', async () => {
+  const expectElementToChange = async () => {
```
kentcdodds: I think I'd prefer to put this function outside the test and have both cases as individual tests that call the function.

src/helpers.js
```diff
@@ -2,16 +2,40 @@ import MutationObserver from '@sheerun/mutationobserver-shim'
 
 const globalObj = typeof window === 'undefined' ? global : window
 
+function runWithRealTimers(callback) {
```
kentcdodds: Let's add a comment right above here indicating that currently we only support jest but could support other frameworks in the future.

src/helpers.js
```diff
@@ -2,16 +2,40 @@ import MutationObserver from '@sheerun/mutationobserver-shim'
...
+  const usingJestFakeTimers = globalObj.setTimeout._isMockFunction && !!jest
```
kentcdodds: I think this would need to be:

```javascript
globalObj.setTimeout._isMockFunction && typeof jest !== 'undefined'
```

Otherwise I _think_ we'd get an `Uncaught ReferenceError: jest is not defined` in other frameworks.

brapifra: Good catch!

src/helpers.js
brapifra: Done :)

brapifra:
Hmm... the nodev8 build job failed. That's weird :thinking:

kentcdodds:
It could be a race condition issue. That sometimes happens with these lower-level async tests 😬

brapifra:
Can you take a look at this? @alexkrolick

kentcdodds:
@all-contributors please add @brapifra for code and tests

kentcdodds:
Once `wait-for-expect` supports this, then I'd be happy to upgrade to that version :)

Issue #343, opened 2019-08-22
Can't find wrapped textarea with getByLabelText


amannn:
- `DOM Testing Library` version: 6.0.0
- `node` version: v12.3.1
- `npm` (or `yarn`) version: yarn@1.17.3

### Relevant code or config:

```js
const {getByLabelText} = render(
  <label>
    Label
    <textarea defaultValue="value" />
  </label>
);
getByLabelText('Label');
```

### What you did:

Run the test.

### What happened:

```
Unable to find a label with the text of: Label

<body>
  <div>
    <label>
      Label
      <textarea>
        value
      </textarea>
    </label>
  </div>
</body>

  113 |     </label>
  114 |   );
> 115 |   getByLabelText('Label');
      |   ^
  116 | });
  117 | 

  at getElementError (node_modules/@testing-library/dom/dist/query-helpers.js:22:10)
  at getAllByLabelText (node_modules/@testing-library/dom/dist/queries/label-text.js:104:43)
  at node_modules/@testing-library/dom/dist/query-helpers.js:59:17
  at Object.getByLabelText (src/components/Test/__tests__/Test-test.js:115:3)
```

Interestingly it works once you remove `defaultValue="value"`

### Reproduction:

https://codesandbox.io/s/react-testing-library-demo-olewk

### Problem description:

According to [the docs](https://testing-library.com/docs/dom-testing-library/api-queries#bylabeltext) this should work:
```html
<label>Username <input /></label>
```

### Suggested solution:

I guess the bug is somewhere here: https://github.com/testing-library/dom-testing-library/blob/master/src/queries/label-text.js

amannn:
I dug a bit in. It seems like you're using the `textContent` here to assert if the label matches:

https://github.com/testing-library/dom-testing-library/blob/master/src/queries/label-text.js#L20

However a textarea value is also part of the `textContent`. 

E.g.:
```
<label>Label<textarea>Value</textarea></label>
```
has a `textContent` of:
```
LabelValue
```

When you try to match for the label later, it fails because "value" is in the string as well.

I guess we need to remove textarea content from the string somewhere?

amannn:
Ok, I've got a PR up 🙂

pawbor:
Hi guys!

First of all, I appreciate your work. This lib is great! 

I've been scanning the code and found out the change introduced by @amannn. It fixes the initial bug but introduces another one:
```
it('passes', () => {
[...]

PR #344, opened 2019-08-22
Support textareas with children.

amannn:
**What**:

Fixes this bug: https://github.com/testing-library/dom-testing-library/issues/343#issue-484097775

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/queries/label-text.js
+++ b/src/queries/label-text.js
@@ -16,9 +16,17 @@ function queryAllLabelsByText(
 ) {
   const matcher = exact ? matches : fuzzyMatches
   const matchNormalizer = makeNormalizer({collapseWhitespace, trim, normalizer})
-  return Array.from(container.querySelectorAll('label')).filter(label =>
-    matcher(label.textContent, label, text, matchNormalizer),
-  )
+  return Array.from(container.querySelectorAll('label')).filter(label => {
+    let textToMatch = label.textContent
+
+    // The children of a textarea are part of `textContent` as well. We
+    // need to remove them from the string so we can match it afterwards.
+    label.querySelectorAll('textarea').forEach(textarea => {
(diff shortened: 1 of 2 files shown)


src/queries/label-text.js
```diff
@@ -16,9 +16,17 @@ function queryAllLabelsByText(
...
-  return Array.from(container.querySelectorAll('label')).filter(label =>
-    matcher(label.textContent, label, text, matchNormalizer),
-  )
+  return Array.from(container.querySelectorAll('label')).filter(label => {
+    let textToMatch = label.textContent
+
+    // The children of a textarea are part of `textContent` as well. We
+    // need to remove them from the string so we can match it afterwards.
+    label.querySelectorAll('textarea').forEach(textarea => {
+      textToMatch = textToMatch.replace(textarea.value, '')
```
alexkrolick: Hmm... What if the label is the same as the text value at that moment?

timdeschryver: I think we should be able to use `label.childNodes[0].textContent` here?

amannn: @alexkrolick `replace` only replaces the first occurence:

```js
'abcabc'.replace('abc', '') // "abc"
```

Therefore this should be fine.

@timdeschryver Sorry, I can't follow? Generally I think it's good to avoid `childNodes` so we support arbitrary nesting for both the `textarea` and the label. There could be some divs and spans wrapped around for example.

amannn:
Thanks for the review! Is there another maintainer who wants to have a look at this? Would be great to get this fixed, so I can remove a workaround in my code base 🙂. Thanks!

kentcdodds:
@all-contributors please add @amannn for code and tests

PR #352, opened 2019-09-19
feat(byRole): Exclude inaccessible elements

### LINKED ISSUE #350: add accessible? option to byRole query

Add a built-in method to query only accessible elements

### Suggested implementation:

`getByRole('listitem', { accessible: true /* default would be breaking */ })`

`accessible` checks for:
1. `:not([aria-hidden="true"])` and no ancestor matches `[aria-hidden="true"]`
2. not hidden visibility ~Confused it with focusability~

Better: It implements https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion

Implement custom query

Since `byRole` is mainly concerned about a11y it should also consider if the node in question is even accessible.

I would make this option configurable for low level tests. These elements are still displayed in the a11y pane in the chrome dev tools and are part of the a11y tree as far as I know. AT (only tested nvda+chrome, expect every AT to match that behavior) does not consider those e.g.

#### Strawman
```html

<nav aria-label="breadcrumbs">
  <ol>
    <li><a /></li>
    <li aria-hidden="true">/</li>
    <li><a /></li>
  </ol>
</nav>
```
will be announced as a list with 2 items while the a11y tree will include 3 elements with the `listitem` role.

eps1lon:
BREAKING
Certain<sup>1</sup> elements will no longer be included in `byRole` queries.
Examples:
* `<ul hidden />` will no longer by included in `*byRole('list')`
* `<main aria-hidden="true"><ul /><main>` will no longer return an element in `*byRole('list')` or `getByRole('main')`

**What**:

Implement https://www.w3.org/TR/core-aam-1.1/#exclude_elements2 with the following deviation:

* `role="none"` and `role="presentation"` is included (not following `MUST` criteria)
  I feel like this is more of a concern for an actual a11y tree implementation. `getByRole('presentation')` not returning anything even if the element is visible is probably confusing but I could see us throwing an error here. We could argue that one should use `getByTestId` instead since this makes it obvious you're not querying the a11y tree.
* descendants of "Children Presentational: True" are included (not following `SHOULD` criteria)

**Why**:

Closes #350

**How**:

Implementation is kept simple and mirrors the spec. Faster implementation is considered for a future PR.

**Checklist**:

      [docs site](https://github.com/alexkrolick/testing-library-docs) Will do once this is approved
- ~[ ] Typescript definitions updated~ Separate type PR is prepared
- [x] Tests
- [x] Ready to be merged
      

I'm open to removing the option and making it default. In any case it should be released in a breaking change since it is potentially breaking.

<sup>1</sup> excluded are elements that
* have `aria-hidden=true` or any of its ancestors
* have the `hidden` attribute or any of its ancestors
* have `display: none` or any of its ancestors
* have `hidden` as their computed `visibility`

### DIFF
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -26,7 +26,7 @@ test('logs error when there are no available roles', () => {
   expect(() => getByRole('article')).toThrowErrorMatchingInlineSnapshot(`
 "Unable to find an accessible element with the role "article"
 
-There are no accessible roles.
+There are no accessible roles. But there might be some inaccessible roles. If you wish to access them, then set the \`hidden\` option to \`true\`. Learn more about this here: https://testing-library.com/docs/dom-testing-library/api-queries#byrole
 
 <div>
   <div />
@@ -40,7 +40,7 @@ test('by default excludes elements that have the html hidden attribute or any of
   expect(() => getByRole('list')).toThrowErrorMatchingInlineSnapshot(`
 "Unable to find an accessible element with the role "list"
 
-There are no accessible roles.
+There are no accessible roles. But there might be some inaccessible roles. If you wish to access them, then set the \`hidden\` option to \`true\`. Learn more about this here: https://testing-library.com/docs/dom-testing-library/api-queries#byrole
 
 <div>
   <div
@@ -60,7 +60,7 @@ test('by default excludes elements which have display: none or any of their pare
   expect(() => getByRole('list')).toThrowErrorMatchingInlineSnapshot(`
 "Unable to find an accessible element with the role "list"
 
-There are no accessible roles.
+There are no accessible roles. But there might be some inaccessible roles. If you wish to access them, then set the \`hidden\` option to \`true\`. Learn more about this here: https://testing-library.com/docs/dom-testing-library/api-queries#byrole
 
 <div>
   <div
@@ -80,7 +80,7 @@ test('by default excludes elements which have visibility hidden', () => {
   expect(() => getByRole('list')).toThrowErrorMatchingInlineSnapshot(`
 "Unable to find an accessible element with the role "list"
 
-There are no accessible roles.
+There are no accessible roles. But there might be some inaccessible roles. If you wish to access them, then set the \`hidden\` option to \`true\`. Learn more about this here: https://testing-library.com/docs/dom-testing-library/api-queries#byrole
 
 <div>
   <div
@@ -104,7 +104,7 @@ test('by default excludes elements which have aria-hidden="true" or any of their
   expect(() => getByRole('list')).toThrowErrorMatchingInlineSnapshot(`
 "Unable to find an accessible element with the role "list"
 
-There are no accessible roles.
+There are no accessible roles. But there might be some inaccessible roles. If you wish to access them, then set the \`hidden\` option to \`true\`. Learn more about this here: https://testing-library.com/docs/dom-testing-library/api-queries#byrole
 
 <div>
   <div
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -3,9 +3,9 @@ import {render} from './helpers/test-utils'
 test('logs available roles when it fails', () => {
   const {getByRole} = render(`<h1>Hi</h1>`)
   expect(() => getByRole('article')).toThrowErrorMatchingInlineSnapshot(`
-"Unable to find an element with the role "article"
+"Unable to find an accessible  element with the role "article"
 
-Here are the available roles:
+Here are the accessible roles:
 
   heading:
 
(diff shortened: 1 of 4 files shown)


src/__tests__/role.js
```diff
@@ -24,12 +24,108 @@ Here are the available roles:
...
 `)
 })
+
+test('by default excludes elements that have the html hidden attribute or any of their parents', () => {
+  const {getByRole} = render('<div hidden><ul /></div>')
+
+  expect(() => getByRole('list')).toThrowErrorMatchingInlineSnapshot(`
+"Unable to find an accessible  element with the role "list"
+
+There are no accessible roles.
```
kentcdodds: We should probably add something to this message that says: "But there are some inaccessible roles. If you wish to access them, then set the hidden option to true. Learn more about this here: linktodocs.com"

src/queries/role.js
```diff
@@ -1,48 +1,64 @@
...
+Here are the ${hidden === false ? 'accessible' : 'available'} roles:
 
   ${roles.replace(/\n/g, '\n  ').replace(/\n\s\s\n/g, '\n\n')}
 `.trim()
   }
 
   return `
-Unable to find an element with the role "${role}"
+Unable to find an ${
+    hidden === false ? 'accessible ' : ''
```
kentcdodds: Remove the extra space.

```suggestion
    hidden === false ? 'accessible' : ''
```

eps1lon: I think I need to keep this and remove the unconditional space later on. Otherwise we get

```
Unable to find an  element with the role "${role}
                 ^^ double space
```
for `hidden: false`

src/__tests__/role.js
eps1lon: This would require a lot of refactoring if we want to actually determine if there are hidden roles.

Would it be sufficient to say "There **might** be inaccessible roles" instead?

src/queries/role.js
eps1lon: Extra space is still there (since I'm pretty sure it's required) but I fixed the double space issue.

src/__tests__/role.js
eps1lon: Extended the message and added a link to the docs but didn't implement displaying accessible vs. inaccessible roles.

src/role-helpers.js
```diff
@@ -3,6 +3,59 @@ import {prettyDOM} from './pretty-dom'
...
+    return true
+  }
+
+  // Remove once https://github.com/jsdom/jsdom/issues/2616 is fixed
+  const supportsStyleInheritance = computedStyle.visibility !== ''
+  let visibility = computedStyle.visibility
+
+  let currentElement = element
+  while (currentElement !== null) {
+    if (currentElement.hasAttribute('hidden')) {
```
kentcdodds: Is it permissible to have `hidden="false"`?

eps1lon: This would also be considered hidden. I should use the property here since it does not require knowledge of attributes vs. properties. You can also check this in your user agent's default stylesheet which should have 
```css
[hidden] {
  display: none;
}
```
This is only a recommendation though so we rely on the `hidden` attribute definition:

> The hidden attribute is a boolean attribute.

-- https://html.spec.whatwg.org/multipage/interaction.html#dom-hidden

> The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

-- https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute

eps1lon:
```suggestion
    if (currentElement.hidden === true) {
```

eps1lon: Would be interesting to see what react does with boolean attributes. It would be nice if it would add a warning since `hidden="false"` does not work as one might expect.

And it does:

> Warning: Received the string `false` for the boolean attribute `hidden`. The browser will interpret it as a truthy value. Did you mean hidden={false}?

kentcdodds:
I'm happy with this. Thank you!

eps1lon:
Unfamiliar with semantic-release: Should I have used another message format or was this intended to be a feature release?

kentcdodds:
I set the commit message when I use "Rebase & Merge." I intended this to be a feature release because I consider it highly unlikely to break anyone's existing tests. I'll take the brunt of any anger or frustration if I'm wrong.

jsphstls:
I do think a breaking change should have been a major version. I somehow picked this up with a new lockfile.

My breaking case is that I have a table with a scrolling body which essentially created two headers, one that is visible and one that is sr-only. The strategy is arguable. EIther way, I was used to finding both sets of `columnheaders` in the tests but now only one is found so the tests broke with seemingly unrelated changes. This change is ultimately good but the rollout strategy cost me some time.

eps1lon:
> sr-only

If this means "screen-reader only" then this is either a bug on our side or it detected an actual issue inside your markup. Could you share your markup + style that you would consider screen-reader only? This function should implement specification from the W3C which is either already implemented in assistive technology or will be.

jsphstls:
Apologies, I meant "one that is **visible and aria-hidden** and one that is sr-only". The changes in this PR appear to be working fine.

My point is that I did have tests that broke because of this PR.

eps1lon:
> My point is that I did have tests that broke because of this PR.

Well as kent said 

> I'll take the brunt of any anger or frustration if I'm wrong.

:smile:

kentcdodds:
Sorry! 😬

forgive me

shaunak:
Is there a reason this doesn't exist for `*byText` queries?

PR #355, opened 2019-09-20
Provide humane error messages for missing fireEvent() params

dominykas:
**What**:

Adds specific error message text when `element` / `node` / `event` params are missing for `fireEvent()`

**Why**:

In some cases you may be calling `fireEvent` without giving it an element, e.g.
while it is discouraged, in some cases it is necessary to use `querySelector` or similar or you might be accessing a `results[0]` in an empty array returned by `queryAllBy*`.

When you do that - `fireEvent()` will throw the good old `Cannot convert undefined or null to object`, which means you need to go and look for what exactly happened.

Providing a human readable, specific error message makes it easier to debug such failures.

**How**:

Throw a specific error after testing a condition.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/events.js
+++ b/src/__tests__/events.js
@@ -223,6 +223,24 @@ test('fires shortcut events on Window', () => {
   window.removeEventListener('message', clickSpy)
 })
 
+test('throws a useful error message when firing events on non-existent nodes', () => {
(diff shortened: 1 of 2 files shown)


kentcdodds:
@all-contributors please add @dominykas for code and tests

PR #362, opened 2019-09-25
fix: Crash in IE 11 when using logging roles

eps1lon:
**What**:

* Don't use [`Object.entries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries)

**Why**:

* Crashes in IE 11 when logging roles 

**How**:

* Use `Object.keys` instead

**Checklist**:

- ~[ ] I've prepared a PR for types targetting https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__dom~
- ~[ ] Tests~ Would require running tests in browsers. Since tests are still passing then it doesn't matter how it's implemented.
- [x] Ready to be merged

### DIFF
--- a/src/role-helpers.js
+++ b/src/role-helpers.js
@@ -135,8 +135,9 @@ function getRoles(container, {hidden = false} = {}) {
 function prettyRoles(dom, {hidden}) {
   const roles = getRoles(dom, {hidden})
 
-  return Object.entries(roles)
-    .map(([role, elements]) => {
+  return Object.keys(roles)
+    .map(role => {
+      const elements = roles[role]
       const delimiterBar = '-'.repeat(50)
       const elementsString = elements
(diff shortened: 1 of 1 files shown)


kentcdodds:
I think it's reasonable to expect that people have a polyfill for this API.

eps1lon:
> I think it's reasonable to expect that people have a polyfill for this API.

react itself does not require it and it feels weird to require it because of a testing library.

eps1lon:
We'll just not rely on error throwing in tests. It's brittle anyway. 

I still want to emphasize that requiring any polyfill is not viable for component libraries (since this means test under different conditions compared to what the user has) and problematic for application code for the same reasons unless you also include that polyfill in your application.

PR #363, opened 2019-09-25
fix: byLabel crashing in IE 11

eps1lon:
**What**:

* Don't use [`NodeList.prototype.forEach()
`](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach#Browser_Compatibility)

**Why**:

* Crashes in IE 11 when using `byRoles`

**How**:

* Wrap in `Array.from`

**Checklist**:

- ~[ ] I've prepared a PR for types targetting https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__dom~
- ~[ ] Tests~ Would require running tests in browsers. Since tests are still passing then it doesn't matter how it's implemented.
- [x] Ready to be merged

### DIFF
--- a/src/queries/label-text.js
+++ b/src/queries/label-text.js
@@ -21,7 +21,7 @@ function queryAllLabelsByText(
 
     // The children of a textarea are part of `textContent` as well. We
     // need to remove them from the string so we can match it afterwards.
-    label.querySelectorAll('textarea').forEach(textarea => {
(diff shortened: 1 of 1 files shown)


eps1lon:
> I never know whether array methods are supported on a NodeList or not (it seems to work at random even in my Chrome dev console.

I would've never known either :smile:

PR #364, opened 2019-09-28
feat: Export isInaccessible

eps1lon:
Looking for better names of this function.

**What**:

* Export helper method that is used in `byRoles` queries to exclude inaccessible elements

**Why**:

* To implement a custom test matcher like `expect(element).toBeInaccessible()`
  This is useful if you want to test relationships between elements which are established via `aria-controls`. We want to test that the value of `aria-controls` is valid by checking that `expect(document.getElementById(control.getAttribute('aria-controls'))).toBeInTheDocument()` but we want to make sure that at that point the element is still excluded from the a11y tree.

  Another use case might be if you queried the element with something other than `byRole` and you want to make sure the element is (in)accessible.

**How**:

* "just" export it

**Checklist**:

- [x] I've prepared a PR for types targetting https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__dom
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/role-helpers.js
+++ b/src/__tests__/role-helpers.js
@@ -166,6 +166,9 @@ test('getImplicitAriaRoles returns expected roles for various dom nodes', () => 
 })
 
 test.each([
+  ['<div />', false],
+  ['<div aria-hidden="false" />', false],
+  ['<div style="visibility: visible" />', false],
   ['<div hidden />', true],
   ['<div style="display: none;"/>', true],
   ['<div style="visibility: hidden;"/>', true],
--- a/src/__tests__/role-helpers.js
+++ b/src/__tests__/role-helpers.js
@@ -1,4 +1,9 @@
-import {getRoles, logRoles, getImplicitAriaRoles} from '../role-helpers'
+import {
+  getRoles,
+  logRoles,
+  getImplicitAriaRoles,
+  shouldExcludeFromA11yTree,
+} from '../role-helpers'
 import {render} from './helpers/test-utils'
 
 beforeEach(() => {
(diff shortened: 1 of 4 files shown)


src/__tests__/role-helpers.js
```diff
@@ -160,4 +165,18 @@ test('getImplicitAriaRoles returns expected roles for various dom nodes', () =>
   expect(getImplicitAriaRoles(input)).toEqual(['textbox'])
 })
 
+test.each([
+  ['<div hidden />', true],
+  ['<div style="display: none;"/>', true],
+  ['<div style="visibility: hidden;"/>', true],
+  ['<div aria-hidden="true" />', true],
+])('shouldExcludeFromA11yTree for %s returns %p', (html, expected) => {
```
kentcdodds: So far with these test cases, the function could simply be defined as: `() => true` and these tests would still pass. Could we have just one test case that returns `false`?

eps1lon:
> I'm not sure of the best name either. Maybe simply isHidden

What do you think about `isInaccessible`? (Although I'd be fine with the current name since it is currently coupled to the spec and I'm not sure we'll ever want to expand what it does).

`isHidden` has a visual implication which does not apply to `aria-hidden` or "visually hidden" (i.e. content with zero-width, zero-opacity etc).

src/__tests__/role-helpers.js
eps1lon: Done.

Although it could still be defined as `() => (window.i = (window.i || 0) + 1) > 3` :wink:

alexkrolick:
🤔  Could include the word ARIA in the name to indicate its relationship to that spec: `isAriaHidden`, `isAriaVisible`, ...?

eps1lon:
> 🤔 Could include the word ARIA in the name to indicate its relationship to that spec: `isAriaHidden`, `isAriaVisible`, ...?

`isAriaHidden` may be confused with `aria-hidden` and `isAriaVisible` might be even more confusing since `aria-visible` doesn't exist.

It's really just an implementation of "Excluding Elements from the Accessibility Tree". I thought about the name for a bit but I'd honest would just punt the name towards spec folks.

kentcdodds:
lol, actually, I think I like `isInaccessible` better.

eps1lon:
> lol, actually, I think I like `isInaccessible` better.

Ok let's go with that.

PR #365, opened 2019-10-01
fix: shouldExcludeFromA11yTree runs in IE11

briandipalma:
IE11 returns `undefined` for missing `parentElement`.

fix: In IE11 when an element lacks `parentElement` it returns `undefined` instead of `null`.

**Why**: Because otherwise you get these errors

```
TypeError: Unable to get property 'hidden' of undefined or null reference
	error properties: Object({ number: -2146823281 })
	    at <Jasmine>
	   at shouldExcludeFromA11yTree (eval code:824:5)
	   at Anonymous function (eval code:995:5)
	   at Anonymous function (eval code:28:7)
	   at filter (eval code:11:5)
	   at queryAllByRole (eval code:994:3)
```

**Checklist**:

- [ ] I've prepared a PR for types targetting https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__dom N/A
- [ ] Tests N/A
- [x ] Ready to be merged

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -385,6 +385,9 @@ test('queryAllByRole returns semantic html elements', () => {
       <h4>Heading 4</h4>
       <h5>Heading 5</h5>
       <h6>Heading 6</h6>
+      <svg role="img">
+        <rect width="100" height="100" />
+      </svg>
       <ol>
         <li></li>
         <li></li>
@@ -429,6 +432,7 @@ test('queryAllByRole returns semantic html elements', () => {
   expect(queryAllByRole('row')).toHaveLength(3)
   expect(queryAllByRole(/rowgroup/i)).toHaveLength(2)
   expect(queryAllByRole(/(table)|(textbox)/i)).toHaveLength(3)
+  expect(queryAllByRole(/img/i)).toHaveLength(1)
 })
 
 test('getAll* matchers return an array', () => {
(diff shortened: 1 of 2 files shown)


eps1lon:
Could you include a reproduction that fails with IE 11? This would help a lot :pray:

briandipalma:
Yes it seems like it's not implemented on `Node` `Only supported on Element.` from https://developer.mozilla.org/en-US/docs/Web/API/Node/parentElement

But the element that failed for us was an SVG and I found this: https://stackoverflow.com/a/36270354/1927079

```In Internet Explorer, parentElement is undefined for SVG elements, whereas parentNode is defined.```

Do your tests run in IE11?

eps1lon:
> But the element that failed for us was an SVG and I found this:

Interesting. I'll verify that and add it to the MDN docs since `SVGElement` inherits from `Element`.

> Do your tests run in IE11?

We don't run any tests in browsers and so far browser support hasn't been a priority from what I can tell. I use this library in browser tests though so anytime I do find one I'll try to fix it (#362, #363).

briandipalma:
I'll try and add a test that covers the failure I see here.

briandipalma:
Added a test but this would only ever fail in IE11.

briandipalma:
Updated the tests, is this what you were thinking?

briandipalma:
Is this OK?

PR #367, opened 2019-10-02
Add strict check for `dom` type

mbelsky:
Hey,

I face an issue and come with this pr :)

**What**:

Add strict check for `dom`'s `outerHTML` field

**Why**:

To improve DX. 

As a new user I've called `debug(x)` where `x` was a result of `queryAll*`. I've missed that `queryAll*` returns an array and got `TypeError: Cannot read property 'length' of undefined` in my terminal.

**Checklist**:

- [x] I've prepared [a PR for types](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38901)
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/pretty-dom.js
+++ b/src/pretty-dom.js
@@ -20,16 +20,32 @@ const getMaxLength = dom =>
 
 const {DOMElement, DOMCollection} = prettyFormat.plugins
 
-function prettyDOM(
-  dom = getDocument().body,
-  maxLength = getMaxLength(dom),
-  options,
-) {
+function prettyDOM(dom, maxLength, options) {
+  if (!dom) {
+    dom = getDocument().body
+  }
+  if (typeof maxLength !== 'number') {
+    maxLength = getMaxLength(dom)
+  }
+
   if (maxLength === 0) {
     return ''
   }
   if (dom.documentElement) {
     dom = dom.documentElement
+  }
+
+  let domTypeName = typeof dom
+  if (domTypeName === 'object') {
+    domTypeName = dom.constructor.name
+  } else {
+    // To don't fall with `in` operator
+    dom = {}
+  }
+  if (!('outerHTML' in dom)) {
+    throw new TypeError(
+      `Expected an element or document but got ${domTypeName}`,
+    )
   }
 
   const debugContent = prettyFormat(dom, {
--- a/src/__tests__/pretty-dom.js
+++ b/src/__tests__/pretty-dom.js
@@ -26,14 +26,16 @@ test('prettyDOM supports truncating the output length', () => {
 })
 
 test('prettyDOM defaults to document.body', () => {
+  const defaultInlineSnapshot = `
+  "<body>
+    <div>
+      Hello World!
+    </div>
+  </body>"
+`
   renderIntoDocument('<div>Hello World!</div>')
-  expect(prettyDOM()).toMatchInlineSnapshot(`
-    "<body>
-      <div>
-        Hello World!
-      </div>
-    </body>"
-  `)
+  expect(prettyDOM()).toMatchInlineSnapshot(defaultInlineSnapshot)
+  expect(prettyDOM(null)).toMatchInlineSnapshot(defaultInlineSnapshot)
 })
 
 test('prettyDOM supports receiving the document element', () => {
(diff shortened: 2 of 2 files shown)


src/pretty-dom.js
```diff
@@ -25,6 +25,11 @@ function prettyDOM(
   maxLength = getMaxLength(dom),
   options,
 ) {
+  if (!(dom instanceof HTMLDocument) && !(dom instanceof HTMLElement)) {
```
eps1lon: I suppose the error is thrown at `dom.outerHTML.length` which means `Element` is sufficient. Otherwise we would also throw on svg elements.

Do we need to consider cross-realm instanceof checks here as well? Getting the window of an element would not work if we didn't get an Element in the first place. I think checking for `typeof dom.outerHTML === 'string'` might be sufficient here in asserting that the given value implements `Element`.

mbelsky: > I suppose the error is thrown at `dom.outerHTML.length` which means `Element` is sufficient. Otherwise we would also throw on svg elements.

Unfortunately it isn't because `document` isn't an instance of `Element`

> I think checking for `typeof dom.outerHTML === 'string'` might be sufficient here in asserting that the given value implements `Element`.

I absolutely agree that `typeof dom.outerHTML === 'string'` checking is enough however it may hide a real problem from a developer

eps1lon: > Unfortunately it isn't because document isn't an instance of Element

Which is why you would keep the HTMLDocument check.

Cross-realm problem still exists.

> I absolutely agree that typeof dom.outerHTML === 'string' checking is enough however it may hide a real problem from a developer

That would be? 

IMO runtime type checking is not necessary here. If people care about type checking then you already have very good solutions with flow or typescript.

mbelsky: Types may lies. Right now `dom` typed as `HTMLElement` 

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/403ac51161e5773d0ec7eb965004e18e9d4a03c8/types/testing-library__dom/pretty-dom.d.ts#L3

However there is a test for rendering `document`

https://github.com/testing-library/dom-testing-library/blob/a401bb50c8301c27b2698d299b3ebc1483cc7633/src/__tests__/pretty-dom.js#L39-L40

Runtime checking isn't a silver bullet however it allows to fail fast when a library's user use it as not expected and teaches she how to do it right.

eps1lon: > Types may lies

Did it occur to you that it was a mistake instead of an intention to specify something wrong?

mbelsky: Sure, it did.

For me as a developer it's so cool to have strict contracts in a library. What do you think about an alternative conditional here like:

```diff
- if (!(dom instanceof HTMLDocument) && !(dom instanceof HTMLElement)) {
-     throw new TypeError(
-      'Debuggable value should be an instance of HTMLDocument or HTMLElement.',
-    )
- }
+ if (!('outerHTML' in dom)) {
+     throw new TypeError(
+      'Debuggable object should have `outerHTML` field like an HTMLElement instance.',
+    )
+ }

```

eps1lon: Sounds better but you're still using `HTMLElement` which is too narrow. No need throw so many technicallaties around "Expected an element or document but got X".

mbelsky: Thanks for suggestion. I've updated this check and add a test

mbelsky:
I didn't extend `pretty-dom.js` functional and cover added condition with a test. However ci fails with:

```
Jest: "global" coverage threshold for branches (100%) not met: 99.59%
...
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @testing-library/dom@0.0.0-semantically-released validate: `kcd-scripts validate`
npm ERR! Exit status 1
npm ERR! 
```

Does it mean that I should write some tests for `pretty-dom.js`?

src/pretty-dom.js
```diff
@@ -31,6 +31,13 @@ function prettyDOM(
   if (dom.documentElement) {
     dom = dom.documentElement
   }
+  if (!('outerHTML' in dom)) {
+    const domType = typeof dom
+    const domTypeName = domType === 'object' ? dom.constructor.name : domType
```
eps1lon: This branch is likely not covered. Also: This will throw on `null` because `typeof null === 'object'`.

mbelsky: Nice catch. Thank you for explaining.

src/pretty-dom.js
```diff
@@ -20,18 +20,34 @@ const getMaxLength = dom =>
 
 const {DOMElement, DOMCollection} = prettyFormat.plugins
 
-function prettyDOM(
-  dom = getDocument().body,
-  maxLength = getMaxLength(dom),
-  options,
-) {
```
mbelsky: `inCypress` has failed with `null` parameter so in addition I replaced arguments default initializations.

kentcdodds:
@all-contributors please add @mbelsky for code and tests

PR #373, opened 2019-10-06
Add more robust selector handling for getByLabelText

### LINKED ISSUE #372: Inconsistent behavior of `selector` option
- `DOM Testing Library` version: 6.3.0
- `node` version: 8.12.0 (& whatever codesandbox is on)
- `npm` (or `yarn`) version: 6.4.1 (& whatever codesandbox is on)

### Relevant code or config:

I have two main examples:

```jsx
{/* Label associated with aria-labelledby */}
<label id="label1">Hi</label>
<span aria-labelledby="label1">Bye</span>
<input id="input1" aria-labelledby="label1" />

it('should find the input element', () => {
  const result = render(<Comp />);
  result.getByLabelText('Hi', { selector: 'input' });
  // Throws because multiple elements were matched
});

```

```js
{/* Label associated by nesting */}
<label>
  <span>This one works</span>
  <input />
</label>

it('should not work because the label is nested in a span', () => {
  const result = render(<Comp />);
  result.getByLabelText('This one works');
  // Works fine
});
```

The documentation [here](https://testing-library.com/docs/dom-testing-library/api-queries#bylabeltext) suggests that the second one should not work and that the `selector` option is necessary.

### What you did:

I was trying to test an interaction with a dropdown I made using Downshift. My hope was that I could select the input element using `getByLabelText` but that does not appear to be the case. In Downshift, multiple elements (e.g. the input element, the option list itself, the root div) get associated with the label via `aria-labelledby`. When I tried to filter the options down using the `selector` option, that lead me down this rabbit hole.

### What happened:

It did not work the way I had hoped (filtering the results down based on the selector I provided). However, it also doesn't appear to work the way the documentation says it's supposed to work. The documentation I linked about provides this example:

```jsx
<label> <span>Username</span> <input /> </label>
```

And says this:
```
It will NOT find the input node for label text broken up by elements. For this case, you can provide a selector in the options:
```

So it appears that the `selector` option for the `getByLabelText` is intended for filtering out elements that are _inside_ labels. But as can be seen the codesandbox below, it is not actually necessary in the example given.

### Reproduction:

https://codesandbox.io/embed/react-testing-library-demo-jipd1

### Problem description:

It seems like the purpose and behavior of the `selector` option is inconsistent. In the reproduction above, you can see that it works with the `getByText` query the way I, personally, would expect it to. If two elements have the same text but only one matches the provided selector, that one gets returned. For `getByLabelText` though, it does not work that way. If I have multiple elements that point to the same label via `aria-labelledby` and I do a `getByLabelText`, providing a `selector` doesn't do anything and it throws an error unless I use `getAllByLabelText`.

### Suggested solution:

I think it would make sense for it to behave more like the way it does for `getByText`. So it would:

1. Find all labels with the matching text
2. Find all elements that are labelled by those labels
3. Filter down the results based on the provided selector.

If you agree, I would be happy to take a stab at this and put up a PR.

airjp73:
**What**:

Adds more robust handling of the `selector` option that can be passed to the `getByLabelText` family of queries. Fixes #372.

**Why**:

More info in related issue (#372)

**How**:

Just some tweaking to `label-text.js`. It wasn't trying to find _all_ the possible matches, and instead would short-circuit when it found something. For example, if the label had a `for` attribute, it wouldn't also check for elements that were `aria-labelledby` that label. Then I also added another step at the end where it filters down the elements it found based on the `selector` option.

**Checklist**:

- [X] Documentation added to the [docs site](https://github.com/alexkrolick/testing-library-docs) ([this PR](https://github.com/testing-library/testing-library-docs/pull/282))
- [ ] I've prepared a PR for types targetting https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__dom N/A
- [X] Tests
- [X] Ready to be merged


src/queries/label-text.js
```diff
@@ -89,7 +94,13 @@ function queryAllByLabelText(
...
   )
 
-  return Array.from(new Set([...labelledElements, ...ariaLabelledElements]))
+  const allMatches = Array.from(
+    new Set([...labelledElements, ...ariaLabelledElements]),
+  )
+  const elementsMatchingSelector = new Set(container.querySelectorAll(selector))
+  return allMatches.filter(matchingElement =>
+    elementsMatchingSelector.has(matchingElement),
+  )
```
kentcdodds: I think we'd be better off with `matches` here: https://developer.mozilla.org/en-US/docs/Web/API/Element/matches

src/__tests__/element-queries.js
```diff
@@ -212,6 +212,36 @@ test('can get sibling elements with aria-labelledby attrib ute', () => {
...
+      <span>Test Label</span>
+      <input id="input1" />
+    </label>
+  `)
+
+  const result = getAllByLabelText('Test Label', {selector: 'input'})
+  expect(result).toHaveLength(1)
+  expect(result[0].id).toBe('input1')
+})
+
```
kentcdodds: Could you add another test for the situation I found myself in where we had two elements that had the same label text?

```javascript
  const {getByLabelText} = render(`
    <label>
      Test Label
      <input />
    </label>
    <label>
      Test Label
      <textarea></textarea>
    </label>
  `)
  // select the input
```

Thanks!

src/queries/label-text.js
airjp73: I had a feeling this wasn't the best way. Probably could have done a little more digging haha.

airjp73:
> This is looking good.
> 
> I _want_ to make this a non-breaking change, but it really is a breaking change 🙁

Yeah, it totally is. Even one of the cases I pointed out in the issue
```
const {getByLabelText} = render(
  <label>
    <span>Username</span>
    <input />
  </label>
);

// This worked before even though the documentation said it didn't
// and now you need to provided a selector
const input = getByLabelText('Username');
```

kentcdodds:
Wait. So to be clear, what you're saying is that with these changes I can't get the input in your example with `getByText(/username/i)` (without the selector)?

airjp73:
More specifically, the query will return the input _and_ the span as matches. So if you only want the input it needs to be filtered down with the selector.

airjp73:
What do you think the correct behavior is in that situation? The `textContent` of the label could be made up of any number of elements, so I'm not sure how we would filter it by default.

Example:
```js
const {getAllByLabelText} = render(`
  <label>
    <h1>Test</h1>
    <h2>Label</h2>
    <input id="input1" />
  </label>
`)

// Returns 3 elements [h1, h2, input]
const result = getAllByLabelText(/Test Label/i)
```

alexkrolick:
That case with the h1/h2 should only return the input element. `getByLabelText` returns the labelled elements, not the label content.

airjp73:
That definitely makes sense. How should we qualify which children are the label and which are labeled? The current behavior with label children (before this change), is to do `label.querySelector(selector)` where `selector` defaults to `*`. So all children of a matched label are picked up.

Does one of these make sense?
- Default `selector` to grab inputs and textareas
- Special case label children to only pick up labels and textareas
- Figure out which elements, specifically, contain the text we matched on and exclude them

airjp73:
I went ahead with something like my second option. If that's not right I can totally change it :+1:

So it will now pick up
* Any element referenced by a matched label's `for` attribute
* Any element that is `aria-labelledby` a matched label
* Any `input` or `textarea` that is a child of a matched label

Then those results are filtered down based on the provided `selector` option.

src/queries/label-text.js
```diff
@@ -51,21 +52,25 @@ function queryAllByLabelText(
...
-        )
+        container
+          .querySelectorAll(`[aria-labelledby~="${label.getAttribute('id')}"]`)
+          .forEach(element => elementsForLabel.push(element))
       }
       if (label.childNodes.length) {
         // <label>text: <input /></label>
-        return label.querySelector(selector)
+        label
+          .querySelectorAll('input, textarea')
```
kentcdodds: This is insufficient. We need to select *every* form control element type. Unfortunately there's not a common class that all form control elements inherit from, but I think we'd be safe if we found every element that supports [the `.labels` property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/labels).

I'm not sure of the best way to do this, but here's a list of all the element types: https://developer.mozilla.org/en-US/docs/Web/API (Everything named something like HTML*Element).

airjp73:
Happy to help!

src/queries/label-text.js
airjp73: I'll look into that and see if I can find all of them. :+1:

airjp73:
I think I got all of them. One thing that's a little unsatisfying is it necessitates updating the list of supported elements manually if another element is added.

airjp73:
Somewhat related to this change, I wonder if it's worth doing a quick look at the other queries to make sure the behavior of the selector option is consistent across them?

kentcdodds:
I believe all the other queries are directed at the element that's actually returned, rather than taking a hop like this one does, but if you'd like to double check that'd be great :)

kentcdodds:
Another thing that we should maybe consider is adding a `labelSelector` option which works the way the original `selector` worked. We wouldn't even have to document it (because I think it's unlikely people would use it), but that way people would have a migration path if they used the old `selector` option.

airjp73:
I'm not fully sure what the `labelSelector` would look like.

The main case where people will run into issues is if they have a label that does multiple of these things:
* Has a `for` prop
* Has an `id` prop
* Has children

Before, if the label had a `for` prop, it would only query for that one element and never query for anything with `aria-labelledby`. And if the label had an `id`, it would query for elements that are `aria-labelledby` the label and never query the children of the label. Now all those conditions can be checked for the same label.

So if we added a `labelSelector`, would we also change that -- only query for one of those cases for any given label? Or would it just be a selector that only applies to label children and not those other cases?

airjp73:
> I believe all the other queries are directed at the element that's actually returned, rather than taking a hop like this one does, but if you'd like to double check that'd be great :)

I checked the docs and it looks like `ByText` and `ByLabelText` are the only queries that have the `selector` option anyway so I don't think this is anything we need to worry about.

kentcdodds:
Awesome. Thanks for checking. I think we're about ready to make this happen. But since it's a breaking change, I want to take advantage of that churn to make other breaking changes we've been wanting to make (#376). I'd love it if you want to weigh in :)

weyert:
I am having the same issue when working with Downshift I would love to switch back to getLabelByText. My current solution is to use getByRole('textbox') but that doesn't make me excited.

kentcdodds:
I'm finally back from a bit of a break. I'll be taking a hard look at breaking shipping some breaking changes soon.

kentcdodds:
Sorry this has taken so long. Really hoping to get to this soon. It's just been a beast to make enough time for this major version bump. And I try to avoid doing those 😅

kentcdodds:
I made a little mistake with the auto-release here, but the latest is still pointing to the right version so hopefully we don't have issues with people getting 6.16.0 by mistake.

In any case, you will soon be able to test this via `npm install @testing-library/dom@beta`

airjp73:
Awesome! Glad I was able to help. :)

dreyks:
hmm... I've got a bit lost on where this change was introduced, i guess it was [6.16.0](https://github.com/testing-library/dom-testing-library/releases/tag/v6.16.0) but that release is tagged as pre-release and this change is highlighted once more in [v7.0.0-beta.1](https://github.com/testing-library/dom-testing-library/releases/tag/v7.0.0-beta.1), where it was properly marked as BREAKING

unfortunately, that mention didn't make it to the final [7.0.0](https://github.com/testing-library/dom-testing-library/releases/tag/v7.0.0) release notes, so I suggest we add it there cc @kentcdodds 

also this is breaking in one more way not mentioned in this PR: now `getByLabelText` finds all inputs that are inside a given label, which leads to `TestingLibraryElementError: Found multiple elements with the text` in cases where there actually are several inputs inside a single label

so this change can be breaking even for those who don't use `selector`

kentcdodds:
Whoops! Sorry about that. We had a few bumps in the release and missed a few things in the notes 😬 I'll get that added asap.

airjp73:
@kentcdodds, do you want me to put in a PR to address @dreyks suggestion that it return the first input instead of all of them in the case that a label contains multiple inputs?

kentcdodds:
That seems reasonable to me, yes. Thank you! I'll consider this a bug fix rather than a breaking change.

Issue #379, opened 2019-10-10
wait* utilities take more time since 9.2.0 (running into Timeout)


kissthom:
- `react-testing-library` version: 9.2.0
- `react` version: 16.9.5
- `node` version: 12.6.0
- `yarn` version:  1.19.0

### What you did:

Upgraded from version 9.1.4 to 9.2.0. 

### What happened:

Some of our tests which are using the "wait*" utilities provided by the library are running into jest timeout after the upgrade. In some cases waiting for a dom element to appear takes 7-8x times more (~3 seconds) as before with version 9.1.4. 

### Reproduction:

I've created a simple example to reproduce the problem although with this example the difference is not so critical as in our application.  

https://codesandbox.io/s/rtl-wait-slow-g7c21

Steps to reproduce:
1. Run the tests and check out the results in the console (average time for waiting for select.options elements to appear 40-60ms)
2. Switch the rtl version to 9.2.0
3. Run the tests and check out the results again (2-3 times slower) (average time for waiting for select.options elements to appear 110-150ms)

Any help is appreciated.

eps1lon:
Yeah that query can be quite expensive for which I don't really have an easy solution. We can probably optimize it a bit but the general runtime implication of traversing the whole tree will always be there.

So I have some ideas how to make this faster but this is not a priority for me. If somebody wants to work on this then I have a couple of suggestions:
* don't use querySelectorAll but traverse top-down and skip subtrees we know will be inaccessible (aria-hidden, hidden)
* cache `isAccessible` results (I think we have O(n^2) calls it to where n is the number of nodes)

As to your perspective I would generally ask if you really need waitFor queries. Personally I don't find them very useful since I'd like to see explicitly in the test how long something takes in my UI. So far any waiting I used could be replaced with explicitly advancing the timer and then making the assertion. I think this applies to your repro really well. At least I would like my test to fail if it takes 1s instead of .1s to transition. Now this applies mostly to your transitions but not really if you're waiting for some API. Maybe you should rather wait a fixed amount and then expect a certain state. After all your user doesn't want to look for 20s at a spinner either. But I understand that there are some use cases where wait makes more sense. 

Another option may be to reduce to polling time i.e. is it important to assert every frame (16ms) or is every 10 frames (160ms) enough?

[...]

PR #381, opened 2019-10-13
feat(byRole): Improve performance

eps1lon:
**What**:

Addresses #379

**Why**:

Queries can be used when polling the DOM. If the query is slower than the polling rate we'll run into timeouts

**How**:

Since |elements| >> |elements with role| and getting the role is far less expensive than deciding if an element is part of the a11y tree it's far cheaper to query the elements for the role before checking if an element is part of the a11y tree.

**Checklist**:

- [ ] ~I've prepared a PR for types targetting https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__dom~
- [ ] Tests, testing in Codesandbox CI
- [ ] Ready to be merged

### DIFF
--- a/src/queries/role.js
+++ b/src/queries/role.js
@@ -14,9 +14,6 @@ function queryAllByRole(
   const matchNormalizer = makeNormalizer({collapseWhitespace, trim, normalizer})
 
   return Array.from(container.querySelectorAll('*'))
-    .filter(element => {
-      return hidden === false ? isInaccessible(element) === false : true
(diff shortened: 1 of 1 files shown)


kentcdodds:
Whoops! Didn't realize that this code isn't full covered by tests 😬

eps1lon:
Hm I wanted to see how this Codesandbox CI works. Guess I have to add a config anyway :smile:

PR #383, opened 2019-10-13
chore: Add codesandbox ci config

eps1lon:
Please don't merge this yet

Testing:
* [x] if codesandbox CI runs
* [x] how it works with transitive dependencies (the codesandbox from kent only has a direct dependency on `@testing-library/react` but we would like to resolve its dependency on `@testing-library/dom` to the canary from this PR as well)

  Edit: It doesn't at the moment so we use https://github.com/kentcdodds/dom-testing-library-with-anything

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,3 +1,3 @@
 {
-  "sandboxes": ["github/kentcdodds/dom-testing-library-with-anything"]
+  "sandboxes": ["github/kentcdodds/react-testing-library-examples"]
 }
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,0 +1,3 @@
+{
+  "sandboxes": ["vanilla", "github/kentcdodds/react-testing-library-examples"]
+}
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,3 +1,3 @@
 {
-  "sandboxes": ["vanilla", "github/kentcdodds/react-testing-library-examples"]
+  "sandboxes": ["vanilla"]
 }
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,3 +1,3 @@
 {
-  "sandboxes": ["vanilla"]
+  "sandboxes": ["github/kentcdodds/dom-testing-library-with-anything"]
 }
--- a/src/index.js
+++ b/src/index.js
@@ -25,7 +25,3 @@ export {
   queries,
   queryHelpers,
 }
-
-export function csb() {
-  return 'yay!'
-}
--- a/src/index.js
+++ b/src/index.js
@@ -25,3 +25,7 @@ export {
   queries,
   queryHelpers,
 }
+
+export function csb() {
+  return 'yay!'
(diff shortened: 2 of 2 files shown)


eps1lon:
It might be that either only mui-org is whitelisted or it doesn't pick up that `@testing-library/dom` is a transitive dependency.

@CompuIves Could you take a look or forward this to someone who's working on codesandbox CI?

alexkrolick:
I assume it will only ru once it is on master

eps1lon:
> I assume it will only run once it is on master (?)

It will run on PRs.

I forgot the e-mail back though. Repositories need to be explicitly whitelisted.

CompuIves:
This is a really interesting case! So it would make sense if we add an extra option or something like:

```
"dependencyResolutions": {
  "@testing-library/react": "canary"
}
```

so the sandbox adds this dependency as this version?

eps1lon:
@CompuIves Yes. Though you probably meant `"@testing-library/dom": "canary"`. I'd rather see codesandbox implement the `resolutions` field which improves interop with `yarn`.

Or did you mean we need to specify each dependency individually where we want codesandbox to resolve indirect dependencies to the canary?

CompuIves:
> Or did you mean we need to specify each dependency individually where we want codesandbox to resolve indirect dependencies to the canary?

Right, with my example it would additionally install `@testing-library/react@canary` together with the built version of `@testing-library/dom`. I could also implement the resolution field in the bundler if that's better, but it would take longer.

eps1lon:
> Right, with my example it would additionally install @testing-library/react@canary

There wouldn't be one. And even if you would need to build it so that the `dependencies` entry for `@testing-library/dom` points to the canary.

CompuIves:
Hmm, I'm not sure if we're talking about the same thing. Is this to make `testing-library/dom-testing-library` work with CodeSandbox CI so you can test it for PRs? And with `canary` you mean the `canary` tag on npm or the latest build from CodeSandbox CI?

eps1lon:
/react is depending on /dom. I want codesandbox to resolve /dom to the built version of codesandbox CI. Right now we have two versions when building /dom in codesandbox: the release on npm and the release on codesandboc

kentcdodds:
Let me know when you're ready for me to merge this.

eps1lon:
> Let me know when you're ready for me to merge this.

@kentcdodds I'm using react-testing-library-examples since I couldn't find another working codesandbox that uses `@testing-library/dom`. `kentcdodds/dom-testing-library-with-anything` isn't working in codesandbox.

eps1lon:
This isn't useful at the moment since transitive dependencies of packages on npm are not resolved to the version built in CodeSandbox CI even with `resolutions`: https://codesandbox.io/s/jolly-ellis-3wjgm

kentcdodds:
> I'm using react-testing-library-examples

That's fine

eps1lon:
Now it's working again :smile: Let's merge this and see if it happens again.

kentcdodds:
Thanks!

kentcdodds:
@all-contributors please add @eps1lon for infrastructure

PR #386, opened 2019-10-18
test: add bubbling tests

pheeria:
**What**:
Add tests to check event propagation.

**Why**:
To have a better coverage. Currently, only firing events on elements directly is being tested.

**How**:
1. Create a `div` (being the most generic element)
1. Add an event listener
1. Create an inner `div`
1. Wrap an inner `div` inside of a `div`
1. Fire an event on inner `div`
1. Expect the event listener on the outer `div` to fire

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

I also fixed what I believe to be copy-paste error with the `message` event.

### DIFF
--- a/src/__tests__/events.js
+++ b/src/__tests__/events.js
@@ -333,7 +333,7 @@ test('fires shortcut events on Window', () => {
   window.addEventListener('click', clickSpy)
   fireEvent.click(window)
   expect(clickSpy).toHaveBeenCalledTimes(1)
-  window.removeEventListener('message', clickSpy)
+  window.removeEventListener('click', clickSpy)
 })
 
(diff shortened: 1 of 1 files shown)


src/__tests__/events.js
```diff
@@ -137,6 +137,90 @@ const eventTypes = [
   },
 ]
 
+const bubblingEvents = [
```
kentcdodds: I think it would be simpler if this were split into two arrays and they were arrays of strings only. e.g. `bubblingEvents` and `nonBubblingEvents`.

pheeria: This was actually how I originally wrote it :)
Done!

kentcdodds:
@all-contributors please add @pheeria for tests

Issue #390, opened 2019-10-21
Performance concern with jsdom and getByRole


oliviertassinari:
- `DOM Testing Library` version: 6.8.1
- `node` version: v12.6.0

I apologize in advance for not following the issue template; I'm not sure it serves performance discussion well.

We are in the progress of migrating the test suite of Material-UI from enzyme shallow + mount API to testing-library. However, we have hit a roadblock (we should consider what we do with this block, do we force our way in, move around it, etc?). While the test suite takes about 14s to run on v3, it now takes about 70s (on my machine). It's a x5 increase. Diving more into the issue, I could identify this problem:

![Capture d’écran 2019-10-22 à 01 18 17](https://user-images.githubusercontent.com/3165635/67250226-dc39d880-f46a-11e9-98ca-23a035e1c3a7.png)

Once we aggregate all the data, we realize that we indeed spend most of our time in this function:

![Capture d’écran 2019-10-22 à 01 26 21](https://user-images.githubusercontent.com/3165635/67250321-2b800900-f46b-11e9-9c02-442c5b5051e3.png)

So it seems that the numerous calls to `window.getComputedStyle()` are using +65% of the available CPU resources. If we run the same tests in a chrome headless, the performance are much better, about 20s instead of 70s. 

So my question is, do we have a viable workaround? In the case of Material-UI, we have migrated about 28% of our tests. If I extrapolate, once we hit 100%, our unit tests will take about 4 minutes (to be compared with ~14s in v3). Should we do something about it? If yes, what options do we have? Thank you!

kentcdodds:
Cc @eps1lon... This is pretty bad. I'm starting to think we should switch this from an opt out to an opt in...

eps1lon:
Honestly a 5x hit is somewhat ok for me considering it's an increase in confidence. I don't really buy in into perf for tests especially if we are still in the 1 minute ballpark. We should sync first if we prioritize fast or safe tests. I at least won't put much work into making tests fast. I outlined workarounds and I'm happy to assist with the work that can be done.

eps1lon:
Machine time is for me personally one of the least important concerns. So while I think this shouldn't matter I'm a consequentialist at the end of the day and if nobody tests a11y because of X then I still want to improve X.

I can't make any guarantees about the gains though. It'll always be slower than whatever your testing approach was before. If you are ok with low-quality<sup>1</sup>, inaccessible software then you should switch back to enzyme.

[...]

PR #392, opened 2019-10-24
feat: Include hidden elements by default

eps1lon:
**What**:

- default for `hidden` is now `true` meaning inaccessible roles are included by default
- make the default value configurable

**Why**:

- it was demanded

**How**:

- change default value (read from config)
- adjust test descriptions

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -132,7 +132,7 @@ There are no accessible roles. But there might be some inaccessible roles. If yo
 `)
 })
 
- test('can exclude excludes elements which have visibility hidden', () => {
+test('can exclude elements which have visibility hidden', () => {
   const {getByRole} = render('<div style="visibility: hidden;"><ul /></div>')
 
   expect(() => getByRole('list', {hidden: false}))
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -1,11 +1,11 @@
 import {render} from './helpers/test-utils'
 
-test('by default logs accessible roles when it fails', () => {
+test('by default logs available roles when it fails', () => {
   const {getByRole} = render(`<h1>Hi</h1>`)
   expect(() => getByRole('article')).toThrowErrorMatchingInlineSnapshot(`
-"Unable to find an accessible element with the role "article"
+"Unable to find an element with the role "article"
 
-Here are the accessible roles:
+Here are the available roles:
 
   heading:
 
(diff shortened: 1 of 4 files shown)


src/__tests__/role.js
```diff
@@ -117,7 +135,8 @@ There are no accessible roles. But there might be some inaccessible roles. If yo
 test('by default excludes elements which have visibility hidden', () => {
```
afontcu:
```suggestion
test('can exclude excludes elements which have visibility hidden', () => {
```

src/__tests__/role.js
```diff
@@ -21,20 +21,36 @@ Here are the accessible roles:
 `)
 })
 
-test('when hidden: true logs available roles when it fails', () => {
-  const {getByRole} = render(`<div hidden><h1>Hi</h1></div>`)
-  expect(() => getByRole('article', {hidden: true}))
+test('when hidde: false logs accessible roles when it fails', () => {
```
kentcdodds:
```suggestion
test('when hidden: false logs accessible roles when it fails', () => {
```

eps1lon:
> I think we should leave it at its current default

-- https://github.com/testing-library/dom-testing-library/issues/390#issuecomment-545532483

I want to make sure we're in sync here: This changes the default value. This basically restores the behavior to where it was before ByRole got introduced.

src/__tests__/role.js
```diff
@@ -114,10 +132,11 @@ There are no accessible roles. But there might be some inaccessible roles. If yo
 `)
 })
 
-test('by default excludes elements which have visibility hidden', () => {
+ test('can exclude excludes elements which have visibility hidden', () => {
```
afontcu: Looks like I messed up the suggestion. my bad! 🙌 

```suggestion
test('can exclude elements which have visibility hidden', () => {
```

kentcdodds:
I'm starting to think that it's ok to leave this enabled by default. My biggest challenge making this decision is that I no longer have a codebase with a bunch of tests to get a feel for what the real-world runtime of tests will be like. All I have to go off of is from folks like @oliviertassinari and others who are running thousands of tests.

I think that running 2000+ tests in less than a minute is pretty good personally, especially considering the improved confidence you're getting. I'd really love to hear from others though.

oliviertassinari:
@kentcdodds I think that @eps1lon has one of the best positions to provide a great answer this tradeoff. I have personally too limited context. I can only relate what I have experienced. My observations with Material-UI's have been: 1. performance slowness comes from style computation (`window.getComputedStyle`) with JSDOM. 2. style computation output doesn't impact the results of Material-UI tests (yet).

kentcdodds:
I think you're right.

@eps1lon, I defer to you. What do you suggest we do?

eps1lon:
I wasn't a fan of introducing the original change as non-breaking and would therefore not introduce another one (we already got two reports). I'd say we keep the old default and make it configurable so that folks who actually measure a non-tolerable performance hit can deactivate this feature. But I wouldn't advertise it too much. Usually we should stay below 100ms cost which is perfectly fine. 

For folks running there whole test suite while working I can only recommend not doing that since it penalizes additional tests which is not a situation you want to be in. jest has an awesome cli to help with these issues (mocha not so much) and both have the option to put (ir)relevant tests on a allow/disallow list.

kentcdodds:
Sounds great. Let me know when you have something for me to review/merge and we'll make it happen. 👍

eps1lon:
> Sounds great. Let me know when you have something for me to review/merge and we'll make it happen.

#398 extracts the configure feature from this PR.

kentcdodds:
Thanks @eps1lon 👍

PR #395, opened 2019-10-24
feat(ByRole): Improve perf if multiple elements are found

eps1lon:
**What**:

Improves perf >20% reduction if ByRole returns multiple elements. The more elements the bigger the reduction

**Why**:

Perf concerns were raised about `isInaccessible`

**How**:

An element being inaccessible does not imply every descendant is inaccessible because elements can re-appear by setting `visibility: visible` in a `visibility: hidden` subtree. However, they can't override `hidden` or `aria-hidden=true` or `display: none`. So we cache `isSubtreeInaccessible` return values for each element. For collections of elements (e.g. `option` or `listitem`) this reduces getComputedStyle calls from `O(n * depth(n))` to `O(n) + O(depth(n))` where n is the number of elements of the given role. It enables a very fast best case for collections where the containing element (e.g. the `listbox`) is hidden. We only have to compute the style once for subsequent `listitems` and can exit early once we hit the `listbox` and the cache `subtreeIsInaccessible` returns `true`.

To avoid passing the computed style too much around I removed `getComputedStyle().visibility` compatibility for jsdom < 15.2. Once jest 25 is released this will be fixed for jest users anyway and CSS was one of the lesser concerns of other collaborators. 

We use a `WeakMap` for the cache. Those are supported in IE 11: https://caniuse.com/#search=Weakmap

**Checklist**:

- [x] Tests
- [ ] Ready to be merged 
  - [ ] review
  - [x] testing canary in Material-UI https://github.com/mui-org/material-ui/pull/18021
  - [x] enchmark removal


kentcdodds:
> An element being inaccessible does not imply every descendant is inaccessible because elements can re-appear by setting visibility: visible in a visibility: hidden subtree.

Ah rats. What a shame! Haha. Thank you so much for working on this @eps1lon! I appreciate your help on this.

PR #406, opened 2019-11-27
fix: Various elements and their default roles

eps1lon:
DO NOT MERGE. new version of `aria-query` isn't released yet. Instead we're using a version built from `aria-query@master`.

Opening this already to get some comments on how we address the "with a accessible name" constraint.

**What**:

Bump `aria-query`

**Why**:

Fixes these bugs:
- `<input>` being considered a `textbox`
- `listbox` and `combobox` not being recognized as implicit roles on certain elements

**How**:

Built `aria-query` from their `master` branch via CodeSandbox CI.

**Checklist**:

- [x] Tests
- [ ] Ready to be merged


src/role-helpers.js
```diff
@@ -69,16 +69,54 @@ function getImplicitAriaRoles(currentNode) {
     }
   }
 
+  // <form /> with an accessible name?
+  if (currentNode.matches('form')) {
```
eps1lon: This is just restoring the old implementation that did not care about accessible name.

Technically `<form>` does not have the `role="form"` because the accessible name is missing (don't ask me why. I haven't tested it yet with AT; just following the spec). We could apply some heuristics to guess if it actually has one but if we want to be strict we would need to run accessible name computation (expensive).

Assuming I finish `byRole('form', { name: 'some-name' })` we could leave this quirk and then ignore any element returned `getImplicitAriaRoles` if the accessible name is empty (and potentially warn if a form is queried without providing a name).

kentcdodds:
What's the status here now that `aria-query` has been updated?

eps1lon:
Seems like I lost track of this as well. This PR is superseded by #446

PR #408, opened 2019-11-29
feat(byRole): Add `name` filter

eps1lon:
**What**:
Add a `name` option to `ByRole` queries. If a `name` is provided then the error messages will include the accessible name of each element with a role.

**Why**:

Better alternative to `byText` (especially for buttons) or `byLabelText` when `aria-label` or `aria-labelledby` is used (e.g. for custom Select implementations) or `byTestId` for "semantic" elements e.g. menuitem. 

There's a bit more to write in https://testing-library.com/docs/guide-which-query to explain why you still might want to use byLabelText('street') instead of `byRole('textbox', { name: 'street' })`.

**How**:

Using https://github.com/eps1lon/dom-accessibility-api#readme which implements https://www.w3.org/TR/accname-1.2/ and is tested against web-platform-tests

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs). See https://github.com/testing-library/testing-library-docs/pull/368 ([Preview](https://deploy-preview-368--testing-library.netlify.com/docs/dom-testing-library/api-queries#byrole))
- [x] Tests
- [x] Ready to be merged
      

I would hope that (after review and testing phase) we can move that package under the testing-library umbrella and move isInaccessible into dom-accessibility-api. In the end I'd like to add a single configuration that allows mocking `window.getComputedStyle` for the a11y context. That operation is quite expensive and I think allowing to pretend that every element is visible is sufficient for unit testing. The rationale being that we don't exclude certain groups of people by (accidentally) implementing inaccessible component but instead exclude everyone if we assert on elements that are actually visually hidden.

### DIFF
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -11,6 +11,7 @@ Here are the accessible roles:
 
   heading:
 
+  Name "Hi":
   <h1 />
 
   --------------------------------------------------
@@ -33,6 +34,7 @@ Here are the available roles:
 
   heading:
 
+  Name "Hi":
   <h1 />
 
   --------------------------------------------------
@@ -227,7 +229,7 @@ test('can be filtered by accessible name', () => {
 test('accessible name comparison is case sensitive', () => {
   const {getByRole} = render(`<h1>Sign <em>up</em></h1>`)
 
-  // actual:  "Sign up", 
+  // actual:  "Sign up",
   // queried: "Sign Up"
   expect(() => getByRole('heading', {name: 'Sign Up'}))
     .toThrowErrorMatchingInlineSnapshot(`
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -1,6 +1,6 @@
 import {configure, getConfig} from '../config'
-import {render, renderIntoDocument} from './helpers/test-utils'
 import {getQueriesForElement} from '../get-queries-for-element'
+import {render, renderIntoDocument} from './helpers/test-utils'
 
 test('by default logs accessible roles when it fails', () => {
   const {getByRole} = render(`<h1>Hi</h1>`)
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -224,28 +224,11 @@ test('can be filtered by accessible name', () => {
   ).not.toBeNull()
 })
 
-test('accessible name filter implements TextMatch', () => {
-  const {getByRole} = render(
-    `<h1>Sign <em>up</em></h1><h2>Details</h2><h2>Your Signature</h2>`,
-  )
-
-  // subset via regex
-  expect(getByRole('heading', {name: /gn u/})).not.toBeNull()
-  // regex
-  expect(getByRole('heading', {name: /^sign/i})).not.toBeNull()
-  // function
-  expect(
-    getByRole('heading', {
-      name: (name, element) => {
-        return element.nodeName === 'H2' && name === 'Your Signature'
-      },
-    }),
-  ).not.toBeNull()
-})
-
-test('includes accesible names in error message', () => {
+test('accessible name comparison is case sensitive', () => {
   const {getByRole} = render(`<h1>Sign <em>up</em></h1>`)
 
+  // actual:  "Sign up", 
+  // queried: "Sign Up"
   expect(() => getByRole('heading', {name: 'Sign Up'}))
     .toThrowErrorMatchingInlineSnapshot(`
 "Unable to find an accessible element with the role "heading" and name "Sign Up"
@@ -268,6 +251,29 @@ Here are the accessible roles:
   </h1>
 </div>"
 `)
+})
+
+test('accessible name filter implements TextMatch', () => {
+  const {getByRole} = render(
+    `<h1>Sign <em>up</em></h1><h2>Details</h2><h2>Your Signature</h2>`,
+  )
+
+  // subset via regex
+  expect(getByRole('heading', {name: /gn u/})).not.toBeNull()
+  // regex
+  expect(getByRole('heading', {name: /^sign/i})).not.toBeNull()
+  // function
+  expect(
+    getByRole('heading', {
+      name: (name, element) => {
+        return element.nodeName === 'H2' && name === 'Your Signature'
+      },
+    }),
+  ).not.toBeNull()
+})
+
+test('TextMatch serialization in error message', () => {
+  const {getByRole} = render(`<h1>Sign <em>up</em></h1>`)
 
   expect(() => getByRole('heading', {name: /Login/}))
     .toThrowErrorMatchingInlineSnapshot(`
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -1,5 +1,6 @@
 import {configure, getConfig} from '../config'
-import {render} from './helpers/test-utils'
+import {render, renderIntoDocument} from './helpers/test-utils'
+import {getQueriesForElement} from '../get-queries-for-element'
 
 test('by default logs accessible roles when it fails', () => {
   const {getByRole} = render(`<h1>Hi</h1>`)
@@ -181,6 +182,73 @@ test('can include inaccessible roles', () => {
   const {getByRole} = render('<div hidden><ul  /></div>')
 
   expect(getByRole('list', {hidden: true})).not.toBeNull()
+})
+
+test('can be filtered by accessible name', () => {
+  const {getByRole} = renderIntoDocument(
+    `
+<div>
+  <h1>Order</h1>
+  <h2>Delivery Adress</h2>
+  <form aria-label="Delivery Adress">
+    <label>
+      <div>Street</div>
+      <input type="text" />
+    </label>
+    <input type="submit" />
+  </form>
+  <h2>Invoice Adress</h2>
+  <form aria-label="Invoice Adress">
+    <label>
+      <div>Street</div>
+      <input type="text" />
+    </label>
+    <input type="submit" />
+  </form>
+</div>`,
+  )
+
+  const deliveryForm = getByRole('form', {name: 'Delivery Adress'})
+  expect(deliveryForm).not.toBeNull()
+
+  expect(
+    // TODO: upstream bug in `aria-query`; should be `button` role
+    getQueriesForElement(deliveryForm).getByRole('textbox', {name: 'Submit'}),
+  ).not.toBeNull()
+
+  const invoiceForm = getByRole('form', {name: 'Delivery Adress'})
+  expect(invoiceForm).not.toBeNull()
+
+  expect(
+    getQueriesForElement(invoiceForm).getByRole('textbox', {name: 'Street'}),
+  ).not.toBeNull()
+})
+
+test('includes accesible names in error message', () => {
+  const {getByRole} = render(`<h1>Sign <em>up</em></h1>`)
+
+  expect(() => getByRole('heading', {name: 'Sign Up'}))
+    .toThrowErrorMatchingInlineSnapshot(`
+"Unable to find an accessible element with the role "heading" and name "Sign Up"
+
+Here are the accessible roles:
+
+  heading:
+
+  Name "Sign up":
+  <h1 />
+
+  --------------------------------------------------
+
+<div>
+  <h1>
+    Sign 
+    <em>
+      up
+    </em>
+  </h1>
+</div>"
+`)
 })
 
 describe('configuration', () => {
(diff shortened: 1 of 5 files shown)


src/queries/role.js
```diff
@@ -56,6 +58,11 @@ function queryAllByRole(
           }) === false
         : true
     })
+    .filter(element => {
+      return typeof name === 'string'
+        ? computeAccessibleName(element) === name
```
kentcdodds: Could we make the `name` option support everything that `TextMatch` supports? Like function and regex?

kentcdodds:
Also, I completely agree with your comment about moving things around a bit 👍

src/queries/role.js
eps1lon: I'd prefer to keep the initial feature as simple as possible. I already have a hard time picturing use cases for `getByRole(/regex/)`. Just because we can do it doesn't mean we should. Especially since we have to teach a new option and all the overloads for it.

It's very likely a feature that will be requested and should be available at some point. Just not from the start and without a non-technical justification

One goal of this feature was that you don't need to use regular expressions to properly ignore certain whitespace (which is already handled by `computeAccessibleName`).

alexkrolick: I think he meant `getByRole('button', {name: /xyz/})`

alexkrolick:
One thing I am wondering is why this shouldn't support combining the other matchers as well:

```js
// validate through aria-labelledby or htmlFor
getByRole('input', {label: 'username'})

// validate through aria-labelledby or htmlFor
getByRole('heading', {label: 'username'})
```

Seems like a one off when we should be figuring out how to intersect queries in general. There have been a few proposals on how to do that.

src/queries/role.js
eps1lon: I know. I'd like to see actual use cases first though since we don't parse text with regular expressions. I'm not opposed to this feature per se. Just want to keep the PR as small as possible since I'd like to keep the amount of work I have to put into features I don't need as small as possible.

eps1lon:
> getByLabelText('uh oh', {role: 'dialog'})

Not sure what this is supposed to query. A dialog doesn't have a label. It's important to note that `getByLabelText('street')` is not equivalent to `getByRole('textbox', { name: 'street' })` because the `<label>` element has other behavior as well: Clicking a label will focus the control it is associated with.

> getByRole('textbox', {displayValue: 'username'})

I would have to look into this more. I don't know how display value is handled currently but we need consider that one might include the actual `value` attribute which is sent to the server/form library while e.g. `<div role="textbox">value</div>` does no such thing necessarily.

alexkrolick:
That was just an example, but this is from MDN:

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role

```html
<div role="dialog" aria-labelledby="dialog1Title" aria-describedby="dialog1Desc">
  <h2 id="dialog1Title">Your personal details were successfully updated</h2>
  <p id="dialog1Desc">You can change your details at any time in the user account section.</p>
  <button>Close</button>
</div>
```

eps1lon:
> That was just an example, but this is from MDN:

Sure. But be careful to not mix a `<label>` with the accessible name. They share some features but also have distinct features (e.g. clicking a `<label>` focuses the associated input). 

If you used `getByLabelText` for the accessible name only then you can switch. If your test implied the click-focus feature then you shouldn't (or if you do add another test for click-focus).

eps1lon:
Plan for this in the next week: 
- [ ] update branch
- [ ] write docs (most important part of this feature)
- [ ] unify approach for "loose mode" (where styles are ignored)
- [ ] another review pass for `dom-accessibility-api`
- [ ] Implement https://github.com/testing-library/dom-testing-library/pull/408#discussion_r352263350
- [ ] then this is ready to go from my perspective

kentcdodds:
Awesome! Let us know when you're ready for a review. Thanks a bunch!

src/queries/role.js
eps1lon: Implemented.

babramczyk:
I'm also curious about the inversion that @alexkrolick mentioned. Maybe the example he gave wasn't applicable (finding a dialog by label text), but others would be.

```javascript
getByText(/my website/i, { role: 'heading' })
```

This seems a little more expressive, at least for this use case than the query added in this PR:

```javascript
getByRole('heading', { name: /my website/i })
```

I want to get something with specific text, and the fact that it's a heading is supplementary to that. I concede that this is nitpicky, though.

eps1lon:
> I want to get something with specific text, and the fact that it's a heading is supplementary to that. I concede that this is nitpicky, though.

I wouldn't agree with that since roles might be button, dialog, heading etc. These are not secondary at all.

The point is to encourage writing out what kind of element you're dealing with. It's more important that the element is e.g. a button than that it reads "log in". The first one is part of the UI/UX. The second one is subject to localization.

babramczyk:
Fair enough. I honestly can see scenarios in which either the role or the text are the primary thing you're concerned with.

src/__tests__/role.js
```diff
@@ -183,6 +184,138 @@ test('can include inaccessible roles', () => {
...
+    <input type="submit" />
+  </form>
+</div>`,
+  )
+
+  const deliveryForm = getByRole('form', {name: 'Delivery Adress'})
+  expect(deliveryForm).not.toBeNull()
+
+  expect(
+    // TODO: upstream bug in `aria-query`; should be `button` role
```
kentcdodds: Should we contribute a fix to aria-query before merging this?

src/__tests__/role.js
```diff
@@ -183,6 +184,138 @@ test('can include inaccessible roles', () => {
...
+
+  expect(() => getByRole('heading', {name: 'Sign Up'}))
+    .toThrowErrorMatchingInlineSnapshot(`
+"Unable to find an accessible element with the role "heading" and name "Sign Up"
+
+Here are the accessible roles:
+
+  heading:
+
+  Name "Sign up":
```
kentcdodds: The only difference between this and the query is the `u` is not capitalized. Could we make this difference more stark so this makes more sense why it failed to find an element please? Thanks!

src/__tests__/role.js
```diff
@@ -183,6 +184,138 @@ test('can include inaccessible roles', () => {
...
+  <h1>
+    Sign 
+    <em>
+      up
+    </em>
+  </h1>
+</div>"
+`)
+
+  expect(() => getByRole('heading', {name: /Login/}))
```
kentcdodds: Not sure it's necessary to have all three of these because they're all testing the same codepath and their differences are covered in other tests.

src/queries/role.js
```diff
@@ -70,6 +72,21 @@ function queryAllByRole(
...
+      }
+
+      const accessibleName = computeAccessibleName(element)
+      if (typeof name === 'string') {
+        return name === accessibleName
+      }
+      if (typeof name === 'function') {
+        return name(accessibleName, element)
+      }
+      return name.test(accessibleName)
```
kentcdodds: Why don't we not assume it's a regex and check whether it is a regex. If it's not then we can throw a more helpful error message.

Honestly, we should probably have some code already that handles TextMatch options and *that* code could do this. Don't we already have some TextMatch code?

src/role-helpers.js
```diff
@@ -138,23 +139,30 @@ function getRoles(container, {hidden = false} = {}) {
...
         .join('\n\n')
 
       return `${role}:\n\n${elementsString}\n\n${delimiterBar}`
     })
     .join('\n')
 }
 
-const logRoles = (dom, {hidden = false} = {}) =>
-  console.log(prettyRoles(dom, {hidden}))
+const logRoles = (dom, {hidden = false, includeName = false} = {}) =>
```
kentcdodds: Why would we not want to always include the name?

src/__tests__/role.js
eps1lon: Right. They released a new major. Opening a separate PR that bumps aria-query and then we'll see if that worked.

src/__tests__/role.js
eps1lon: It's more about showcasing how the `name` argument is serialized. It may be too verbose in which case the codepath should change. And overall: Isn't that testing implementation details :wink:

src/queries/role.js
eps1lon: That is how TextMatch is treated in other instances as well:
- https://github.com/testing-library/dom-testing-library/blob/a03f05671f849802c2d890e6de4532fb6b52f28c/src/matches.js#L11-L12
- https://github.com/testing-library/dom-testing-library/blob/a03f05671f849802c2d890e6de4532fb6b52f28c/src/matches.js#L26-L27
 
We could add a runtime TypeError though nowadays I would rather let types do the heavy lifting.

src/role-helpers.js
eps1lon: Haven't thought much about this. Just wanted to not change existing behavior. 
Since this is only used by library consumers it might be a good idea to include the name by default to promote that we now support accessible name queries?

src/__tests__/role.js
eps1lon: I had trouble understanding it as well. I would like to keep the test to make sure that `{ name: string }` is always case-sensitive. I'll add a comment to the test explaining the issue.

eps1lon:
Addressed https://github.com/testing-library/dom-testing-library/pull/408#discussion_r374779780 and https://github.com/testing-library/dom-testing-library/pull/408#discussion_r374780437 in 3c7dd0e54bdb1ec615354f93dedd098640b89531.

src/queries/role.js
eps1lon: You're right we can reuse existing logic: 

```patch
diff --git a/src/queries/role.js b/src/queries/role.js
index 3d93a48..a442bfd 100644
--- a/src/queries/role.js
+++ b/src/queries/role.js
@@ -78,14 +78,12 @@ function queryAllByRole(
         return true
       }
 
-      const accessibleName = computeAccessibleName(element)
-      if (typeof name === 'string') {
-        return name === accessibleName
-      }
-      if (typeof name === 'function') {
-        return name(accessibleName, element)
-      }
-      return name.test(accessibleName)
+      return matches(
+        computeAccessibleName(element),
+        element,
+        name,
+        text => text,
+      )
     })
```

-- a542ebcf544663bed58b13d48c6a0af0d4fa9d5f

src/role-helpers.js
alexkrolick: Yes, since it's debug output, it seems fine to change.

src/__tests__/role.js
kentcdodds:
```suggestion
  expect(() => getByRole('heading', {name: /something that does not match/}))
```

I think this makes it more clear that it's not just a typo...

src/role-helpers.js
```diff
@@ -138,23 +139,30 @@ function getRoles(container, {hidden = false} = {}) {
...
       const elementsString = elements
-        .map(el => prettyDOM(el.cloneNode(false)))
+        .map(
+          el =>
+            `${
+              includeName === true
+                ? `Name "${computeAccessibleName(el)}":\n`
+                : ''
+            }${prettyDOM(el.cloneNode(false))}`,
+        )
```
kentcdodds: Prettier makes this look funny. Why don't we do this:

```suggestion
        .map(el =>
          [
            includeName === true
              ? `Name "${computeAccessibleName(el)}":\n`
              : '',
            prettyDOM(el.cloneNode(false)),
          ].join(''),
        )
```

kentcdodds: Or maybe it would be better to just do this:

```suggestion
        .map(el => {
          const nameString =
            includeName === true ? `Name "${computeAccessibleName(el)}":\n` : ''
          const domString = prettyDOM(el.cloneNode(false))
          return `${nameString}${domString}`
        })
```

src/role-helpers.js
```diff
@@ -138,23 +139,30 @@ function getRoles(container, {hidden = false} = {}) {
...
         .join('\n\n')
 
       return `${role}:\n\n${elementsString}\n\n${delimiterBar}`
     })
     .join('\n')
 }
 
-const logRoles = (dom, {hidden = false} = {}) =>
-  console.log(prettyRoles(dom, {hidden}))
+const logRoles = (dom, {hidden = false, includeName = true} = {}) =>
```
kentcdodds: Why do we provide the option to not include the name?

kentcdodds:
I'm working on this now. I'll go ahead and remove the `includeName` feature because I think it's easier to add it later if we really need it than to include it if we don't really have a use case. Thanks!

eps1lon:
Agreed with everything. Thanks for finishing this up!

PR #415, opened 2019-12-13
feat(wait): wait will now also run your callback on DOM changes

### LINKED ISSUE #376: BREAKING: make waitForDomChange accept a callback
Just an idea. When I use `wait`, I give it an assertion to wait for something to pass and I know my callback is called every 50ms until it passes. The nice thing about this is that we don't care what goes on during that time. We just wait for that assertion to pass.

However with `waitForDomChange`, we're just waiting for a single update to the DOM, but that's less helpful and arguably a bit more like implementation details because we shouldn't care how many DOM changes there are, we should wait until the DOM changes to a desirable state.

I'd like to think of `waitForDomChange` as a variant to `wait` which re-runs a callback whenever the DOM changes rather than arbitrarily every 50ms.

So it'd look like this (the same as `waitForElement`):

```typescript
function waitForDomChange<T>(
  callback: () => T,
  options?: {
    container?: HTMLElement
    timeout?: number
    mutationObserverOptions?: MutationObserverInit
  }
): Promise<T>
```

I expect that it would work similar to `wait` if you provide no callback. So this would be a breaking change only for people who are passing `options` (they would have to pass a callback of `undefined` ahead of the options, or even better, an actual callback making an actual useful assertion).

So this change would not break existing code that does this:

```javascript
await waitForDomChange()
```

But it *would* break existing code that does this:

```javascript
await waitForDomChange({container: el})
```

And that would need to change to:

```javascript
await waitForDomChange(undefined, {container: el})
```

For the exact same behavior as before, or, preferably someone would give a callback with an assertion:

```javascript
await waitForDomChange(() => expect(getByText(/status/i).toHaveTextContent('finished'), {container: el})
```

I think this is a good time to make this change considering we're about to release another breaking change: #373

kentcdodds:
**What**: Make `wait` use a MutationObserver in addition to the previous behavior

**Why**: This way we can get rid of `waitForDomChange` (in a future release) *and* we drastically simplify the other async utilities. `wait` also gets a super-power of being able to re-run the callback faster than the interval when the DOM is changed, so this should speed up people's tests who are using `wait` today. Closes #376

**How**: Remove `wait-for-expect` (cc @lgandecki) because it wont work nicely with this approach. I also refactored the other utilities to build on top of `wait` (which is something I've wanted to do for a long time now).

**Checklist**:

- [ ] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs)
- [ ] ~I've prepared a PR for types targeting [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__dom)~ (this should not be necessary. No APIs have changed)
- [x] Tests
- [x] Ready to be merged
- [x] Set lower timeout default

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -19,9 +19,11 @@ let config = {
 
   // called when getBy* queries fail. (message, container) => Error
   getElementError(message, container) {
-    return new Error(
+    const error = new Error(
       [message, prettyDOM(container)].filter(Boolean).join('\n\n'),
     )
+    error.name = 'TestingLibraryElementError'
+    return error
   },
 }
 
--- a/src/config.js
+++ b/src/config.js
@@ -5,7 +5,7 @@ import {prettyDOM} from './pretty-dom'
 // './queries' are query functions.
 let config = {
   testIdAttribute: 'data-testid',
-  asyncUtilTimeout: 4500,
+  asyncUtilTimeout: 1000,
   // this is to support React's async `act` function.
   // forcing react-testing-library to wrap all async functions would've been
   // a total nightmare (consider wrapping every findBy* query and then also
--- a/src/wait-for-dom-change.js
+++ b/src/wait-for-dom-change.js
@@ -8,6 +8,9 @@ import {
 } from './helpers'
 import {getConfig} from './config'
 
+// deprecated... TODO: remove this method. People should use wait instead
+// the reasoning is that waiting for just any DOM change is an implementation
+// detail. People should be waiting for a specific thing to change.
 function waitForDomChange({
   container = getDocument(),
   timeout = getConfig().asyncUtilTimeout,
--- a/src/wait-for-dom-change.js
+++ b/src/wait-for-dom-change.js
@@ -8,6 +8,8 @@ import {
 } from './helpers'
 import {getConfig} from './config'
 
+let hasWarned = false
+
 // deprecated... TODO: remove this method. People should use wait instead
 // the reasoning is that waiting for just any DOM change is an implementation
 // detail. People should be waiting for a specific thing to change.
@@ -21,6 +23,13 @@ function waitForDomChange({
     characterData: true,
   },
 } = {}) {
+  if (!hasWarned) {
+    hasWarned = true
+    // eslint-disable-next-line no-console
+    console.warn(
+      `\`waitForDomChange\` has been deprecated. Use \`wait\` instead: https://testing-library.com/docs/dom-testing-library/api-async#wait.`,
+    )
+  }
   return new Promise((resolve, reject) => {
     const timer = setTimeout(onTimeout, timeout)
     const observer = newMutationObserver(onMutation)
--- a/src/wait-for-dom-change.js
+++ b/src/wait-for-dom-change.js
@@ -25,7 +25,6 @@ function waitForDomChange({
 } = {}) {
   if (!hasWarned) {
     hasWarned = true
-    // eslint-disable-next-line no-console
     console.warn(
       `\`waitForDomChange\` has been deprecated. Use \`wait\` instead: https://testing-library.com/docs/dom-testing-library/api-async#wait.`,
     )
--- a/src/wait-for-element-to-be-removed.js
+++ b/src/wait-for-element-to-be-removed.js
@@ -1,85 +1,48 @@
-import {
-  getDocument,
-  newMutationObserver,
-  setImmediate,
-  setTimeout,
-  clearTimeout,
-  runWithRealTimers,
-} from './helpers'
-import {getConfig} from './config'
+import {wait} from './wait'
 
-function waitForElementToBeRemoved(
-  callback,
-  {
-    container = getDocument(),
-    timeout = getConfig().asyncUtilTimeout,
-    mutationObserverOptions = {
-      subtree: true,
-      childList: true,
-      attributes: true,
-      characterData: true,
-    },
-  } = {},
-) {
-  return new Promise((resolve, reject) => {
-    if (typeof callback !== 'function') {
-      reject(
-        new Error(
-          'waitForElementToBeRemoved requires a function as the first parameter',
-        ),
-      )
-    }
-    const timer = setTimeout(onTimeout, timeout)
-    const observer = newMutationObserver(onMutation)
+const isRemoved = result => !result || (Array.isArray(result) && !result.length)
 
-    // Check if the element is not present synchronously,
-    // As the name waitForElementToBeRemoved should check `present` --> `removed`
+async function waitForElementToBeRemoved(callback, options) {
+  if (!callback) {
+    return Promise.reject(
+      new Error(
+        'waitForElementToBeRemoved requires a callback as the first parameter',
+      ),
+    )
+  }
+
+  // Check if the element is not present synchronously,
+  // As the name implies, waitForElementToBeRemoved should check `present` --> `removed`
+  if (isRemoved(callback())) {
+    throw new Error(
+      'The callback function which was passed did not return an element or non-empty array of elements. waitForElementToBeRemoved requires that the element(s) exist before waiting for removal.',
+    )
+  }
+
+  return wait(() => {
+    let result
     try {
-      const result = callback()
-      if (!result || (Array.isArray(result) && !result.length)) {
-        onDone(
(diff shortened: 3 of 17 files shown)


src/wait-for-element.js
```diff
@@ -1,71 +1,21 @@
-import {
```
kentcdodds: It's easier to just look at the post-change version of this file: https://github.com/testing-library/dom-testing-library/blob/pr/wait-improvements/src/wait-for-element.js

src/wait-for-element-to-be-removed.js
```diff
@@ -1,85 +1,48 @@
-import {
```
kentcdodds: It's easier to just look at the post-change version of this file: https://github.com/testing-library/dom-testing-library/blob/pr/wait-improvements/src/wait-for-element-to-be-removed.js

kentcdodds:
That's weird. I think the codesandbox thing is a fluke.

mustafawm:
Hey Kent,

I was gonna post a suggestion/PR, but since this one is about `wait`, I will just leave it here:

Can we expose `wait` under a second alias: `until` or `till`?  I think the code would read _nicer:_
```js
  await till(() => expect(MockRedirect).toHaveBeenCalledWith({to: '/'}, {}));
```

_I know we can alias our imports.. just thought the API would look nicer.. though it might not be worth introducing new API changes (at this time)_

kentcdodds:
Hi @mustafawm, thanks for the feedback. I do prefer to avoid aliases when possible. We have an alias for `getQueriesForElement` (it's `within`), but I'm not a super-fan of it. The reason I made the alias was because I thought the function name was kinda long, but I didn't want to make a breaking change to rename it.

I'm not sure there's enough justification to alias this function, but you certainly could do this in [your own utils module](https://testing-library.com/docs/react-testing-library/setup#configuring-jest-with-test-utils).

Good luck!

mustafawm:
Hey Ken, thanks for your quick reply.

Agree. tbh, I wasn't feeling strongly about it .. just thought I put it out there and see :)  

Thanks for the utils module suggestion 🙏

src/wait-for-element-to-be-removed.js
```diff
@@ -1,85 +1,48 @@
...
+        'waitForElementToBeRemoved requires a callback as the first parameter',
+      ),
+    )
+  }
+
+  // Check if the element is not present synchronously,
+  // As the name implies, waitForElementToBeRemoved should check `present` --> `removed`
+  if (isRemoved(callback())) {
+    throw new Error(
+      'The callback function which was passed did not return an element or non-empty array of elements. waitForElementToBeRemoved requires that the element(s) exist before waiting for removal.',
```
afontcu: nitpicking ^^ sry about that

```suggestion
      'The callback function which was passed did not return an element or non-empty array of elements. waitForElementToBeRemoved requires that the element(s) exist(s) before waiting for removal.',
```

lgandecki:
Hey Kent, 

Thanks for pinging me here and letting me know - it will be bittersweet to see the installs of wait-for-expect plummet but it's for a good reason - the tiny helper is usable on it's own outside of dom-testing scope so we shouldn't be adding more complexity to it, and the dom-testing could use some improvements.. :)

I also like the idea of simplifying the API. To be honest in a lot of projects I've seen and helped with, people didn't even bother to use waitForDomChange, wait "just worked". Optimizations (savings of maybe a few dozens of ms per test) were not worth the extra pause (and learning) to decide which of the wait* tools the programmer should use in a given situation.

I'm glad that the initial design we've came up with lives on. It's so intuitive and clean. My only regret is the default long timeout - I was basing it off what I knew from browser-based testing (selenium, chimp, cypress, etc). But in more of a low-level environment the timeout hardly ever needs to go that high, with the downside that once things break the tests take forever to fail. (In browser-based testing the test would maybe take 8 seconds to fail instead of 4 seconds to pass, but in integration we can jump from 100 ms to 4500 ms which is much more annoying). It brakes the flow. It might be worth pointing this out more explicitly in the docs, that people might want to consider to change the default, if they run lightweight tests. But, at the same time, this family of libraries is all based on "smart defaults" :) 

Also - if you are going to do some breaking changes, this might be something to rethink, maybe?

src/wait.js
```diff
@@ -1,8 +1,79 @@
...
+  return new Promise((resolve, reject) => {
+    let lastError, lastTimer
+    const overallTimeoutTimer = setTimeout(onTimeout, timeout)
+
+    const observer = newMutationObserver(checkCallback)
+    runWithRealTimers(() =>
+      observer.observe(container, mutationObserverOptions),
+    )
+
+    function onDone(error, result) {
```
afontcu: Looks like `onDone` is expecting 2 params, but we will never send (and use) them both at the same time.

```js
function onDone(result, cb) {
  clearTimeout(overallTimeoutTimer)
  clearTimeout(lastTimer)
  setImmediate(() => observer.disconnect()) 
  cb(result)  
}
```

...and then: 

```js
onDone(callback(), resolve)
```

```js
onDone(lastError || new Error('Timed out in wait.'), reject)
```

What if we passed the callback action as a parameter, too, so we remove simplify both its public API and its implementation? Not a big deal, just something that crossed my mind.

_disclaimer:_ I'm not using my computer right now so I couldn't give it a go.

kentcdodds: Sorry, I'm not sure I understand what you're suggesting.

afontcu: If I got it right, `onDone` receives `error` and `result`, and depending on the received param (it's either one or the other, never the two of them) it calls `reject` or `resolve`.

What I was suggesting is that `onDone` received the appropiate payload (either `error` or `result`) and the appropiate callback function (either `reject` or `resolve`).

It's a small refactor, so not a big deal :)

kentcdodds: Oooohhhh, I understand. Sorry, I was just confused. Yes. I think that's a good idea 👍

eps1lon: I think the point is to be consistent with how node callbacks work ("error-first callbacks"). Then `promisify` "just" works. Otherwise the first argument becomes polymorphic which I'd like to avoid as much as possible.

kentcdodds:
Maybe we should change the default timeout as part of this change as well. I think it makes more sense for it to be ~1 second. If someone's test is taking longer than that for a single `wait` then they probably aren't mocking something they should be (in the context of Cypress and other E2E situations, we could change the default for those environments).

What do folks think about that?

afontcu:
+1 at setting a lower default value for the timeout – I've never found a case where a 4.5s timeout would make a (well-written and properly mocked) test pass while a lower one would make the same test fail.

I'm ok with 1s.

kentcdodds:
Rebased master and added two commits. If folks could take a look at the last two commits that would be super duper. Thanks!

src/config.js
```diff
@@ -3,7 +3,7 @@
 // './queries' are query functions.
 let config = {
   testIdAttribute: 'data-testid',
-  asyncUtilTimeout: 4500,
+  asyncUtilTimeout: 1000,
```
kentcdodds: Just realized I amended the first of those two changes. So this is all that first change was.

src/wait-for-dom-change.js
```diff
@@ -8,6 +8,9 @@ import {
 } from './helpers'
 import {getConfig} from './config'
 
+// deprecated... TODO: remove this method. People should use wait instead
+// the reasoning is that waiting for just any DOM change is an implementation
+// detail. People should be waiting for a specific thing to change.
```
afontcu: If we plan to remove this method in the next major release, should be add a warning so people know they are relying on a soon-to-be-gone method? Is it too invasive?

kentcdodds: I'm kinda thinking that we should add the warning in the major version bump, and then remove it in the next one.

kentcdodds:
I'm starting to think that instead of deprecating we should just remove it entirely what do people think?

To be clear, I'm talking about `waitForDomChange` and `waitForElement`.

`waitForElement` can be done really easily with `wait` and `waitForDomChange` is testing implementation details (and can be done really easily with `wait` as well.

Thoughts?

kentcdodds:
Changed my mind. We'll just have a deprecation warning for this release. People will be less mad at us 😅

Issue #417, opened 2019-12-13
Screen Export: Should it include the debug method?


jkdowdle:
### Describe the feature you'd like:

I was just trying out the screen object to power the queries of my tests and the first thing I did was try using the debug method, expecting it to be there like from the object you get back from render in 
'@testing-library/react'.

```
import React from 'react'
import { screen } from '@testing-library/react'

test('works', () => {
    setup()

    screen.debug()
})
```

And I noticed that it was not available. I would love to submit a PR to add it, but before I went looking around on how to do that I thought I would double check and make sure that that was not a conscious decision to leave out. Maybe some constraints on the difficulty of adding it or something like that. 

### Describe alternatives you've considered:
 We can continue getting the debug method off the returned object from rendering. 

Or, as was mentioned bellow, use the logDOM export. 

```
import {logDOM} from '@testing-library/react'

logDOM() // logs out document.body
```

### Teachability, Documentation, Adoption, Migration Strategy:
This would increase the surface area by a little bit, but since screen was already added, it would not be that much more to add. I think users, like myself, would expect it to be there. 

I want to thank all the maintainers, keep up the great work!

kentcdodds:
Hi @jkdowdle,

It's already supported:

```js
import {logDOM} from '@testing-library/react'

logDOM() // logs out document.body
```

Though I can see the value of having it on `screen` so you can call/remove it without having to update imports.

Thoughts from anyone else?

balazsorban44:
I did not think of it before, but now I am sure I would unconsciously go and try to get `debug` from `screen` every time! :sweat_smile:

kentcdodds:
I'm not opposed to adding it, but I'd like to hear arguments for/against.

alexkrolick:
Enzyme has a debug method for nodes, might make it easier to migrate nog screen did too

benmonro:
I vote Yes as this would make it a lot easier to report on debug in the testing lint rule.  Actually for that matter why not Make it a top level export?  I'd even support a breaking change to remove it from the results as that makes it quite tricky to lint when used with Custom render functions.

For note, the goal of the no debug lint rule is to keep it out of CI, and only use it locally temporarily when Debugging

jkdowdle:
@benmonro That sounds like the best argument so far for only allowing the top level export for the debug method. I personally haven't used the eslint rules yet for @testing-library but I'd like to start. Thank you

benmonro:
[...]

PR #429, opened 2020-01-21
Add a debug method to screen.

jkdowdle:
**What**:
A debug method is being added to the screen object.

**Why**:
Facilitate debugging tests without having to always modify imports.

**How**:
Added the debug method to the screen object, following the pattern for the debug method from render in react-testing-library.

**Checklist**:

- [x] Tests
- [ ] Ready to be merged
      

**Would close**
 https://github.com/testing-library/dom-testing-library/issues/417

**Docs PR**: 
https://github.com/testing-library/testing-library-docs/pull/362

### DIFF
--- a/src/__tests__/screen.js
+++ b/src/__tests__/screen.js
@@ -20,7 +20,6 @@ test('exposes debug method', () => {
   renderIntoDocument(
     `<button>test</button><span>multi-test</span><div>multi-test</div>`,
   )
-  expect(screen.debug).toBeInstanceOf(Function)
   // log document
   screen.debug()
   expect(console.log).toHaveBeenCalledTimes(1)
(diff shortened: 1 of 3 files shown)


jkdowdle:
I am not familiar with adding types, could I get some direction on how to do that? Thank you, 

I see something like 

```
export type Screen<Q extends Queries = typeof queries> = BoundFunctions<Q>;
```

What could I do to add the debug type?

src/__tests__/screen.js
```diff
@@ -1,9 +1,63 @@
...
   screen.getByText(/hello world/i)
   await screen.findByText(/hello world/i)
   expect(screen.queryByText(/hello world/i)).not.toBeNull()
 })
+
+test('exposes debug method', () => {
+  renderIntoDocument(
+    `<button>test</button><span>multi-test</span><div>multi-test</div>`,
+  )
+  expect(screen.debug).toBeInstanceOf(Function)
```
kentcdodds: I don't think we don't need this assertion because we make it implicitly by calling it. Implicit assertions aren't always the best, but when it's something as simple as what type something is I don't think it's all that useful to have.

src/__tests__/screen.js
```diff
@@ -1,9 +1,63 @@
...
+      </button>
+      <span>
+        multi-test
+      </span>
+      <div>
+        multi-test
+      </div>
+    </body>"
+  `)
+  // log single element
```
kentcdodds: Let's add a `console.log.mockClear()` between these so that each group of assertions can be more independent from the others.

src/get-queries-for-element.js
```diff
@@ -7,14 +7,19 @@ import * as defaultQueries from './queries'
 /**
  * @param {HTMLElement} element container
  * @param {FuncMap} queries object of functions
+ * @param {Object} initialValue for reducer
```
kentcdodds: Interesting idea. This probably isn't how I would have done it, but I'm actually pretty cool with it.

src/__tests__/screen.js
jkdowdle: That does make it a-lot more clear. Cool, I appreciate this suggestion.

jkdowdle:
@kentcdodds suggestions for the tests were added. Thanks!

jkdowdle:
> Love it. Thank you!

Thank you! I appreciate it!

PR #446, opened 2020-02-04
fix: Various implicit roles in ByRole

eps1lon:
With #408 we'll be able to correctly identify some of these implicit roles (e.g. `<form>` or `<section>`)

**What**:

Fix various issues with implicit roles 

**Why**:

Examples for elements with the wrong implicit role when using `aria-query@3`:
- `<form>` without an accessible name having a role (though even with it now has is not queried by the `form` role. this requires #408)
- `<select>` being identified as both `combobox` and `listbox` even though it is only a `combobox`
- `<input list="" />` not being identified as a combobox

**How**:

Bump `aria-query` to `^4.0.1`

**Checklist**:

- ~[ ] I've prepared a PR for types targeting~
- [x] Tests
- [ ] Ready to be merged


src/__tests__/element-queries.js
```diff
@@ -422,21 +427,27 @@ test('queryAllByRole returns semantic html elements', () => {
...
   expect(queryAllByRole(/grid/i)).toHaveLength(1)
-  expect(queryAllByRole(/form/i)).toHaveLength(1)
+  expect(queryAllByRole(/form/i)).toHaveLength(0)
   expect(queryAllByRole(/button/i)).toHaveLength(1)
   expect(queryAllByRole(/heading/i)).toHaveLength(6)
   expect(queryAllByRole('list')).toHaveLength(2)
   expect(queryAllByRole(/listitem/i)).toHaveLength(3)
-  expect(queryAllByRole(/textbox/i)).toHaveLength(2)
+  // TODO: with https://github.com/A11yance/aria-query/pull/42
+  // the actual value will match `toHaveLength(2)`
```
eps1lon: Can be verified by using https://pkg.csb.dev/eps1lon/aria-query/commit/44b2c4e3/aria-query as the `aria-query` version.

src/__tests__/role-helpers.js
```diff
@@ -149,18 +146,18 @@ test('getRoles returns expected roles for various dom nodes', () => {
...
 
 test('getImplicitAriaRoles returns expected roles for various dom nodes', () => {
-  const {section, h1, form, radio, input} = setup()
+  const {unnamedSection, h1, unnamedForm, radio, input} = setup()
 
-  expect(getImplicitAriaRoles(section)).toEqual(['region'])
+  expect(getImplicitAriaRoles(unnamedSection)).toEqual([])
   expect(getImplicitAriaRoles(h1)).toEqual(['heading'])
-  expect(getImplicitAriaRoles(form)).toEqual(['form'])
+  expect(getImplicitAriaRoles(unnamedForm)).toEqual([])
```
kentcdodds: Should we add a named form so we can make sure we capture the named one? Same for the section?

kentcdodds:
> this requires #408

Does this mean that we should make sure we merge #408 before this PR?

eps1lon:
Would be better to not have a this released without #408. Then we can add tests with named elements.

kentcdodds:
Working on getting this mergable right now.

src/__tests__/__snapshots__/role-helpers.js.snap
```diff
@@ -1,15 +1,7 @@
...
 
 exports[`logRoles calls console.log with output from prettyRoles 1`] = `
-"region:
-
-Name "":
-<section
-  data-testid="a-section"
-/>
-
---------------------------------------------------
```
kentcdodds: This change appears to be a result of the aria-query update. Maybe that's a bug?

nixxquality:
This was a breaking change in my case.

If anyone else is having issues with ****ByRole("form") not working anymore, add an ARIA label to your form.

```diff
   return (
-    <form onSubmit={ onSubmit }>
+    <form onSubmit={ onSubmit } aria-label="Configuration">
       <div id="floatsubmit">
```

smeijer:
Is that form bug something for which an issue should be created?

kentcdodds:
I don't think an aria label should be required on the form. This seems like a bug.

nixxquality:
It was an upstream change in aria-query, and it follows the [w3c spec](https://www.w3.org/TR/html-aria/#document-conformance-requirements-for-use-of-aria-attributes-in-html#el-form).
The only problem I had with it was that it was a breaking change in aria-change (4.0) but it was a minor change in this project, sending me on a bit of a journey to find out why my code suddenly wasn't working (because it wasn't correct in the first place, mind you).

kentcdodds:
Oh, that's very interesting. So a label is required on a form according to the spec?

nickmccurdy:
Should we bump the major version? The semver spec recommends doing that retroactively after it’s discovered that a breaking change has been already published accidentally.

smeijer:
> Oh, that's very interesting. So a label is required on a form according to the spec?

It does look that way:

> `role=form` if the form element has an `accessible name`. Otherwise, no corresponding role.
--

It still feels very weird to me. Other elements (like `heading`, `article`) get an implicit role, but `form` does not? Both Chrome and Firefox show me a `role=form` for forms without accessible names. Are they messing up here?

On a positive note, the label seems to work nicely. 

And maybe that's the entire idea behind it. A heading and article will make sense due to their content. But selecting a `form` without a `label`, might not. Think of `invoice address` and `shipping address`. Without proper labels, it's impossible to tell which "input: address" has the focus.

```html
<form aria-labelledby="signup-form">
  <h1 id="signup-form">
    Signup Form
  </h1>
  
  <input placeholder="email" />
</form>
```

```js
screen.getByRole('form', { name: /signup form/i })
```

[playground](https://testing-playground.com/gist/b2b117e502be76f223b69e6f57fe56de/ea02a702b3ea9aa954efde75f00c9a6618af4699)

PR #447, opened 2020-02-04
chore: Bump jsdom to 15

eps1lon:
Not sure which version we should use. While the latest is 16 it isn't the default version used in jest. Ideally we would test with jsdom 15 and 16 just to be aware of changes between those versions.

**What**:

Bump used jsdom version to 15.2.1

**Why**:

jsdom 15 has MutationObserver support as well as perf improvements. 

**How**:

- fixing breaking changes in jest and jsodm

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/wait-for-dom-change.js
+++ b/src/__tests__/wait-for-dom-change.js
@@ -18,22 +18,10 @@ test('waits for the dom to change in the document', async () => {
   setTimeout(() => container.firstChild.setAttribute('id', 'foo'))
   const mutationResult = await promise
   expect(mutationResult).toMatchInlineSnapshot(`
-Array [
-  Object {
-    "addedNodes": Array [],
-    "attributeName": "id",
-    "attributeNamespace": null,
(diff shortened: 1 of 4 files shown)


src/__tests__/wait-for-dom-change.js
```diff
@@ -73,23 +49,7 @@ describe('timers', () => {
...
-      "previousSibling": null,
-      "removedNodes": Array [],
-      "target": <div
-        id="foo"
-      />,
-      "type": "attributes",
-    },
-  ]
-  `)
+    await expect(promise).resolves.toMatchSnapshot()
```
eps1lon: Fixes 

> Jest: Multiple inline snapshots for the same call are not supported.

eps1lon:
@kentcdodds Will do. I didn't look closely what the purpose of the test was.

Should we leverage custom snapshot serializers or serialize them directly in the test?

PR #451, opened 2020-02-12
Fixed ByLabel queries to work with selects

costaraphael:
**What**:

Updating the ByLabel queries to work with `<select>` components. It didn't work properly with `<select>` components because the text from the options is part of the label's `textContent` property.

**Why**:

It seems super counter-intuitive that it works really well for other inputs (even `<textarea>`), but not for `<select>`. I think having this working properly will make this library an even greater joy to use :heart:

**How**:

I used the same strategy used for `<textarea>` components, but instead removing the `textContent` value from the `<select>` components from the label's `textContent`. 

**Checklist**:

- [ ] Documentation added to the (_I'm not sure it's needed in this case_)
- [ ] I've prepared a PR for types targeting (_I'm not sure either_)
- [ x] Tests
- [ x] Ready to be merged
      

If anyone is having this same issue and this fix is not available, you can use this workaround:

```js
getByLabelText(/^Label/)
// instead of
getByLabelText('Label')
```

I would also like to add that you rule for making front-end testing such a pleasant experience! :heart:

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -830,3 +830,16 @@ test('can get a textarea with children', () => {
   `)
   getByLabelText('Label')
 })
+
+test('can get a select with options', () => {
+  const {getByLabelText} = renderIntoDocument(`
+    <label>
+      Label
+      <select>
+        <option>Some</option>
+        <option>Options</option>
+      </select>
+    </label>
+  `)
+  getByLabelText('Label')
+})
--- a/src/queries/label-text.js
+++ b/src/queries/label-text.js
@@ -25,6 +25,12 @@ function queryAllLabelsByText(
       textToMatch = textToMatch.replace(textarea.value, '')
     })
 
+    // The children of a select are also part of `textContent`, so we
+    // need also to remove their text.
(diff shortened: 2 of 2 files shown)


costaraphael:
It would be much better IMO.

I'll have a look at this and come back with my findings :blush:

costaraphael:
@kentcdodds I tested this and it seems to work fine:

```js
const textToMatch = Array.from(label.childNodes)
  .filter(n => n.nodeName === "#text")
  .map(n => n.wholeText)
  .join('')
```

There is a hiccup with this approach, though. It limits the usage of inline elements within the label text (like `span`, `b`, or `i`), which would be a breaking change from the current behaviour.

While I'm at this, I was looking at other issues/PRs (mainly this one: https://github.com/testing-library/dom-testing-library/issues/372) and this seems like something desirable from the library POV. I'm not sure if this is the right place to ask this, but what is the reason behind this?

kentcdodds:
Sorry for the delay. If you would like to make that update based on your text node version I would be willing to merge that. I'd love a review from someone else watching the repo though :)

PR #458, opened 2020-03-04
Release Beta

kentcdodds:
**What**: Release the `beta` branch as stable

**Why**: Because we have needed breaking changes

**How**: Please see the commits for granular changes

**Checklist**:

- [x] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs)
- [x] I've prepared a PR for types targeting [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__dom)
- [x] Tests
- [x] Ready to be merged

Please test this in your projects:

```
npm install @testing-library/dom@beta
# yarn add @testing-library/dom@beta
```


package.json
```diff
@@ -61,7 +59,8 @@
     "rules": {
       "import/prefer-default-export": "off",
       "import/no-unassigned-import": "off",
-      "import/no-useless-path-segments": "off"
+      "import/no-useless-path-segments": "off",
+      "no-console": "off"
```
weyert: Any reason why you globally disable `no-consol`-rule?

kentcdodds: I did and it was annoying.

eps1lon: Once of those things where I'd like a different set of rules for local dev. Basically a set of rules for prod and dev. Because we do like to catch dangling debug statements but we do sometimes need them for debugging locally.

timdeschryver:
I encountered a small (non blocking) typing problem with Angular Testing Library, because `@types/testing-library__dom` hasn't been updated with the latest changes. 

I'm not really comfortable with `@types`, but do they also publish beta versions or should we wait until this version gets released?

kentcdodds:
I don't know how DefinitelyTyped handles beta versions. I'm not going to bother. We'll be merging this before too long.

kentcdodds:
Just an update. I'll planning on working on this and doing the release next Tuesday or Wednesday. 👍👍

kentcdodds:
I'm ready to release this tomorrow. I'd like to get the green light from as many maintainers as possible. Please leave a review @testing-library/all-maintainers!

kentcdodds:
Keep in mind that we like to avoid major releases like this, so if there's anything you have in mind for a breaking change, please speak now or forever hold your peace... well, maybe not forever, but yeah.

timdeschryver:
👍 for Angular Testing Library

afontcu:
I'm giving it a go right know in Vue Testing Lib: https://github.com/testing-library/vue-testing-library/pull/125 

would appreciate your input on both the empty callback when using `wait`,  and also the failing stopwatch tests ([source](https://github.com/testing-library/vue-testing-library/blob/bump-dtl-major/src/__tests__/stopwatch.js)). Not sure how to implement it now 🤔 (it's failing on empty callback for wait, but I guess I need to rethink the whole test case)

benmonro:
Given that the no op option for wait was removed is there a work around for folks that have usages of it?  That one seems like it would cause a lot of breaking tests that would have to be fixed one by one.  Which is probably worth doing but my 2 cents would be to give a deprecation warning on that before removing it like the other async functions.

Speaking of which... Why get rid of wait for element and wait for element to be removed?  I think they really improve the readability of tests.  I get they can be done with wait or find* but for me it does improve the style of the test when you don't need to do an expect on those things...

kentcdodds:
Alright, I'm working on this now: https://twitter.com/kentcdodds/status/1238197276465483776

benmonro:
@kentcdodds any thoughts on my questions before you release? :)

kentcdodds:
Yes, that's what I'm doing in the livestream 😉

benmonro:
damn i'm in meetings.  i can't make it

kentcdodds:
> i'm in meetings. i can't make it

Don't worry, what I meant was, I'm typing out the answer in the livestream :)

benmonro:
ok posted some comments in the youtube comments

kentcdodds:
For the empty callback with `wait` the solution is quite simple:

```diff
- await wait()
+ await wait(() => {})
```

So that can be a find/replace update, and then the recommendation would be for people to go through and put assertions in that callback instead of having an empty callback.

As for `waitForElementToBeRemoved`, that is *not* getting deprecated or removed. But `waitForDomChange` is removed because that's basically what `wait` does now, and `waitForElement` is what `find*` queries do *and* if you can't use a query, then `wait` does that now.

So the solution for those will be:

```diff
- const element = await waitForElement(() => container.querySelector('.loading'))
+ const element = await wait(() => container.querySelector('.loading'))

- await waitForDomChange()
+ await wait(() => {})

- await waitForDomChange(mutationObserverOptions)
+ await wait(() => {}, mutationObserverOptions)
```

So that's actually pretty straightforward. Most of those should be find/replace. The idea is, because `wait` now supports mutation observer (as well as the interval), we no longer need these one-offs. Also `waitForDomChange` was testing implementation details (the user doesn't care about any old DOM change, they care about specific DOM changes), and the `wait` with a no-op callback is also implementation details (because the user doesn't care how many ticks of the event loop something takes, they just care about waiting for something specific to happen).

Regarding readability, I don't think they're any more readable personally. Especially since you shouldn't be using `waitForElement` very often (you should use a query). In the above diff, I don't see `waitForElement` being any more understandable than the `wait` version.

`waitForElementToBeRemoved` is kept around because it's harder to implement what it does in userland. So it provides some actual utility.

For `waitForDomChange` I'm not a huge fan of including things and linting against them. `debug` is special because it logs something to the terminal that people wouldn't need a lint rule to tell them to remove. `waitForDomChange` is something that you could use to make your test pass, so you may decide to use it to do that. If your really like it for your workflow, then you can copy/paste what we have now and move it to your project, but I really don't want to encourage anyone to rely on specific and single DOM changes in their tests without waiting for a specific thing to happen (ie an assertion to pass).

benmonro:
I guess one thing i'm hung up on is the naming, instead of `wait` what about naming it `waitFor` 

I've been teaching folks on my team not to use wait because of the fact that it doesn't use mutation observers.  now i'm going to have to teach them _to_ use it.  waitFor also solves the readability issue imo

kentcdodds:
The library we used to use was called `waitForExpect` and I didn't like that because it was too specific to `expect`. `waitFor` sounds pretty good and I don't really have arguments against it. I'd really like to hear what other people have to say about it.

Maybe I'll do a twitter poll too 😅

kentcdodds:
Hmmm... What if we alias `wait` to `waitFor` and then in the *next* major version bump we switch it?

I don't really like that but it's an idea.

benmonro:
not sure how i feel about the alias.  like we're already breaking changes anyway.  people are already going to have to replace waitForElement and waitForDomChange.

kentcdodds:
Yeah, let's not do the alias... I'm leaning toward renaming to `waitFor` though... Just not excited about the fallout... Don't want to rock the boat too much for something small like this.

kentcdodds:
Alright, I'm not excited about making this change, but I think it's for the best, so we'll be renaming `wait` (as it is in this PR) to `waitFor` (also, here's the poll where `waitFor` is coming out with a big lead anyway: https://twitter.com/benmonro/status/1238209158622535680)

kentcdodds:
Some people are suggesting: `await until(() => expect(foo).toBe(true))`.

Curious what others have to say. I'm still leaning toward `waitFor`...

benmonro:
I don't hate until, but I think i still prefer waitFor as that's more in line w/ the other helpers & history of the library

timdeschryver:
it should be possible to have eslint fix these occurrences of `wait` and `waitForDomChanges` to make the upgrade path easier. See https://github.com/testing-library/eslint-plugin-testing-library/issues/86

PR #459, opened 2020-03-04
Cleanup repo

MichaelDeBoey:


kentcdodds:
@all-contributors please add @MichaelDeBoey for code

NMinhNguyen:
Sorry to be commenting on a merged PR, but any chance the Node v10 requirement could be made a tiny bit lower, e.g. `10.12` ([ESLint](https://github.com/eslint/eslint/pull/12700/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2)) or `10.13` ([Jest](https://github.com/babel/babel/pull/10747/files))? Reason being I work in a big corporate where updating to the latest Node version isn't as easy as I would like it to be 😔 I understand that this is not directly this package's problem, but would like to understand why `10.18` specifically and not a bit lower?

kentcdodds:
Pretty sure we could manage regular node 10 actually

MichaelDeBoey:
@NMiinhNguyen I've set the minimum requirement to `v10.18`, because `semantic-release@17`, requires it
https://github.com/semantic-release/semantic-release/releases/tag/v17.0.0

kentcdodds:
But we only use semantic release when we're releasing the package and we only do today with node 12

MichaelDeBoey:
We do, but users of the `travis-release` script can make another choice if they want:

```yml
// example of a possible .travis.yml

jobs:
  include:
    - stage: release
      node_js: 10.13
      script: kcd-scripts travis-release
```

NMinhNguyen:
@MichaelDeBoey as far as the consumption of this library is concerned, `semantic-release` is not relevant because it's not a direct dependency of this package and is used when releasing only.

kentcdodds:
In any case, I think Node 10 is the version I want my packages supporting and if we have a dep that prevents that then we should change the dep or did the dep.

PR #460, opened 2020-03-04
feat(waitForElementToBeRemoved): support passing an element directly

kentcdodds:
**What**: feat(waitForElementToBeRemoved): support passing an element directly

**Why**: Makes using this async utility simpler

**How**: Overload the argument for this so the callback can be a DOM node (or array of DOM nodes). In that case, we capture those and reassign the callback to a function that checks whether those are still contained in their original parent element

**Checklist**:

- [x] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs): https://github.com/testing-library/testing-library-docs/pull/394
- [ ] I've prepared a PR for types targeting [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__dom) (Sorry, won't get to this. Will hope the community can make this happen later)
- [x] Tests
- [x] Ready to be merged
      

Really excited about this one.

### DIFF
--- a/src/__tests__/wait-for-element-to-be-removed.js
+++ b/src/__tests__/wait-for-element-to-be-removed.js
@@ -49,7 +49,7 @@ test('requires an element to exist first', () => {
   return expect(
     waitForElementToBeRemoved(null),
   ).rejects.toThrowErrorMatchingInlineSnapshot(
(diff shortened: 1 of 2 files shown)


kentcdodds:
Feel free to comment later if you have questions or feedback. We can address them before we publish beta.

PR #461, opened 2020-03-04
feat(wait): remove default no-op callback

kentcdodds:
**What**: Removes the default no-op callback from `wait`

**Why**: That encourages relying on the implementation detail of waiting for the "next tick" and explicit expectations should be used instead.

**How**: Removed the default and added a test to check for the rejected response

**Checklist**:

- [x] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs) https://github.com/testing-library/testing-library-docs/pull/394
- [ ] I've prepared a PR for types targeting [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__dom) (still in beta...)
- [x] Tests
- [x] Ready to be merged
      

BREAKING CHANGE: If you used `wait()` in the past, you now have to supply a callback. Relying on the "next tick" is an implementation detail and should be avoided in favor of explicit expectations within your wait callback.

### DIFF
--- a/src/__tests__/wait.js
+++ b/src/__tests__/wait.js
@@ -12,10 +12,9 @@ test('it waits for the data to be loaded', async () => {
 })
 
 test('wait defaults to a noop callback', async () => {
-  const handler = jest.fn()
-  Promise.resolve().then(handler)
-  await wait()
-  expect(handler).toHaveBeenCalledTimes(1)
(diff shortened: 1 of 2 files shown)


kentcdodds:
If you have comments or feedback, please feel free to comment and we can get things addressed before the beta is released.

PR #464, opened 2020-03-04
fix(prettyRoles): crash in older browsers

eps1lon:
**What**:

Fixes crash when logging roles in older browsers (namely IE11)

**Why**:

Objec.entries is not implemented in older browsers

**How**:

Use Object.keys

**Checklist**:

- ~[ ]~ I've prepared a PR for types targeting
- [ ] Tests (Testing codesandbox canary on Material-UI)
- [ ] Ready to be merged


eps1lon:
Requires a bit more work (https://github.com/facebook/jest/pull/9635) but doesn't hurt to get this in anyway. At least this way I can hopefully fall back to `pretty-format@24`

PR #469, opened 2020-03-07
fix: use wait in findBy queries

### LINKED ISSUE #468: `waitForElement` has been deprecated message showing even when not used
- `@testing-library/react` version: ^10.0.0-beta.2
- `react` version: 16.13.0
- `node` version: 12
- `npm` (or `yarn`) version: 1.12.2

### Relevant code or config:

```js
  it.only('should not submit with empty fields', async () => {
    const { username } = renderLogin();

    user.type(username, variables.username);
    user.click(screen.getByText(/Login/i));
    await screen.findByText('Password is required');
    screen.debug();
  });
```

### What you did:

Run above test with latest beta version

### What happened:

Test passes however in my console I get a deprecation warning for API I didn't use

```
 console.warn node_modules/@testing-library/dom/dist/wait-for-element.js:17
    `waitForElement` has been deprecated. Use a `find*` query (preferred: https://testing-library.com/docs/dom-testing-library/api-queries#findby) or use `wait` instead (it's the same API, so you can find/replace): https://testing-library.com/docs/dom-testing-library/api-async#wait
```

### Suggested solution:

Only show above warning if waitForElement is used

image

timdeschryver:
**What**: Using the `wait` method instead of `waitForElement` for the `findBy` queries

**Why**: To prevent that deprecations are logged

 Closes #468

**How**: Replace `waitForElement` with `wait` (and by renaming `waitForElementOptions` to `waitOptions`)

**Checklist**:

- [ ] Tests
- [x] Ready to be merged
      

Something else that I've encountered, is that the changes made in #461 didn't make it into the [beta branch](https://github.com/testing-library/dom-testing-library/blob/beta/src/wait.js#L12).

### DIFF
--- a/src/query-helpers.js
+++ b/src/query-helpers.js
@@ -1,5 +1,5 @@
 import {fuzzyMatches, matches, makeNormalizer} from './matches'
-import {waitForElement} from './wait-for-element'
+import {wait} from './wait'
 import {getConfig} from './config'
 
 function getMultipleElementsFoundError(message, container) {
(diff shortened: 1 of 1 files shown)


timdeschryver:
In case you missed it, while making these changes I saw the changes made in #461 didn't make it into the [beta branch](https://github.com/testing-library/dom-testing-library/blob/beta/src/wait.js#L12), and I'm not sure where these changes are...

PR #478, opened 2020-03-13
docs: corrects deprecation warning for waitForDomChange

egilsster:
**What**:

If you are migrating from `waitForDomChanges`, it suggests using `wait` instead, which is also deprecated and it suggests `waitFor`. The link in this particular warning is linked to `waitFor`.

```
`waitForElement` has been deprecated. Use a `find*` query ...
```

```
`waitForDomChange` has been deprecated. Use `wait` instead ...
```

but then it says `wait` is also deprecated:

```
`wait` has been deprecated and replaced by `waitFor` instead. In most cases you should be able to find/replace `wait` with `waitFor`.
```

**Why**:

☝️ 

**How**:

Updated deprecation warning string with the correct suggestion

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/wait-for-dom-change.js
+++ b/src/__tests__/wait-for-dom-change.js
@@ -38,7 +38,7 @@ test('waits for the dom to change in the document', async () => {
   expect(console.warn.mock.calls).toMatchInlineSnapshot(`
 Array [
   Array [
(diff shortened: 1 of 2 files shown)


kentcdodds:
@all-contributors please add @egilsster for code

PR #487, opened 2020-03-15
fix(allByLabelText): forEach on NodeList is not supported in edge

eps1lon:
Copy of #463 which got lost when merging beta into master.


kentcdodds:
Still no idea why we lost three changes. But hopefully this was the last! Thank you.

PR #489, opened 2020-03-18
upgrade dom-accessibility-api

ysgk:
**What**: Upgraded dom-accessibility-api to ^0.4.1

**Why**: To support es5 environment. dom-accessibility-api supports es5 from 0.4.1.

PR: https://github.com/eps1lon/dom-accessibility-api/pull/141

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@babel/runtime": "^7.8.4",
     "@types/testing-library__dom": "^6.12.1",
     "aria-query": "^4.0.2",
-    "dom-accessibility-api": "^0.3.0",
+    "dom-accessibility-api": "^0.4.1",
     "pretty-format": "^25.1.0"
   },
   "devDependencies": {


eps1lon:
I need to release it as 1.x anyway. 0.x semantics aren't helpful right now (API is stable, features should be picked up by `^`).

ysgk:
OK, I'll wait for it. Thanks.

MichaelDeBoey:
When doing #493, I kept getting the following error:

```sh
[build] [umd.min] [!] Error: 'computeAccessibleName' is not exported by node_modules/dom-accessibility-api/dist/index.js, imported by src/role-helpers.js
```

So I kept the `dom-accessiblility-api` out of it

eps1lon:
@MichaelDeBoey Let me check this out. Common issue with rollup falsely flagging "not exported from".

eps1lon:
#493 will include the update to dom-accessibility-api. Still leaving this at 0.x. Just need to remember to do patch releases.

PR #493, opened 2020-03-21
Update dependencies + use kcd-scripts' husky

MichaelDeBoey:


package.json
```diff
@@ -17,10 +17,11 @@
     "end-to-end",
     "e2e"
   ],
-  "author": "Kent C. Dodds <me@kentcdodds.com> (https://kentcdodds.com)",
+  "author": "Kent C. Dodds <kentcdodds@gmail.com> (https://kentcdodds.com)",
```
kentcdodds: Let's keep this as me@kentcdodds.com

MichaelDeBoey: I just copied it over from `generator-kcd-oss`, will change it again 🙂

package.json
```diff
@@ -20,7 +20,8 @@
   "author": "Kent C. Dodds <me@kentcdodds.com> (https://kentcdodds.com)",
   "license": "MIT",
   "engines": {
-    "node": ">=10"
+    "node": ">=10",
+    "npm": ">=6"
```
eps1lon: This engines are for consuming code. I don't think this library requires npm.

MichaelDeBoey: It's also checked I think

eps1lon: What do you mean by "it's checked"?

MichaelDeBoey: When you run `npm install`, the version is checked.
Just like you have when using `yarn`.

eps1lon: Yeah but it shouldn't. I can install this library just as well with yarn or pnpm.

MichaelDeBoey: The version is only checked when using `npm`

eps1lon: > You can also use the “engines” field to specify which versions of npm are capable of properly installing your program.

-- https://docs.npmjs.com/files/package.json#engines

Why do we need npm 6 to install `@testing-library/dom`?

MichaelDeBoey: It's not necessary, but just setting it as a sensible baseline
Most people are already using `npm@6`

eps1lon: Any code that is not necessary should be removed.

kentcdodds: Yeah, I think we should probably just remove the npm engine. It's really not necessary.

MichaelDeBoey: I'll remove it then 🙂

PR #496, opened 2020-03-23
fix(fireEvent): Set composed property on relevant synthetic events

apalaniuk:
**What**:

When using the `fireEvent` interface to generate synthetic events, if [events which are normally composed](https://developer.mozilla.org/en-US/docs/Web/API/Event/composed) are dispatched on elements that are within the shadow DOM of a web component, and the listeners expected to handle that event type are registered on the custom element itself (a common configuration), the event listeners will not be triggered as expected, as the event will stop bubbling at the shadow DOM boundary. This deviates from the behaviour of the matching native browser events which are composed, which will propagate outside of the shadow DOM.

**Why**:

This change updates synthetic event init definitions to more closely match their related browser events by setting their `composed` property to `true` [if that event type is composed](https://w3c.github.io/uievents/). Without this change, workarounds which couple tests to implementation details are often required in order to find the parent element against which events may be dispatched and handled, rather than dispatching the event against the found element directly and allowing for standard event propagation.

**How**:

A `composed` property is added to existing event initialization definitions where relevant. This allows for events to be composed in any environment that has, at minimum, the `window.Event` constructor.

**Checklist**:

- [x] Tests
- [x] Ready to be merged


src/__tests__/events.js
```diff
@@ -137,6 +137,12 @@ const eventTypes = [
   },
 ]
 
+const allEvents = eventTypes
```
apalaniuk: I'm happy to revert this to be more straightforward, but it was getting a bit verbose/has potential for drift.

src/__tests__/events.js
```diff
@@ -137,6 +137,12 @@ const eventTypes = [
...
+  .reduce((acc, curr) => {
+    acc.push(...curr.events)
+    return acc
+  }, [])
```
kentcdodds: This is great. Thank you for doing this.

I think we could simplify it:

```suggestion
const allEvents = eventTypes.flatMap(({events}) => events)
```

kentcdodds: While we're at it, I think we could/should export the `eventMap` itself and then import it into this file. I don't think we want to expose it from the library though so we'll need to be thoughtful about how we do that. Perhaps we could put the `eventMap` in a separate file and import it into both this file and the `events.js` file.

What do you think? I know it's more than you signed up for so if you'd rather not do that in this PR that's fine. We can do that later.

apalaniuk: Thanks for the feedback (and the great libraries)!

I avoided `flat`/`flatMap` in case there were concerns about [support](https://node.green/#ES2019-features-Array-prototype--flat--flatMap-) as the current node engines field is `>= 10`, but if not, I also prefer simpler.

> Perhaps we could put the eventMap in a separate file and import it into both this file and the events.js file

Had similar thoughts so, definitely agree.

The changes are straightforward, so it's no problem to do now, I think. :+1:

src/event-map.js
```diff
@@ -0,0 +1,350 @@
...
+      defaultInit: {bubbles: true, cancelable: true},
+    },
+    // pointer events
+    pointerOver: {
+      EventType: 'PointerEvent',
+      defaultInit: {bubbles: true, cancelable: true, composed: true},
+    },
+    pointerEnter: {
+      EventType: 'PointerEvent',
+      defaultInit: {bubbles: false, cancelable: false},
```
apalaniuk: I'm deferring to https://w3c.github.io/pointerevents/index.html#attributes-and-default-actions, but I noticed that Chromium-based browsers are non-conformant and set these as composed, while at least Firefox seems conformant. Calling it out in case you have a different preference.

src/__tests__/events.js
```diff
@@ -185,45 +188,61 @@ const bubblingEvents = [
   'pointerUp',
   'pointerCancel',
   'pointerOut',
+  'popState',
 ]
 
-const nonBubblingEvents = [
+const nonBubblingEvents = allEvents.filter(evt => !bubblingEvents.includes(evt))
+
+const composedEvents = [
```
kentcdodds: Is it possible now to derive these values from the events map?

src/__tests__/events.js
```diff
@@ -1,9 +1,10 @@
+import {eventMap, eventAliasMap} from '../event-map'
 import {fireEvent} from '..'
 
 const eventTypes = [
   {
     type: 'Clipboard',
-    events: ['copy', 'paste'],
+    events: ['copy', 'cut', 'paste'],
```
apalaniuk: Would be nice to derive more from `eventMap` to make sure missing events here aren't a concern, but this doesn't drive all event tests/isn't intended a complete reindexing of `eventMap`, so would require a bit of test rewriting that might be better at some other time to maintain momentary stability (plus when I have some actual time)

kentcdodds:
@all-contributors please add @apalaniuk for code and tests

PR #515, opened 2020-03-31
feat(waitFor): improve error message for non-function callbacks

### LINKED ISSUE #509: Improve error message for non-function arg passed to `waitFor`
New `waitFor` async utils expects a callback function as first argument. Though if something else is passed there, the error looks like this:
wait-for

It's difficult to figure out the actual error with that message (it took me few minutes to found this in a project where I was migrating dom-testing-library to the last version!). I'd like to get a better message for this case.

### Suggested implementation:
This would be as simple as throw an error at first line of `waitFor` implementation if `typeof callback !== 'function'`. I'm happy to implement this myself, though I'm not sure which message should I write so I'd like to discuss that here.

n/a

n/a

Belco90:
**What**:
Closes #509

**Why**:
To improve error message and error stack trace

**How**:
Checking if `callback` is a function.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/wait-for.js
+++ b/src/__tests__/wait-for.js
@@ -1,3 +1,4 @@
+import {renderIntoDocument} from './helpers/test-utils'
 import {waitFor} from '../'
 
 test('waits callback to not throw an error', async () => {
@@ -34,7 +35,11 @@ test('uses generic error if there was no last error', async () => {
 })
 
 test('throws nice error if provided callback is not a function', () => {
-  expect(() => waitFor('not a function')).toThrow(
+  const {queryByTestId} = renderIntoDocument(`
+    <div data-testid="div"></div>
+  `)
+  const someElement = queryByTestId('div')
+  expect(() => waitFor(someElement)).toThrow(
     'Received `callback` arg must be a function',
   )
 })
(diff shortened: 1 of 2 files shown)


src/wait-for.js
```diff
@@ -22,6 +22,10 @@ function waitFor(
     },
   } = {},
 ) {
+  if (typeof callback !== 'function') {
+    throw new Error('Received `callback` arg must be a function')
```
eps1lon:
```suggestion
    throw new TypeError('Received `callback` arg must be a function')
```

src/__tests__/wait-for.js
```diff
@@ -32,3 +32,9 @@ test('uses generic error if there was no last error', async () => {
   ).catch(e => e)
   expect(result).toMatchInlineSnapshot(`[Error: Timed out in waitFor.]`)
 })
+
+test('throws nice error if provided callback is not a function', () => {
+  expect(() => waitFor('not a function')).toThrow(
```
eps1lon: Could you use a more reasonable argument here like a DOM element?

Belco90: Done. I hope that element I'm passing looks fine.

eps1lon:
@Belco90 Looking good, thanks!

PR #525, opened 2020-04-11
test(ci): use latest 10 LTS release

### LINKED ISSUE #521: Build failure on Node 10
As noted here: https://github.com/testing-library/dom-testing-library/pull/520#issuecomment-609961780

image

```
[test]  FAIL   node  src/__node_tests__/index.js
[test]   ● Test suite failed to run
[test] 
[test]     ReferenceError: BigInt is not defined
[test] 
[test]       at createLongLongConversion (node_modules/webidl-conversions/lib/index.js:162:51)
[test]       at Object.<anonymous> (node_modules/webidl-conversions/lib/index.js:224:24)
[test]       at Object.<anonymous> (node_modules/jsdom/lib/jsdom/browser/Window.js:3:27)
[test] 
```

I have no idea why that's happening. I'm guessing that we've got a devDependency that does not support Node 10. This is unfortunate and I'm kinda thinking the easiest solution is to just drop Node 10 from CI and hope we don't ever ship anything that breaks Node 10... 😬 That said, Node 10 doesn't get EOL until mid-2021 (and that's how long I think we should support it) so it may be better to find another solution to this.

Any ideas?

alexkrolick:
Closes #521

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ cache: npm
 notifications:
   email: false
 node_js:
-  - 10.0.0
-  - 12
+  - lts/dubnium
+  - lts/erbium
   - node
 install: npm install
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ cache: npm
 notifications:
   email: false
 node_js:
-  - lts/dubnium
+  - 10
   - lts/erbium
   - node
 install: npm install
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ notifications:
   email: false
 node_js:
   - 10
-  - lts/erbium
+  - 12
   - node
 install: npm install
 script:
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "author": "Kent C. Dodds <me@kentcdodds.com> (https://kentcdodds.com)",
   "license": "MIT",
   "engines": {
-    "node": ">=10.4"
+    "node": ">=10"
   },
   "scripts": {
     "build": "kcd-scripts build --ignore \"**/__tests__/**,**/__node_tests__/**,**/__mocks__/**\" && kcd-scripts build --bundle --no-clean",
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "author": "Kent C. Dodds <me@kentcdodds.com> (https://kentcdodds.com)",
   "license": "MIT",
   "engines": {
-    "node": ">=10"
+    "node": ">=10.4"
   },
   "scripts": {
     "build": "kcd-scripts build --ignore \"**/__tests__/**,**/__node_tests__/**,**/__mocks__/**\" && kcd-scripts build --bundle --no-clean",


.travis.yml
```diff
@@ -3,8 +3,8 @@ cache: npm
 notifications:
   email: false
 node_js:
-  - 10.0.0
-  - 12
+  - lts/dubnium
+  - lts/erbium
```
alexkrolick, line 7: I don't know if these are valid versions for config, kicking CI to see

alexkrolick: It works https://travis-ci.org/github/testing-library/dom-testing-library/builds/673703620

MichaelDeBoey:
I wouldn’t merge this PR for 2 reasons:

1. It’s a breaking change
2. This will run the tests on latest Node 10 and I think we want to make sure the package is working on the lowest supported version.
  This way we can’t be sure it keeps working on 10.0

The breaking change should be reverted upstream instead.

alexkrolick:
- It's not a breaking change, this is for our devDeps.
- CI already runs on Node 10 and 12 (but is failing on Node 10.0.0), this does the same, but uses the latest 10.x and 12.x releases.

MichaelDeBoey:
@alexkrolick It's indeed breaking, since you're now saying the package can only be consumed by Node `10.4` instead of `10.0`.

CI is testing for CI `10.0` (the minimum supported version), so if we change that to `10` or `dubnium` the latest `10.x` (which is `10.20`) will be used.
That will not insure us anymore that our code is working on `10.0`.

Right now CI is failing for `10.0`, because of the changes in [`webidl-conversions`](https://github.com/jsdom/webidl-conversions).
They released a [major release](https://github.com/jsdom/webidl-conversions/releases/tag/v6.0.0) for this, so they're good.

The problem is that [`jsdom`](https://github.com/jsdom/jsdom) changed it's dependency to `webidl-conversions@6.x` without making this a breaking change on their side (see https://github.com/jsdom/jsdom/commit/fe164b3b39f61846193b1f2d06ef7c27da554ebf).

Fixing this in `jsdom`, would fix our CI again.

kentcdodds:
I honestly am not concerned about running CI against the latest 10 version and just hoping we don't break 10.0.0. I think it's pretty unlikely that we will break 10.0.0 and if we do it's not a lot of work to fix it.

MichaelDeBoey:
We will know something breaking a lot faster in the process, which will cause less bug reports, ...

Not my call to make, but that's why I think we shouldn't use latest Node `10.x`

kentcdodds:
You're not wrong. I just see the risk as low and the reward as high so I'm willing to take it.

kentcdodds:
Not only a lot risk, but also a lot impact as well.

alexkrolick:
We could run CI against the lowest version and all the current lts versions we support. That would be 10.0.0, lts/dubnium, lts/erbium.

But what to do with the engines field? Our JSDOM dep has silently raised the number from 10.0.0 to 10.4.0. Do we want to leave it at 10.0.0, keep running CI against 10.0.0, and have it fail until upstream reverts or patches?

kentcdodds:
Let's leave engines as they are and bump CI just so we can get it working. Just because we're having trouble with our development dependencies doesn't mean other people should suffer the consequences of that.

.travis.yml
```diff
@@ -3,8 +3,8 @@ cache: npm
 notifications:
   email: false
 node_js:
-  - 10.0.0
-  - 12
+  - lts/dubnium
```
MichaelDeBoey, line 6:
```suggestion
  - 10
```

.travis.yml
MichaelDeBoey, line 7:
```suggestion
  - 12
```

.travis.yml
alexkrolick: @MichaelDeBoey why do you want to change that?

MichaelDeBoey: The latest Node 10 version will be used.

This will make the change in this PR as minimal as possible

kentcdodds: Yeah, I'm pretty sure that all that needs to happen for this is to change `10.0.0` to `10` and we'll be good. The release names are a little confusing and need a reference lookup anytime we want to check them 😬

alexkrolick: Ok. Committed. I always use the release names with "lts/" to remind people not to change it to the new non-lts release whenever it comes out.

PR #526, opened 2020-04-12
fix: apply spread syntax to container.querySelectorAll() in alt-text.js

Pewww:
**What**:
Improving code in queries/alt-text.js

**Why**:
I changed it because spread syntax can replace Array.from() in the code.

Parameter 'c' was not used within those functions, so I replaced it to '_' to make code more clear and intuitive.

**How**:
I applied spread syntax to container.querySelectorAll().
And, I also changed first parameter of getMultipleError and getMissingError, 'c' to '_'.

**Checklist**:

- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/src/queries/alt-text.js
+++ b/src/queries/alt-text.js
@@ -7,14 +7,14 @@ function queryAllByAltText(
 ) {
   const matcher = exact ? matches : fuzzyMatches
   const matchNormalizer = makeNormalizer({collapseWhitespace, trim, normalizer})
-  return Array.from(container.querySelectorAll('img,input,area')).filter(node =>
+  return [...container.querySelectorAll('img,input,area')].filter(node =>
     matcher(node.getAttribute('alt'), node, alt, matchNormalizer),
   )
 }
 
-const getMultipleError = (c, alt) =>
+const getMultipleError = (_, alt) =>
   `Found multiple elements with the alt text: ${alt}`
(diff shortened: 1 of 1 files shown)


Pewww:
@eps1lon This PR is closer to 'changing' than 'fixing error'.
I think I expressed like Array.from() has an issue, but it's not and sorry for it.

I thought that spread syntax is shorter and more readable than Array.from().
Also, it has performance advantage even if it's subtle.

[https://www.measurethat.net/Benchmarks/Show/1721/0/arrayfrom-vs-spread](https://www.measurethat.net/Benchmarks/Show/1721/0/arrayfrom-vs-spread)
You may have questions about the example, because testing target is not about NodeList.
But, unless copying target is not multidimensional arrays, it can still have benefits like this picture.
image

Pewww:
Now that I see, this PR doesn't seem very meaningful.
If it's unnecessary, you can close it.
Thank's for your comment :D

eps1lon:
Yeah we still have to transpile it to older environments and this will likely result in the same code. Specifically, Array.from is more readable because that is the intention: convert the iterator to an array. Spreading can have different use cases.

Pewww:
@eps1lon I apologize to mention you again in closed PR.
I became curious about two syntax(Array.from(), spread syntax) that we talked about last time.
If you don't mind, may I ask you a question?

Issue #494, opened 2020-03-22
Convert codebase to TypeScript


abraham:
### Describe the feature you'd like:

Accurate and timely TypeScript types.

Currently types are manually maintained in the `@types/` project so they are only updated after the fact and are prone to human error and implementation drift.

### Suggested implementation:

Convert `@testing-library/dom` to TypeScript.

### Describe alternatives you've considered:

The main alternative is to leave things the way they are. Unfortunately this leaves types manually updated after the fact. This is slow, depend on third-parties (Microsoft), and is error prone.

- 10+ days after TL update without updated types https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43102
- Requests for updated types
  - https://github.com/testing-library/dom-testing-library/issues/490
  - https://github.com/testing-library/dom-testing-library/issues/486
  - https://github.com/testing-library/dom-testing-library/issues/226
- Reports of incorrect types
  - https://github.com/testing-library/dom-testing-library/issues/407
  - https://github.com/testing-library/dom-testing-library/issues/308
  - https://github.com/testing-library/dom-testing-library/issues/291
  - https://github.com/testing-library/dom-testing-library/issues/149
  - https://github.com/testing-library/dom-testing-library/issues/142
  - https://github.com/testing-library/dom-testing-library/issues/135
  - https://github.com/testing-library/dom-testing-library/issues/42

I'm sure there are other issues filed on the DefinitelyTyped repo.

Another alternative is to require DefinitelyTyped types as part of general contributions to this project. This would be a big increase to the barrier of entry as contributors would have to know how to write types and update DefinitelyTyped.

### Teachability, Documentation, Adoption, Migration Strategy:

I'd look at doing a multi-stage conversion

1. Initial proof of concept with one or two files to validate the approach and tooling.
1. The `src` files
2. The `tests` files (this could be delayed until a future date)
4. Update testing-library.com documentation to display the latest types (optional but if they exists they might as well be used)

TypeScript does add some additional complexity in tooling and would increase the barrier of entry for contributors. This is also an opportunity to to create a simple and supportive implementation that people less familiar with TypeScript and contribute to. It would be important to create some good contributing guides on working with types (or identify and recommend some good third-party guides).

Potentially there would be bugs uncovered and fixed in the codebase with the conversion.

kentcdodds:
I definitely understand the desire here. I'll hold on giving my opinion until I hear what other maintainers have to say about it.

eps1lon:
[...]

PR #530, opened 2020-04-22
Copied types from DefinitelyTyped into this repository

Lagily:
**What**: Copied types from DefinitelyTyped into this repository and adjusted the package.json, so these types are packed into the npm package. Also removed the dependency to the DefinitelyTyped types, since they are now in this repo.

**Why**: See https://github.com/testing-library/dom-testing-library/issues/494#issuecomment-602720977

**Checklist**:

- [?] Documentation added to the
- [N/A] I've prepared a PR for types targeting
- [x] Tests
- [x] Ready to be merged

Not sure as to where and how you would like documentation to be added. I don't think additional documentation is needed (yet). We should instead make a PR to remove dom-testing-library from DefinitelyTyped - what do you think?


eps1lon:
@kentcdodds Would you be open to adding type annotations via JSDOC? 

We could generate the declarations directly from the source files while simultaneously checking that the types are correct (and actually type check the source files). 

That way we keep any TypeScript syntax out and the current workflow doesn't change.

At least I think this is possible. The typings are simple enough that JSDOC should be sufficient.

kentcdodds:
I think that would be a good iterative approach to migrating to TypeScript (if that's what we end up doing). I'd be willing to take a look at that, yeah.

Lagily:
Sure, I think DefinitelyTyped also runs some sort of tests on their types, I'll investigate and add something

Lagily:
I've added the type tests like DefinitelyTyped does it, I commited in a way that you can see the tests fail, before adjustments are made. If you change the type-tests.ts in a way that types don't make sense (as in it doesn't compile anymore) the npm verify step will now fail.

I honestly don't think it's too much work to port this code base to TypeScript, I personally wouldn't add any more intermediate steps.

package.json
```diff
@@ -29,20 +30,23 @@
...
   "dependencies": {
     "@babel/runtime": "^7.9.2",
-    "@types/testing-library__dom": "^7.0.0",
     "aria-query": "^4.0.2",
     "dom-accessibility-api": "^0.4.2",
     "pretty-format": "^25.1.0"
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^5.1.1",
+    "conditional-type-checks": "^1.0.5",
```
eps1lon, line 48: Why do we need this?

Lagily: that's an opsie, thanks :D removed it

Lagily:
I might give it a shot to port the code base to TypeScript - should I wait out on that?

eps1lon:
> I might give it a shot to port the code base to TypeScript - should I wait out on that?

For now we don't want to use actual TypeScript. After we resolved this PR I'm going to try out types via JSDOC.

types/wait-for-element-to-be-removed.d.ts
```diff
@@ -0,0 +1,8 @@
+export function waitForElementToBeRemoved<T>(
```
timdeschryver, line 1: Fyi, I made a PR to DefinitelyTyped to add the `interval` option here.
It got merged today, so I believe we should also add it here.

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44006/files

Lagily: actually since options is simply passed to waitFor, it should just use the same type, so we don't have to redefine it in multiple places, I'll take care of it

timdeschryver: That's a good idea 👍

eps1lon:
Regarding JSDOC: It doesn't seem like we can incorporate TS types at the moment. I wouldn't know how to type `events.js`.

kentcdodds:
Ok, I'm thinking this is a good idea. Shame about the JSDOC stuff. Let's go ahead and get this merged and then we can explore migrating to TS totally.

Looks like there's a conflict in the package.json (sorry, my bad). I just released a major version to kcd-scripts which now will automatically run a `typecheck` script if it's available. So you don't need to worry about adding that to the `validate` script.

Lagily:
Hmm I tried pulling master into the branch, but it seems that it causes a test to fail. I have pushed the merge into this pull request (and removed the npm validate typecheck). This is the failing test on my machine, after pulling master into it:


 FAIL  src/__tests__/role-helpers.js (5.491s)
  ● shouldExcludeFromA11yTree for <div style="visibility: hidden;"/> returns true
[test]
    expect(received).toBe(expected) // Object.is equality
[test]
[test]     Expected: true
[test]     Received: false
[test]
      183 |   container.firstChild.appendChild(document.createElement('button'))
[test]       184 | 
    > 185 |   expect(isInaccessible(container.querySelector('button'))).toBe(expected)
          |                                                             ^
      186 | })
[test]       187 | 
[test]
      at src/__tests__/role-helpers.js:185:61

kentcdodds:
Sorry, clicked the wrong button.

Also sorry about that, I mistakenly removed `jsdom` from the dev deps even though I forgot we actually use it directly for our node tests.

Go ahead and rebase again if you please and make sure to re-install deps. Thanks!

Lagily:
Sure! Will hopefully run through now - does merging this PR count as a 'migrate to TS pull request is welcome'? :-D

kentcdodds:
A pull request would be welcome, but I'm not certain that it would be merged.

To be frank, I'm insecure about my personal TypeScript abilities (just haven't had enough time with it honestly) so I've been hesitant to add TS to my projects primarily for that reason (though the barrier to contributions is also a consideration).

So I'm willing to take a look and provide feedback on a PR, but I'm not 100% certain that it would get merged.

kentcdodds:
Is there anything left on this PR?

I want to leave this open for a day or two more to make sure everyone who's got an opinion shares it.

Merging this PR means that every API change should be associated with updates to the TypeScript definitions, so it has important implications.

Lagily:
Fair enough. From my point of view, this PR is complete.

kentcdodds:
Ok, let's do this.

nstepien:
I use `@testing-library/react` and I'm getting these errors now:
```
> tsc -p tsconfig.all.json

node_modules/@testing-library/dom/types/queries.d.ts:3:32 - error TS2307: Cannot find module 'wait-for'.

3 import { waitForOptions } from 'wait-for';
                                 ~~~~~~~~~~

node_modules/@testing-library/dom/types/wait-for-dom-change.d.ts:1:32 - error TS2307: Cannot find module 'index'.

1 import { waitForOptions } from "index";
                                 ~~~~~~~

node_modules/@testing-library/dom/types/wait-for-element-to-be-removed.d.ts:1:32 - error 
TS2307: Cannot find module 'wait-for'.

1 import { waitForOptions } from "wait-for";
                                 ~~~~~~~~~~

node_modules/@testing-library/dom/types/wait-for-element.d.ts:1:32 - error TS2307: Cannot find module 'wait-for'.

1 import { waitForOptions } from "wait-for";
                                 ~~~~~~~~~~


Found 4 errors.
```
Looks like you need to use relative paths.

Lagily:
@kentcdodds could you also add me as a contributor?

eps1lon:
@all-contributors add @Lagily for code

PR #538, opened 2020-04-30
Fix waitFor not waiting for an async element

lourenci:
I don't know why yet, but the `waitFor` is not waiting for some elements on cases whereas `waitForElement` is.

I felt more comfortable to open a PR with a test case demonstrating the "situation". Feel free to close the PR for whatever.

Thanks.

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/wait-for.js
+++ b/src/__tests__/wait-for.js
@@ -55,6 +55,12 @@ test('waits and returns an async element', async () => {
     document.querySelector('div').appendChild(element)
   }, 0)
 
-  const newElement = await waitFor(() => document.querySelector('[id="test"]'))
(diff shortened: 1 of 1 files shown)


kentcdodds:
If you can't use a `find*` query, then this is how I would approach what you're trying to accomplish: https://github.com/testing-library/dom-testing-library/pull/538/commits/c5926d29d3b667746cf9b132ea4f85f28cd86af5

lourenci:
I got it and it makes sense for me.

Maybe we should update the docs to be more clear about that? I was guided to understand that I can just change `waitForElement` for `findBy*` or `waitFor`.

Thanks.

PR #540, opened 2020-05-03
feat(ByRole): Allow filter by selected state

eps1lon:
**What**:

Allows to filter results of `ByRole` by their `aria-selected` state via `getByRole(role, { selected: true })`.

There were some concerns allowing filtering by aria attributes (can't find the issue anymore) since users could mistakenly add them arbitrarily to create selectors that look accessible while creating broken UIs.

This is why I think we should restrict filter options to attributes that are only allowed on certain roles. ~For example, I wouldn't implement `getByRole(role, { disabled: true })` since `aria-disabled` is a global attribute i.e. can be applied to any element~<sup>1</sup>. If people want to filter by aria attributes on arbitrary elements then they always have the `getAllByRole(...).filter()` escape hatch.

<sup>1</sup> I was looking at aria-disabled in ARIA 1.1. It is restricted to certain elements in 1.2. The gist still applies.

**Why**:

Spec-compliant convenience wrapper around `getAllByRole(role).filter(el => el.getAttribute('aria-selected') === "true")`

**How**:

- consider implicit and explicit value of `aria-selected`
- reject `selected` if the role does not support this attribute

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs): https://github.com/testing-library/testing-library-docs/pull/458
- [x] Tests
- [x] Ready to be merged


src/__tests__/ariaAttributes.js
```diff
@@ -0,0 +1,94 @@
...
+  <div role="option" aria-selected="false" id="unselected-listbox-option" />
+  <div role="option" id="not-selectable-listbox-option" />
+</div>
+<div role="tree">
+  <div role="treeitem" aria-selected="true" id="selected-treeitem" />
+  <div role="treeitem" aria-selected="false" id="unselected-treeitem" />
+  <div role="treeitem" id="not-selectable-treeitem" />
+</div>
+<input type="radio" aria-selected="true" id="selected-native-radio" />
+<input type="radio" aria-selected="false" id="unselected-native-radio" />
```
kentcdodds, line 29: Is this not implied for radio buttons? Do people have to explicitly add the attribute for everything?

eps1lon: I had this originally but `aria-checked` is different to `aria-selected`. `aria-checked` is implied.

To be honest I don't really understand why you'd want aria-selected on a radio. Apparently assistive technology "may" imply it from focus but that wouldn't apply here.

Right now it's only there for parity with aria-query. 

Though it is removed from the latest [aria editors draft](https://rawgit.com/w3c/aria/master/#aria-selected) and from what I've been told from jsdom folks (who work at google) implementors only use editors drafts.

I'll open an issue on `aria-query` and see how they feel about following editors draft instead. Though response time is quite low so this might take a while.

eps1lon: It's also on the 1.2 working draft. Probably a good time to send a PR to aria-query with aria 1.2

eps1lon: Let's remove this test since it's no longer part of aria 1.2 anyway.

Issue #539, opened 2020-05-01
Improve error stack traces for async errors


kentcdodds:
Here's the stack trace I get:

```
      at Object.getElementError (node_modules/@testing-library/react/node_modules/@testing-library/dom/dist/config.js:30:19)
      at node_modules/@testing-library/react/node_modules/@testing-library/dom/dist/query-helpers.js:76:38
      at node_modules/@testing-library/react/node_modules/@testing-library/dom/dist/query-helpers.js:59:17
      at node_modules/@testing-library/react/node_modules/@testing-library/dom/dist/query-helpers.js:86:84
      at checkCallback (node_modules/@testing-library/react/node_modules/@testing-library/dom/dist/wait-for.js:56:22)
```

Here's the line of code that caused the error:

```javascript
await screen.findByRole('heading', {name: /example/i})
```

I think rather than a stacktrace in the `@testing-library` code, we should give them an error that has the stacktrace in their own code. Similar to what we do for [timeout errors](https://github.com/testing-library/dom-testing-library/blob/9007a673a138a50becd764c58d08c28477bcfc33/src/wait-for.js#L29-L30).

Kinda thinking it could be as simple as changing this:

https://github.com/testing-library/dom-testing-library/blob/9007a673a138a50becd764c58d08c28477bcfc33/src/wait-for.js#L65-L67

to:

```javascript
function onTimeout() {
  let error = timedOutError
  if (lastError) {
    error = lastError
    const userStackTrace = timedOutError.stack.split('\n').slice(1).join('\n')
    error.stack = `${error.stack.split('\n')[0]}\n${userStackTrace}`
  }
  onDone(error, null)
}
```

Technically we do lose some information with this, but I'm pretty sure it's information worth losing. Thoughts?

alexkrolick:
Good idea, it's pretty useless to get a stack trace into the library

eps1lon:
> Good idea, it's pretty useless to get a stack trace into the library

If we assume that we're bug free, sure. But I think we should still have the opportunity to print the whole stack trace. Otherwise debugging can become an issue.

kentcdodds:
I agree, any ideas on how to do that? We could combine the stack traces I guess.

eps1lon:
> I agree, any ideas on how to do that?

Add a config option `showFullStackTrace` and don't mutate the stack if it is true?

kentcdodds:
[...]

PR #542, opened 2020-05-04
Improve error stack traces for async errors

rahulchavan30:
**What**:
Improve error stack traces for async errors

**Why**:
Improve error stack traces for async errors

**How**:
-changed the on timeout function  

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged
      

i have made changes suggested by KCD , please let me know if i need to do any more changes will be happy to do ... this is my first PR to Open source

### DIFF
--- a/src/__tests__/wait-for.js
+++ b/src/__tests__/wait-for.js
@@ -60,7 +60,7 @@ test('does not change the stack trace if the thrown error is not a TestingLibrar
   expect(result.stack).toBe(originalStackTrace)
 })
 
-test('provides a stack trace from the if the throw error is a TestingLibraryElementError', async () => {
+test('provides an improved stack trace if the thrown error is a TestingLibraryElementError', async () => {
   const error = new Error('Throws the full stack trace')
   error.name = 'TestingLibraryElementError'
   const originalStackTrace = error.stack
@@ -70,6 +70,8 @@ test('provides a stack trace from the if the throw error is a TestingLibraryElem
     },
     {timeout: 8, interval: 5},
   ).catch(e => e)
+  // too hard to test that the stack trace is what we want it to be
+  // so we'll just make sure that it's not the same as the origianl
   expect(result.stack).not.toBe(originalStackTrace)
 })
 
--- a/src/__tests__/wait-for.js
+++ b/src/__tests__/wait-for.js
@@ -42,7 +42,7 @@ test('uses full stack error trace when showOriginalStackTrace present', async ()
     },
     {timeout: 8, interval: 5, showOriginalStackTrace: true},
   ).catch(e => e)
-  expect(result).toBe(error)
+  expect(result.stack).toBe(error.stack)
 })
 
 test('throws nice error if provided callback is not a function', () => {
--- a/src/__tests__/wait-for.js
+++ b/src/__tests__/wait-for.js
@@ -34,6 +34,17 @@ test('uses generic error if there was no last error', async () => {
   expect(result).toMatchInlineSnapshot(`[Error: Timed out in waitFor.]`)
 })
 
+test('uses full stack error trace when showOriginalStackTrace present', async () => {
+  const error = new Error('Throws the full stack trace')
+  const result = await waitFor(
+    () => {
+      throw error
+    },
+    {timeout: 8, interval: 5, showOriginalStackTrace: true},
+  ).catch(e => e)
+  expect(result).toBe(error)
+})
+
 test('throws nice error if provided callback is not a function', () => {
   const {queryByTestId} = renderIntoDocument(`
     <div data-testid="div"></div>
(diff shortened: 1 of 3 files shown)


marcosvega91:
Hi @rahulchavan30 good job for your first contribution 🥳 . As I understand this behavior should be enabled based on a config parameter.
I think that you can add a new property in src/config.js and use it in wait-for.js.
Maybe It's good to add also a test case.  
Last thing, to make the PR visibile in the issue you can add a reference like #539.

Regards
Marco

rahulchavan30:
hi macro , 
Thanks for the update ... if i understand clearly i need to do the following  ,

1. add a config parameter showOriginalStackTrace .
2. by default showOriginalStackTrace will be false and when its false i need to return 
onDone(lastError || timedOutError, null)  if its true i need to return  
let error = timedOutError
  if (lastError) {
    error = lastError
    const userStackTrace = timedOutError.stack.split('\n').slice(1).join('\n')
    error.stack = `${error.stack.split('\n')[0]}\n${userStackTrace}`
  }
  onDone(error, null); 
i will mae these  changes and update the PR if you agree ..

marcosvega91:
Hi @rahulchavan30 You're welcome 😸 
Yes that's right. 
You can put only this 

```
const userStackTrace = timedOutError.stack.split('\n').slice(1).join('\n')
error.stack = ${error.stack.split('\n')[0]}\n${userStackTrace}
```

Around  the new if. I think it's more readable

rahulchavan30:
Thanks macro will do the changes and submit ...

rahulchavan30:
Hi Macro , please bear with me , i have made the necessary changes but when i run npm test update  
i get an error 

i have changed the code to the following 

function onTimeout() {
      let error = timedOutError
      let showOriginalStackTrace = getConfig().showOriginalStackTrace
      if (lastError) {
        error = lastError
        if(showOriginalStackTrace){
          const userStackTrace = timedOutError.stack.split('\n').slice(1).join('\n')
          error.stack = `${error.stack.split('\n')[0]}\n${userStackTrace}`
        }
      }
      onDone(error, null)
    }

but i get this code coverage error  wait-for.js                       |   94.74 |    94.12 |     100 |   94.29 | 71-72 
can you help me out with the test i need to satisfy the code coverage ?

rahulchavan30:
Quick update i added a new test case and was able to get the code coverage passing

src/__tests__/wait-for.js
```diff
@@ -34,6 +34,17 @@ test('uses generic error if there was no last error', async () => {
...
 
+test('uses full stack error trace when showOriginalStackTrace present', async () => {
+  const error = new Error('Throws the full stack trace')
+  const result = await waitFor(
+    () => {
+      throw error
+    },
+    {timeout: 8, interval: 5, showOriginalStackTrace: true},
+  ).catch(e => e)
+  expect(result).toBe(error)
```
kentcdodds, line 45: It's possible that the `error.stack` could've been manipulated (if we decide to change the implementation in the future). So I'd change this:

```suggestion
  expect(result.stack).toBe(error.stack)
```

src/__tests__/wait-for.js
```diff
@@ -34,6 +34,17 @@ test('uses generic error if there was no last error', async () => {
...
+  const error = new Error('Throws the full stack trace')
+  const result = await waitFor(
+    () => {
+      throw error
+    },
+    {timeout: 8, interval: 5, showOriginalStackTrace: true},
+  ).catch(e => e)
+  expect(result.stack).toBe(error.stack)
+})
+
```
kentcdodds, line 47: Could we add one test to verify that the stack trace is set correctly?

This might be tricky though. If you struggle with it, go ahead and push what you have and we can help :)

eps1lon: This is especially interesting for different node versions in case node changes how they display stack traces.

rahulchavan30: i am not knowing how to proceed .. any direction of help will be appreciated .. from what i understood you want me to add a new test that verifies the stack ?

eps1lon: I think a test that shows the actual trace would be nice. You'd probably have to strip the parts of the filepaths that are tied to your local filesystem. You also want to only match the last significant stacks. Otherwise we're probably matching node internals.

rahulchavan30: i am sorry i tried alot but i am not sure how to proceed .

kentcdodds: No worries @rahulchavan30! We can help with this part. Thanks for all you've done :)

rahulchavan30: Thank you Kent and thank you so much for the opportunity I really got to learn so many things in short time of span

kentcdodds: That's wonderful to hear :) Take a look at my changes. It ended up being a bit challenging for me as well, but I'm happy with what we're going to ship together :)

kentcdodds:
@all-contributors please add @rahulchavan30 for code and tests

PR #550, opened 2020-05-06
 fix: Exclude container in the list of queryable roles

### LINKED ISSUE #549: findByRole - document ignored
image

Unable to find role "document" and then logs out a list of available roles and clearly shows there is "document".
What gives?

Thank you

eps1lon:
To better understand, review commits individually. The first commit describes the current behavior and the second commit applies the expected behavior.

**What**:
- Closes #549
- Exclude `container` when displaying the available roles for a bound `ByRole`

**Why**:

- The `container` was never part of the considered elements due to how `.querySelectorAll` works.

**How**:

Start with `container.children`

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -321,6 +321,34 @@ Here are the accessible roles:
 `)
 })
 
+test('does not include the container in the queryable roles', () => {
+  const {getByRole} = render(`<li />`, {
+    container: document.createElement('ul'),
+  })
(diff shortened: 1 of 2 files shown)


kentcdodds:
Makes total sense :+1:

Would this also mean that (when using `screen`) `document` wouldn't show up as queryable? I think that would be an improvement because it shows up in the log as queryable but it doesn't make sense to query it.

kentcdodds:
Ah, just saw the issue that inspired this PR. Perfect. Thanks!

PR #568, opened 2020-05-12
fix: Consider explicit role when pretty printing roles

### LINKED ISSUE #553: getByRole's error message is incorrect on button with role "tab"
- `DOM Testing Library` version: 5.5.0
- `node` version: 13.12.0
- `npm` (or `yarn`) version: 1.22.4

### Relevant code or config:

```js
import React from 'react';
import { render, screen } from '@testing-library/react';

const Button = () => <button role="tab" aria-label="my-button" />;

test('button exists', () => {
  render(<Button />);

  screen.getByRole('button', { name: 'my-button' });
});
```

### What happened:

The test (correctly) failed because it could not find an element with the role of button (the button has a role of "tab"). However, when the error message listed the accessible roles, it listed the button's role under "button":

    TestingLibraryElementError: Unable to find an accessible element with the role "button" and name "my-button"

      document:

      Name "":
      <body />

      button:

      Name "my-button":
      <button
        aria-label="my-button"
        role="tab"
      />

### Reproduction:

[Code Sandbox](https://codesandbox.io/s/react-testing-library-demo-pvsse?file=/src/__tests__/button.js)

### Problem description:

The current behaviour is confusing because the error message says there is no element with the role of "button" and then shows that there is an element with the role of "button".

### Suggested solution:

Buttons with other roles (e.g. [tab](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role)) should be shown under their correct roles in the error message.

I'm happy to put up a PR for this but would need pointing in the right direction as I'm unfamiliar with the code 😁

eps1lon:
**What**:

Consider explicit roles first when pretty printing roles.

**Why**:

`getRoles` printed a different role than what `queryAllByRole` could query. Closes #553

**How**:

Check for existence of `role` attribute first and use the first role from that list. I haven't put much thought into fallback roles but we can discuss it when it becomes an issue.

Technically this isn't correct. html-aria disallows (some) roles for certain HTML elements. But we didn't consider this for `queryAllByRole` either and it isn't clear if we should restrict it (e.g. chrome's a11y pane does not).

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/role-helpers.js
+++ b/src/role-helpers.js
@@ -134,7 +134,13 @@ function getRoles(container, {hidden = false} = {}) {
       return hidden === false ? isInaccessible(element) === false : true
     })
     .reduce((acc, node) => {
-      const roles = getImplicitAriaRoles(node)
+      let roles = []
+      // TODO: This violates html-aria which does not allow any role on every element
+      if (node.hasAttribute('role')) {
+        roles = node.getAttribute('role').split(' ').slice(0, 1)
+      } else {
+        roles = getImplicitAriaRoles(node)
+      }
 
       return roles.reduce(
         (rolesAcc, role) =>
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -343,6 +343,35 @@ Here are the accessible roles:
 `)
 })
 
+test('explicit role is most specific', () => {
+  const {getByRole} = renderIntoDocument(
+    `<button role="tab" aria-label="my-tab" />`,
+  )
+
+  expect(() => getByRole('button')).toThrowErrorMatchingInlineSnapshot(`
+"Unable to find an accessible element with the role "button"
+
+Here are the accessible roles:
+
+  tab:
+
+  Name "my-tab":
+  <button
+    aria-label="my-tab"
+    role="tab"
+  />
+
+  --------------------------------------------------
+
+<body>
+  <button
+    aria-label="my-tab"
+    role="tab"
+  />
+</body>"
+`)
+})
+
 describe('configuration', () => {
   let originalConfig
   beforeEach(() => {


src/role-helpers.js
```diff
@@ -134,7 +134,13 @@ function getRoles(container, {hidden = false} = {}) {
       return hidden === false ? isInaccessible(element) === false : true
     })
     .reduce((acc, node) => {
-      const roles = getImplicitAriaRoles(node)
+      let roles = []
+      // TODO: This violates html-aria which does not allow any role on every element
+      if (node.hasAttribute('role')) {
+        roles = node.getAttribute('role').split(' ').slice(0, 1)
```
kentcdodds, line 140: Can an element have multiple roles? If so, is there a reason we're only getting the first one here?

src/role-helpers.js
```diff
@@ -134,7 +134,13 @@ function getRoles(container, {hidden = false} = {}) {
       return hidden === false ? isInaccessible(element) === false : true
     })
     .reduce((acc, node) => {
-      const roles = getImplicitAriaRoles(node)
+      let roles = []
+      // TODO: This violates html-aria which does not allow any role on every element
```
kentcdodds, line 138: I'd love to deal with this TODO before merging if it's not a huge effort.

kentcdodds: Oh, just noticed what you said about this. Maybe rather than `TODO` this should link to some explanation for why we're doing this.

src/role-helpers.js
eps1lon: An element can only have a single role. We only consider arrays right now for historic reasons. At some point aria-query will return a single role but this requires more work that isn't actually blocking this PR. 

If you pass in multiple roles then those are considered fallback roles. We allow querying them but haven't decided yet if we want to include it in the error message. It gets quite complicate since you have to branch one more time. 

I don't think this is a concern of this PR. aria-query returning multiple roles or `queryFallback` not being considered can be solved independently.

src/role-helpers.js
eps1lon: It is a TODO though. The way we currently handle it is wrong. But this can be fixed in a separate PR so that we're able to revert if multiple roles are required still (also better for changelog).

Issue #563, opened 2020-05-12
Upgrading pretty-format bumped the minimum Node version to 10.14.2


exogen:
👋 Hello! I noticed that in between v7.2.2 and v7.3.0, there was a major upgrade to `pretty-format`. Unfortunately, this changes the minimum Node version, so it's possible to minor bump `@testing-library/dom` and not work with the same Node anymore.

- `@testing-library/dom` version: v7.4.0
- Testing Framework and version: jest@24.9.0
- DOM Environment: jsdom@11.12.0

### What you did:

```console
$ yarn
```

### What happened:

```console
$ yarn
error pretty-format@26.0.1: The engine "node" is incompatible with this module. Expected version ">= 10.14.2". Got "10.13.0"
error Found incompatible module.
```
### Problem description:

Due to the `pretty-format` major version bump in #546, the minimum Node version changed to 10.14.2 in between `@testing-library/dom` v7.2.2 and v7.3.0.

Given the value in `engines` and relatively recent #480, it seems intentional to support older Node 10 versions. (For reference, I ran into this because my current client happens to use images containing v10.13.0 for our deploys.)

### Suggested solution:

Downgrade `pretty-format` back to v25, or update this package's stated support.

kentcdodds:
I think we'll probably need to downgrade as suggested and then we can upgrade in the next major version bump.

A PR is welcome.

exogen:
Oops, forgot to mark Fixed in #571. 🎉

PR #571, opened 2020-05-13
chore: revert pretty-format dependency upgrade

exogen:
**What**: Downgrade `pretty-format` to v25.

**Why**: The newer v26 bumps the minimum required version of Node. See #563.

**How**: Update version in package.json

**Checklist**:

- [ ] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs) – N/A
- [ ] I've prepared a PR for types targeting [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__dom) – N/A
- [x] Tests – no new tests, but verified that they pass
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "@babel/runtime": "^7.9.6",
     "aria-query": "^4.0.2",
     "dom-accessibility-api": "^0.4.3",
-    "pretty-format": "^26.0.1"
+    "pretty-format": "^25.0.1"
   },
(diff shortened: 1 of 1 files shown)


exogen:
@kentcdodds If you still want to bump it vs. what was there before, I'm happy to update the range to `^25.5.0` which presumably will still keep the compatible Node version.

PR #585, opened 2020-05-24
Support `clipboardData` like `dataTransfer`.

calebmer:
To support the [paste event](https://developer.mozilla.org/en-US/docs/Web/API/ClipboardEvent) you need the `clipboardData` key (which jsdom ignores). This commit extends how DOM testing library supports `dataTransfer` for the [drop event](https://developer.mozilla.org/en-US/docs/Web/API/DragEvent) to `clipboardData`.

This allows us to write tests for pasting.

**Checklist**:

- [ ] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs) (n/a)
- [ ] I've prepared a PR for types targeting [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__dom) (n/a)
- [ ] Tests (edit made via GitHub, let’s see what CI says)
- [x] Ready to be merged


nickmccurdy:
I've added a basic test asserting that a POJO similar to `ClipboardData` can be passed to the `clipboardData` property of the `ClipboardEvent`.

Eventually as we add paste support to `userEvent` and merge that into this package, we can use more realistic mocks of the DOM APIs, I just didn't think they were necessary for this PR alone since it's a simple value assignment. Also if you're interested, I made a [Paste event playboard](https://codesandbox.io/s/paste-event-playground-6zlo9) to learn the `ClipboardData` API and make sure the test is accurate.

kentcdodds:
@all-contributors please add @calebmer for code

kentcdodds:
@all-contributors please add @nickmccurdy for tests

PR #586, opened 2020-05-25
feat: suggestions for which query to use

benmonro:
**What**:
Adding suggestions for which query to use. (opt in via configure).  When using `configure({throwSuggestions: true})` tests will fail if they are not using the suggested queries.  While this may seem harsh, the intent is more along the lines of the way that `--detectLeaks` works in jest.  

**Why**:
To encourage people to follow best practices when using TL.

**How**:
a new getSuggestedQuery function that follows the order of 'which query should i use?'

Started going down the road of having a separate library but it became apparent pretty quick that it made more sense for this to live in DTL directly.  We can export the suggestion from here for things like testing-playground.com

Docs PR: https://github.com/testing-library/testing-library-docs/pull/477/files

**Checklist**:

- [x] Support all use cases of 'which query should i use'
- [x] Tests
- [x] Ready to be merged


src/query-helpers.js
```diff
@@ -64,6 +74,19 @@ function makeGetAllQuery(allQuery, getMissingError) {
...
+      const rawSuggestions = els.map(element => getSuggestedQuery({element}))
+      const suggestions = [
+        ...new Set(els.map(element => getSuggestedQuery({element}).toString())),
+      ]
+
+      if (
+        suggestions.length === 1 &&
+        !allQuery.name.endsWith(rawSuggestions[0].queryName)
+      ) {
+        throw getSuggestionError(suggestions[0], container)
```
benmonro, line 87: @kentcdodds  was pairing on this with my friend Kevin and he asked a good question.  Should the suggestion Try to actually execute the query it's suggesting just to make sure it works?  Curious to hear your thoughts.
..

src/queries/placeholder-text.js
```diff
@@ -1,7 +1,8 @@
 import {queryAllByAttribute, buildQueries} from './all-utils'
 
-const queryAllByPlaceholderText = queryAllByAttribute.bind(null, 'placeholder')
-
+function queryAllByPlaceholderText(...args) {
+  return queryAllByAttribute('placeholder', ...args)
+}
```
benmonro, line 5: ICYW: all of the other queries use named functions and the suggestion feature needs them to be named.

src/queries/test-id.js
```diff
@@ -2,8 +2,9 @@ import {queryAllByAttribute, getConfig, buildQueries} from './all-utils'
 
 const getTestIdAttribute = () => getConfig().testIdAttribute
 
-const queryAllByTestId = (...args) =>
-  queryAllByAttribute(getTestIdAttribute(), ...args)
+function queryAllByTestId(...args) {
+  return queryAllByAttribute(getTestIdAttribute(), ...args)
+}
```
benmonro, line 7: ICYW: all of the other queries use named functions and the suggestion feature needs them to be named.

src/query-helpers.js
```diff
@@ -49,10 +50,27 @@ function makeSingleQuery(allQuery, getMultipleError) {
...
-    return els[0] || null
+    const element = els[0] || null
+
+    if (getConfig().showSuggestions) {
+      const suggestion = getSuggestedQuery(element)
+      if (suggestion && !allQuery.name.endsWith(suggestion.queryName)) {
+        throw getSuggestionError(suggestion.toString(), container)
+      }
+    }
+    return element
```
smeijer, line 61: Copy from [Twitter](https://twitter.com/meijer_s/status/1265217798298578944?s=20):

I think it's awesome to have suggestions embedded within this project. But as expressed before, I'm still concerned about the developer experience.

It's currently throwing errors when there are slightly better queries available. I'm afraid this would be the first thing I would disable.

Perhaps it's better to log warnings instead? That's also what the config name implies `showSuggestions`.

If we're throwing, a config property like `strict` would be a better match IMO.

benmonro: @smeijer yeah the issue is pointing people exactly to the line we are suggesting them to change.  I'm open to other ideas but throwing an error in jest will take them directly to where the suggestion is pointing. Think of this In the same way that detect leaks works in jest. It only breaks tests when you turn it on.  That's why this is opt-in and disabled by default.

benmonro: I'm good with renaming the config property but `strict` feels a bit too vague to me. @kentcdodds  any thoughts here?

eps1lon: How would the stacks look with `console.warn`?

benmonro: > It's currently throwing errors when there are slightly better queries available.

Actually on many teams people go straight to getByTestId and I would argue it's not just slightly better but WAY better to use getByRole so for me I actually want this to throw in those cases

benmonro: I can try it out @eps1lon  and show the differences here

smeijer:
Copy from [Twitter](https://twitter.com/meijer_s/status/1265285988521324550?s=20) again :innocent: 

@benmonro
 We've mentioned something like "screen.getByLabelText('foo', { silent: true })" and agreed it's not the best solution.  

How do you feel about using an "instance" level for it instead? `screen.showSuggestions(true|false)`, where you can run queries in between:

```js
screen.showSuggestions(false);

document.querySelector('input'); // don't throw this time

screen.showSuggestions(true);
```

Or an alternative, a wrapping method like `within`

Name it `strict`?

```js
const { getByText } = strict(document.getElementById('messages'));
const helloMessage = getByText('hello')
```

This could then be accompanied by a `screen.strict = true`, which enables strict mode for all `screen.x` queries.

benmonro:
@kentcdodds @eps1lon @smeijer I tried using warn (screenshot below).  I'm definitely not a fan for 2 reasons:

1) you don't really know which test it's in.  Failing the test will make it very obvious which test is using the query we're suggesting for improvement.
2) While the difference between `getByLabel` and `getByRole` might be slight, my biggest goal with this is to get people to stop using `getByTestId` which in my view does in fact warrant a fail when there's something better out there.

image

smeijer:
> While the difference between getByLabel and getByRole might be slight, my biggest goal with this is to get people to stop using getByTestId which in my view does in fact warrant a fail when there's something better out there.

Do I understand correctly that you only throw for `getByTestId` queries? Because if throwing only happens for `querySelector` and `getByTestId` when the dev chose to `opt-in`, then I (as a developer) would have zero issues with throwing.

Heck, I can imagine something like that being default in a future version.

benmonro:
>Do I understand correctly that you only throw for getByTestId queries? 

no, I'm not currently having this warn on usages of querySelector, that would be a bit sketchy imo as we'd have to change the way querySelector works. Also I feel like that one *can* be done w/ a lint rule.  As for getByTestId, no I do not only throw on that.  It throws whenever a better option is available.


If we use warn we'd have to parse the stack trace (not sure when I'd have cycles to do that) have a look at my screenshot above (or feel free to pull my branch and play with it).  If you can get the stack trace parsing working I'd be fine w/ having it use a warn (or maybe make that optional as well?).  We can do `new Error().stack` but it would have to ultimately highlight the exact line in their test where the query was called or it's better to just fail the test imo.


I'll lay out the options here (in no particular order) for how to disable the suggestion for a single query, feel free to add more options, but maybe we can take a vote?

option 1: an additional query option in the query itself:
```js
screen.getByTestId("foo", {suggest: false})
```

option 2: just use configure:
```js
configure({showSuggestions:false});
screen.getByTestId("foo")
configure({showSuggestions:true});
```

option 3: Stephan's `screen.showSuggestions` idea
```js
screen.showSuggestions(false);

document.querySelector('input'); // don't throw this time

screen.showSuggestions(true);
```

option 4: more granular configuration option for showSuggestions
```js
configure({
  showSuggestions: { ByLabelText: "warn", ByTestId: "error" }
})
```

smeijer:
Option 5

```js
// strict || strictA11y || ...

const { getByText } = strict(document.getElementById('messages'));
const helloMessage = getByText('hello')
```

A method like that fits nicely together with `within` in terms of code style.

Talking about `within`. If I'm not mistaken, that method also modifies the error stack to print the correct code frame. That's why we have the option `showOriginalStackTrace`.

I do agree, that is a must for this to work nicely. Showing incorrect code frames is nothing more than confusing.

benmonro:
@smeijer ah ok i'll look into how that prints the stack trace.

w/ regards to using `strict(document.querySelector|getElementById)` not sure I follow the point there... imo if you are using any of the primitive options to query the DOM that can be handled by a lint rule. can you elaborate on how you see that working?  it's not clear to me...

smeijer:
Sure let me try to explain my train of thoughts a bit better.

The recommended approach to query a document is using screen:

```js
import { screen } from '@testing-library/dom';
screen.getByText('hello');
```

But, we can also use queries directly:

```js
import { getByText } from '@testing-library/dom';
getByText(document, 'hello');
```

To limit the scope of a query method, we have the `within` helper:

```js
import { within } from '@testing-library/dom';
const view = within(document.getElementById('messages'));
view.getByText('hello');
```

What I'm proposing is, is to add a `strict`, that enables `suggest` on that scope.

```js
import { screen, strict } from '@testing-library/dom';
const view = strict(document.getElementById('messages'));

view.getByText('hello'); // might throw for a11y
screen.getByText('hello'); // won't throw for a11y
```

This `strict` mode, could be implemented next to an option flag (option 1)

```js
import { screen } from '@testing-library/dom';
screen.getByText('hello', { strict: true });
```

Why I'm comparing it to `within`, is because as you can see above, we'll have very similar API's this way.

So, My vote is for an API that matches the last 2 (both of them) code blocks in this comment :innocent: 

_ps. I'm deliberately naming it `strict` and not `suggest`. As throwing errors isn't suggesting. It's enforcing._

src/suggestions.js
```diff
@@ -0,0 +1,83 @@
...
+  if (!label) {
+    const ariaLabelledBy = element.getAttribute('aria-labelledby')
+
+    if (ariaLabelledBy) {
+      // we're using this notation because with the # selector we would have to escape special characters e.g. user.name
+      // see https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector#Escaping_special_characters
+      label = document.querySelector(`[id=${ariaLabelledBy}]`)
+    }
+  }
+
```
kentcdodds, line 23: I *think* we could do this instead:

```suggestion
  const label = element.labels.find(label => Boolean(normalize(label.textContent)))
```

src/suggestions.js
```diff
@@ -0,0 +1,83 @@
...
+  return undefined
+}
+
+function makeSuggestion(queryName, content, name) {
+  return {
+    queryName,
+    toString() {
+      const options = name ? `, {name:/${name}/}` : ''
```
kentcdodds, line 35:
```suggestion
      const options = name ? `, {name: /${name.toLowerCase()}/i}` : ''
```

src/suggestions.js
```diff
@@ -0,0 +1,83 @@
...
+
+export function getSuggestedQuery(element) {
+  const roles = getRoles(element)
+
+  const roleNames = Object.keys(roles)
+  const name = computeAccessibleName(element)
+  if (roleNames.length) {
+    const [role] = roleNames
+    return makeSuggestion('Role', role, name)
+  }
```
kentcdodds, line 49: No sense in computing the name if it's not used.

```suggestion
  if (roleNames.length) {
    const [role] = roleNames
    return makeSuggestion('Role', role, computeAccessibleName(element))
  }
```

src/suggestions.js
```diff
@@ -0,0 +1,83 @@
...
+    return makeSuggestion('LabelText', labelText)
+  }
+
+  const placeholderText = element.getAttribute('placeholder')
+  if (placeholderText) {
+    return makeSuggestion('PlaceholderText', placeholderText)
+  }
+
+  let {textContent} = element
+  textContent = normalize(textContent)
```
kentcdodds, line 62:
```suggestion
  const textContent = normalize(element.textContent)
```

kentcdodds:
> you don't really know which test it's in.

Actually, when someone's using this in their own project, Jest will show a call stack and a code frame on the thing that triggered the warning... That is... except for async queries 😬 Yeah, maybe throwing is better.

We do still need some mechanism to disable the suggestion for a specific query. Of the options mentioned, I prefer `screen.getByTestId("foo", {suggest: false})`. Seems least indirect. I also like the idea of making it an option for `within`. Hopefully that doesn't complicate things too much 😬

src/__tests__/suggestions.js
```diff
@@ -0,0 +1,175 @@
...
+
+  expect(() => screen.getAllByTestId('foo')).not.toThrowError()
+})
+
+test('should suggest getByRole when used with getBy', () => {
+  renderIntoDocument(`<button data-testid="foo">submit</button>`)
+
+  expect(() => screen.getByTestId('foo')).toThrowErrorMatchingInlineSnapshot(`
+"A better query is available, try this:
+*ByRole("button", {name:/submit/})
```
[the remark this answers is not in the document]
benmonro: @kentcdodds I wanted to do that, but it would require having the make*Query functions pass down the first bit of the query.  if you want i can try that, but it was gonna touch a lot of files so I held off...

src/__tests__/suggestions.js
```diff
@@ -0,0 +1,175 @@
...
+    <title data-testid="svg">Close</title>
+    <g><path /></g>
+  </svg>`)
+
+  expect(() => screen.getByTestId('delete')).toThrowError(
+    /\*ByTitle\("Delete"\)/,
+  )
+
+  //Since `ByTitle` and `ByText` will both return the <title> element
+  //`getByText` will always be the suggested query as it is higher up the list.
```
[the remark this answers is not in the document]
benmonro: hahaha 🤣

src/__tests__/suggestions.js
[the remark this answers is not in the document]
benmonro: The reason it would touch a lot of files is because they all end up going through the `queryAllBy` so afaict it would mean having to pass `get|find|query` w/ each make*Query call.

src/query-helpers.js
```diff
@@ -64,6 +82,21 @@ function makeGetAllQuery(allQuery, getMissingError) {
         container,
       )
     }
+    if (getConfig().showSuggestions) {
+      //get a unique list of all suggestion messages.  We are only going to make a suggestion if
```
benmonro, line 86: @kentcdodds you missed one. 🤣 
```suggestion
      // get a unique list of all suggestion messages.  We are only going to make a suggestion if
```

src/suggestions.js
benmonro: oh neat, is that like the inverse of label.control?  I was wondering if something like that exists.  i'll try this one after the kids go to bed. :)

src/suggestions.js
benmonro: @kentcdodds was talking about that with @appleJax (we were pairing on zoom for this - and you should totally @allcontributors him 😄)

wouldn't it be better to be exact on this?  like, you have the exact text of the element?  i'm on the fence would like to hear your thoughts though...

src/suggestions.js
benmonro: ah yeah good catch was playing w/ using that in the ByText but ended up not doing that...

src/query-helpers.js
benmonro: @eps1lon unfortunately warn didn't work :(

src/suggestions.js
smeijer: Why would you want your test to break, due to simple changes in capitalization?

benmonro: @smeijer I'm kinda on the fence, I see both arguments.  On the one hand I agree that maybe a change in capitalization shouldn't break the test, on the other, you could argue that a change in capitalization _is_ a change in content and it should break the test.  But more specifically in this case we have the actual content of the element in hand and we can give an exact value that will not break the test.

smeijer: > But more specifically in this case we have the actual content of the element in hand and we can give an exact value that will not break the test.

That part is for future changes of course 😁 now it will work. Than one day some one will come, and decides that the labels should all be capitalized.

This won't affect how the inputs work, or how your form will be submitted. It's a visual change only. We don't want tests to break in visual changes. For that, other test methods exist.

benmonro: agree to disagree, I think some folks would call it a content change, something that the user sees.

smeijer:
I like how you've put this thing together! Way cleaner than what I could have come up with. :+1: :heart_eyes:

benmonro:
Thanks @smeijer looking forward to seeing it in testing playground and the chrome extension. 😎

src/__tests__/suggestions.js
```diff
@@ -0,0 +1,201 @@
+import {configure} from '../config'
+import {screen} from '..'
+import {renderIntoDocument} from './helpers/test-utils'
+
+beforeAll(() => {
+  configure({showSuggestions: true})
+})
+
+afterAll(() => {
+  configure({showSuggestions: false})
```
kentcdodds, line 10: "showSuggestions" is probably not the best config option since it throws. Maybe "throwSuggestions" or "failOnSuboptimalQueries" or something 🤔

src/query-helpers.js
```diff
@@ -75,18 +86,79 @@ function makeFindQuery(getter) {
...
 }
 
+const wrapSingleQueryWithSuggestion = (query, queryAllByName, variant) => (
+  container,
+  ...args
+) => {
+  const element = query(container, ...args)
+  const [{suggest = true} = {}] = args.slice(-1)
+
+  if (getConfig().showSuggestions && suggest) {
```
kentcdodds, line 96: I think that `suggest` should override `showSuggestions`. So maybe it should be:

```suggestion
  const [{suggest = getConfig().showSuggestions} = {}] = args.slice(-1)

  if (suggest) {
```

src/query-helpers.js
```diff
@@ -75,18 +86,79 @@ function makeFindQuery(getter) {
...
+}
+
+const wrapAllByQueryWithSuggestion = (query, queryAllByName, variant) => (
+  container,
+  ...args
+) => {
+  const els = query(container, ...args)
+
+  const [{suggest = true}] = args.slice(-1)
+  if (suggest && getConfig().showSuggestions) {
```
kentcdodds, line 113:
```suggestion
  const [{suggest = getConfig().showSuggestions} = {}] = args.slice(-1)

  if (suggest) {
```

src/suggestions.js
kentcdodds: I see the capitalization like styling. It doesn't matter if the button is blue or green, I'm gonna click it if it says: "Win a million dollars" 😉

I won't care if that's in all caps or all lower case either: "WIN A MILLION DOLLARS" or "win a million DOlLaRs". I'm clicking that either way 😜

benmonro: haha. You had me at win a million dollars. Hold off on merging I'm gonna push the changes you suggested...

src/__tests__/suggestions.js
benmonro: i can go w/ throwSuggestions. :)

src/query-helpers.js
benmonro: done

src/query-helpers.js
benmonro: done

src/__tests__/suggestions.js
```diff
@@ -0,0 +1,175 @@
...
+
+test('should suggest *ByRole when used with getAllBy', () => {
+  renderIntoDocument(`
+    <button data-testid="foo">submit</button>
+    <button data-testid="foo">submit</button>`)
+
+  expect(() => screen.getAllByTestId('foo'))
+    .toThrowErrorMatchingInlineSnapshot(`
+"A better query is available, try this:
+*ByRole("button", {name:/submit/})
```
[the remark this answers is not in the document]
benmonro: @kentcdodds got this working, had to move the suggestions after the queries and created a couple wrappers for this so they could be used in each file, wasn't too bad, have a look and let m know if it's all Kosher.

src/suggestions.js
benmonro: @kentcdodds was able to use that for the first bit, but for ariaLabelledBy it didn't work so I had to keep that part.

benmonro:
@kentcdodds @smeijer all the feedback has been addressed.  docs PR is good to merge as well as soon as this merges.

src/__tests__/suggestions.js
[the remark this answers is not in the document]
benmonro: @kentcdodds,  got this working last night.

src/suggestions.js
benmonro: all set ready to go

kentcdodds:
Yes 👍

benmonro:
@timdeschryver @kentcdodds added the type to matcher types.

types/matches.d.ts
```diff
@@ -1,29 +1,33 @@
...
-    /** Use normalizer with getDefaultNormalizer instead */
-    collapseWhitespace?: boolean;
-    normalizer?: NormalizerFn;
+  exact?: boolean
+  /** Use normalizer with getDefaultNormalizer instead */
+  trim?: boolean
+  /** Use normalizer with getDefaultNormalizer instead */
+  collapseWhitespace?: boolean
+  normalizer?: NormalizerFn
+  /** set query suppress suggestions */
```
kentcdodds, line 13: This comment doesn't make sense to me 😅

benmonro: lol sorry i rushed it you're right haha
```suggestion
  /** suppress suggestions for a specific query */
```

benmonro:
@allcontributors please add @appleJax for code & tests

benmonro:
@kentcdodds can you `@allcontributors` @appleJax he and I were pair programming for a lot of this and a lot of the ideas came from him even though i was typing the `()`s and the `[]`s.  :)

benmonro:
@allcontributors please add @benmonro for code, ideas, tests, docs

smeijer:
:tada:

smeijer:
I don't think that this pull-request would have been here if it wasn't for http://testing-playground.com, so how about this one as well?

@allcontributors please add @smeijer for ideas

benmonro:
@smeijer  yep yep

PR #588, opened 2020-05-27
Fix misleading advice in waitForElement deprecation warning

pelotom:
Fixes https://github.com/testing-library/react-testing-library/issues/677


kentcdodds:
@all-contributors please add @pelotom for code and tests

PR #598, opened 2020-05-31
fix: update config type

timdeschryver:
**What**: We forgot to add the `throwSuggestions` option in the global config

**Why**: To have intellisense. In TS projects we also currently get an error while trying to do the following.

```ts
configure({ throwSuggestions: true });
```

**How**:

**Checklist**:

- [ ] Tests
- [x] Ready to be merged
      

Should we try to keep the naming consistent? 
For the global config it's `throwSuggestions` , while the config on a query basis is `suggest`.

```
configure({throwSuggestions: true})
// vs
screen.getByText('Increment', {suggest: true})
```

### DIFF
--- a/types/config.d.ts
+++ b/types/config.d.ts
@@ -3,6 +3,7 @@ export interface Config {
     asyncWrapper(cb: (...args: any[]) => any): Promise<any>;
     asyncUtilTimeout: number;
     defaultHidden: boolean;
+    throwSuggestions: boolean;
 }
 
 export interface ConfigFn {


kentcdodds:
Yes, these should be consistent. That was an oversight. I suggest we add the option and deprecate the old one.

benmonro:
>Should we try to keep the naming consistent?
For the global config it's throwSuggestions , while the config on a query basis is suggest.

@timdeschryver actually I think it's ok to keep the naming different.  Actually I'm hoping to add a way to choose between doing a warn and a throw, but in either case you'd still want to turn it off for a specific query so I think it's fine to keep them slightly different.

also thanks for putting this PR in, thought i got all the typings, but must have missed it.

PR #599, opened 2020-05-31
fix: more improvements to suggestions

benmonro:
**What**: 

* Fixed bug that would suggest a query that would yield a different element than the one being verified.  Also added additional test for label > title.
* Fixed bug that would suggest get by role w/ name that had un-escaped regexp that would not yield the element intended.

**Why**: suggestions should give the same results, just using better queries.

**How**: 
* use specific || implied roles.
* escape regexp

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/suggestions.js
+++ b/src/__tests__/suggestions.js
@@ -35,6 +35,18 @@ test('respects ignores', () => {
   ).not.toThrowError()
 })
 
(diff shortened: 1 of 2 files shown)


benmonro:
Thanks for letting me iterate @kentcdodds. 😁 As I build more features into the extension and find issues, I'm fixing as I go.

Issue #537, opened 2020-04-28
querySelectorAll is not a function when being called by getByTestId


Nubzor:
In the beginning, I would like to thank you for the time you folks put to develop and maintain the library. 

I face an issue related to missing .querySelectorAll function while calling getByTestId method which is odd because when I check the object property then I can see that the function exists :shrug:

Code example: 

```import { render, getByTestId } from "@testing-library/svelte";
import Loader from "./Loader.svelte";

describe("Loader component", () => {
  test("should render component correctly", () => {
    const { container } = render(Loader);

    console.log(typeof container.querySelectorAll)

    expect(container).toContainElement(getByTestId("spinner"));
    expect(container).toBeVisible();
  });
});
```

Output: 

``` FAIL  src/Loader/Loader.spec.js
  Loader component
    × should render component correctly (22ms)

  ● Loader component › should render component correctly

    TypeError: container.querySelectorAll is not a function

> 10 |     expect(container).toContainElement(getByTestId('spinner'));

-------

Test Suites: 2 failed, 2 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        3.162s
Ran all test suites matching /src/i.
  console.log     
    Log:  function

```

There is a list of dependencies and their versions: 

```
    ...
    "@testing-library/jest-dom": "^5.5.0",
    "@testing-library/svelte": "^3.0.0",
    ...
    "jest": "^25.4.0",
    "jest-transform-svelte": "^2.1.1",
    "svelte": "^3.21.0",
```

Node: 12.16.1 on Windows 10

I am not sure whether this is a bug or I missed something during the configuration, any tips would be appreciated.

Greetings, 

Mateusz

Jnforja:
Hi,

I'd like to work on this issue. If I understood correctly, the goal is to have all of the base queries throw an error when the first argument of a base query isn't a DOM node. Which includes the following queries and its variants:
- ByLabelText
- ByPlaceholderText
- ByText
- ByAltText
- ByTitle
- ByDisplayValue
- ByRole
- ByTestId

Is this what is intended?

kentcdodds:
To be clear, these days we're encouraging people to use `screen` which means that most of the time people won't be passing the `container` argument themselve's (it'll be pre-bound). But having this runtime check definitely wouldn't hurt.

Jnforja:
> Every query is built on top of the `queryAllBy*` variant, so you'd only need to worry about that one for each type :)
> 
> Thanks!

[...]

PR #604, opened 2020-06-04
feat: Add check for container type on queries

Jnforja:
**What**: 
Add check for container type on queries

**Why**: 
So users of this library get an informative warning when
    the first argument of a base-query is not an appropriate container.

**How**: 
1. Created a helper function that does a runtime check on the type of a container.
2. On every queryAllBy* variant called the new helper function with the given container. 

**Checklist**:

- [x] Tests
- [ ] Ready to be merged
      

Some decisions I'd like to point out for review:
- As the base-queries are part of the public API, and to not risk breaking clients, I took a more relaxed approach to type checking containers and only check for the properties querySelector and querySelectorAll, instead of going for something more restrictive like nodeType.
- Although I'm not 100% sure, from the code I read, the types that seem appropriate to be containers are Element, Document, and DocumentFragment. So those were the ones that I referenced in the error message.
-  Initially, I added the helper function to the query-helpers.js file as query related functions are the only ones to use it. But then I remembered that query-helpers.js is part of the public API, so I ended up moving the helper function to helpers.js as it seemed like the best place yet.

Please let me know if there's anything that I should change in this PR and I'll get to it :)

Ref: testing-library/dom-testing-library#537


kentcdodds:
@all-contributors please add @Jnforja for code and tests

Issue #545, opened 2020-05-05
getByLabelText doesn't concat multipart aria-labelledby


sheesania:
- `React Testing Library` version: 9.5.0
- `node` version: 12.4.0
- `npm` (or `yarn`) version: 6.13.6

### What you did:

Along the lines of [this MDN article](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Multipart_labels), I have an input labelled by a combination of three different elements: a label, a text field, and another label. So the input has an `aria-labelledby` attribute with a value like `label1 text-field label2`. In my test, I'm trying to grab the input based on that full, concatenated label, with something like `getByLabelText('text of label1 value of text-field text of label2')`.

### What happened:

I can successfully grab the input with `getByLabelText('text of label1')` or `getByLabelText('text of label2')`, but giving it the full, concatenated label `getByLabelText('text of label1 value of text-field text of label2')` results in nothing being found. Even if I removed the text field ID from `aria-labelledby`, I still can't get the input using `getByLabelText('text of label1 text of label2')`.

### Reproduction:

Here's a codesandbox: https://codesandbox.io/s/rtl-aria-labelledby-multipart-vpjpx?from-embed=&file=/src/__tests__/index.js

### Problem description:

Based on my understanding of [the MDN article](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Multipart_labels), the "label" for an input with a multiple IDs in `aria-labelledby` should be a concatenation of the values of the elements with those IDs. So I expected `getByLabelText` to retrieve the input based on this concatenated label (especially because it can handle `aria-labelledby` attributes with just one ID).

### Suggested solution:

From looking at [label-text.js](https://github.com/testing-library/dom-testing-library/blob/master/src/queries/label-text.js), it seems like there are a couple things going on...

- `getByLabelText` [only looks at `<label>` elements](https://github.com/testing-library/dom-testing-library/blob/9007a673a138a50becd764c58d08c28477bcfc33/src/queries/label-text.js#L19), so that `<input>` that's part of my `aria-labelledby` doesn't get picked up at all. (However, even when I removed the input's ID from `aria-labelledby`, I still had the concatenation problem, so this is kind of a separate issue.)
- `getByLabelText` does a sweep for individual `<label>` elements matching the given text (before trying to figure out what input elements they're labeling). This means it doesn't catch _combinations_ of `<label>` elements (or any other kind of element) that match the given text when concatenated.

[...]

PR #607, opened 2020-06-06
Get element by label input value and by labels concat values

delca85:
**What**:

- `getAllLabelsByText` now retrieves label by `input` elements not empty value too

- `getByLabelText` accepts a new argument named `concat` that enables retrieving elements by multiple concatenated labels

**Why**:
According to #545  retrieving element through `input` labels was impossible as well as using a combination of labels to get a labelled element.

**How**:

1. `queryAllLabelsByText` now makes a query not only on elements of type `label` but on `input` elements too. The empty input are excluded.
2. in order to make `queryAllLabelsByText` gets the labels even if the text to be matched is a concatenation of labels, the `concat` argument is passed to the `matcher`. The `matcher` checks the value of `concat` to decide if the condition to be satisfied is an equality or an includes (in case of `concat` true) condition between the `text` and the candidate label.
3. a last check is performed to know if all the provided `text` is composed by the found labels

**Checklist**:

- [x] add `input` type to elements queried by `queryAllLabelsByText`
- [x] remove `input` elements with empty value
- [x] add `concat` argument to all the methods called in the chain
- [x] change `matcher` behavior according to `concat` value
- [x] check if the labels compose the received `text` to be matched

- [x] Tests
- [x] Ready to be merged
      

Like said in #545 comments, I am not sure this is the right way to fix this wrong behavior and probably I am not considering some use case but I would like to receive some suggestions and share thoughts on it.


kentcdodds:
Hi @delca85! Thank you for this. This is a pretty hard problem. Unfortunately, I'm not sure that this is the right solution, but I don't know what the right solution is 😬 To compound on that, I don't have bandwidth to invest into this myself right now, so I'm calling on the rest of the @testing-library/all-maintainers (and anyone else seeing my bat signal 🦇) to come help us collaborate on a proper solution for this use case.

Sorry I can't be of more help.

bat signal

delca85:
Hi @kentcdodds! It was a pleasure for me to work on this. I would be happy to understand a bit more about the right solution so I hope your call will make someone help us!

kentcdodds:
I think we can close this one now :) Thanks again!

PR #614, opened 2020-06-09
Proposal setup to make the TypeScript migration easier

ghostd:
**What**:

Add a TypeScript config to generate the declaration files from the TS source files.

**Why**:

The aim is to make the TS transition easier (and avoid big PR as mentioned [here](https://github.com/testing-library/dom-testing-library/issues/494#issuecomment-613040674))

**How**:
I added a "manual" script `generate-types` which generates the declaration files from the TS source files. This approach allows us to migrate each JS file one by one. I don't know what could be the best way to insert this command into the build/validation process.

This PR includes a first migration (see `get-node-text.ts`)

**Checklist**:

      [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__dom) "N/A"
- [ ] Tests
- [ ] Ready to be merged
      

Any thoughts are welcome.

#494


ghostd:
I'm not sure to understand why the build fails. The "typecheck" step works for me :-/

babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  presets: ['kcd-scripts/babel', '@babel/preset-typescript'],
```
kentcdodds, line 2: This should not be necessary since there's a `tsconfig.json` to the root of the repo.

https://github.com/kentcdodds/kcd-scripts/blob/50170f0b090593c39f30edc997f62af1c00ecfe1/src/config/babelrc.js#L62

ghostd: I removed it

types/event-map.d.ts
```diff
@@ -0,0 +1,72 @@
+declare type EventMapKey = string
```
kentcdodds, line 1: Are `.d.ts` files necessary when the source is written in TS?

tests/jest.config.dom.js
```diff
@@ -11,4 +11,10 @@ module.exports = {
     '/__node_tests__/',
   ],
   testEnvironment: 'jest-environment-jsdom',
+  transform: {
+    '^.+\\.(js|jsx|ts|tsx)$': path.join(
+      __dirname,
+      '../node_modules/kcd-scripts/dist/config/babel-transform',
+    ),
+  },
```
kentcdodds, line 19: Could you make a PR to kcd-scripts so we don't have to do this?

https://github.com/kentcdodds/kcd-scripts/blob/50170f0b090593c39f30edc997f62af1c00ecfe1/src/config/jest.config.js#L50

tests/jest.config.node.js
```diff
@@ -12,4 +12,10 @@ module.exports = {
     '/__node_tests__/',
   ],
   testMatch: ['**/__node_tests__/**.js'],
+  transform: {
+    '^.+\\.(js|jsx|ts|tsx)$': path.join(
+      __dirname,
+      '../node_modules/kcd-scripts/dist/config/babel-transform',
+    ),
+  },
```
kentcdodds, line 20: Same as above

types/get-node-text.d.ts
```diff
@@ -1 +1,2 @@
-export function getNodeText(node: HTMLElement): string;
+declare function getNodeText(node: HTMLElement): string;
+export { getNodeText };
```
kentcdodds, line 2: Same here. Do we need `.d.ts` files when the source is in TypeScript?

types/event-map.d.ts
smeijer: Shouldn't be

types/events.d.ts
```diff
@@ -1,95 +1,104 @@
 export type EventType =
```
smeijer, line 1: Can't we use the native `HTMLElementEventMap` instead of maintaining a custom type?

types/event-map.d.ts
```diff
@@ -0,0 +1,72 @@
...
+declare type LegacyKeyboardEventInit = {
+  charCode?: number
+} & KeyboardEventInit
+declare type EventMapValue =
+  | {
+      EventType: 'AnimationEvent'
```
smeijer, line 7: Do we want `EventType` here to be a string, or the type it represents from `lib.dom`?

```ts
EventType: 'AnimationEvent' | AnimationEvent
```

types/events.d.ts
eps1lon: Didn't know this existed. I'd prefer this :+1:

types/event-map.d.ts
ghostd: The string is used at runtime:
https://github.com/testing-library/dom-testing-library/blob/281eb8b62700dec9a79ad315b6055a61f61f3c9f/src/events.js#L51-L58

So we need a TS macro to transform a TS interface name into string. I don't know how to do that.

types/events.d.ts
smeijer: Yeah, `add.addEventListener` is typed, so my first thought is that we should be able to reuse those types.

```ts
// HTMLDocument
addEventListener<K extends keyof DocumentEventMap>(type: K, listener: (this: Document, ev: DocumentEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
addEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions): void;

// HTMLElement
addEventListener<K extends keyof HTMLElementEventMap>(type: K, listener: (this: HTMLElement, ev: HTMLElementEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
addEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions): void;
```

@ghostd, please correct me if I'm wrong. I might miss something fundamental.

ghostd: From my understanding, removing the custom `EventType` will not prevent us to maintain the event map (since the metadata are 'erased' at runtime), and for now, the lib.dom.ts is still missing some events ([composition ones](https://github.com/microsoft/TypeScript/issues/35634#issuecomment-636840255) and maybe others).
Maybe we can have a custom type which would be a union of the `DocumentEventMap`, `HTMLElementEventMap` and the missing events used by `dom-testing-library`. In this way we will just maintain the missing types. I'll give a shot

ghostd: I think i misunderstood your point.

We can simply define `EventType` this way:
```
type EventType = keyof (typeof eventMap);
```
When we add an entry into `eventMap`, TS will infer the new definition of `EventType`

Does this answer to your point?

ghostd: Ok, it seems i were wrong. The best i can get for now is:
```
type EventType = keyof HTMLElementEventMap | keyof WindowEventMap;
```
Then i can use this type to define the `eventMap`:
```
const eventMap: Record<EventType, EventMapValue> = {...all init values...}
```
There are 2 consequences:
* I have to change the key to be all lowercase (eg. `dblClick` -> `dblclick`). That does not seems to be a breaking change because `events.js` already mades a text transform:
https://github.com/testing-library/dom-testing-library/blob/b23a2bc8a2c570d8faaa76694e05bf781bbe36a0/src/events.js#L25
* the future `EventType` will be larger than the present one, it contains some key the library don't use, eg: `devicelight`,  `canplay`,  `volumechange`... I guess this is not a big deal

What do you think about this?

smeijer: I think I would prefer this solution. 

@eps1lon, what do you think?

eps1lon: Sounds good to me. You two seem to know more about built-in types for this problem so I can't help much.

ghostd: I found this as an example from the documentation:
```
fireEvent.keyDown(domNode, { key: 'Enter', code: 'Enter' })
```
The case of the keys of `eventMap` seems to be part of the public API. Changing this would be a breaking change. I propose to have `keydown` in `eventMap` (so the key will be type checked against the lib.dom.ts interface), and add `keyDown` into `eventAliasMap`

I'll do a commit this day to show the changes.

ghostd: I've started something, and now i've to maintain the camelcase event list into the `eventAliasMap`:
```typescript
export const eventAliasMap: Record<string, EventType> = {
  doubleClick: 'dblclick',

  // Aliases to keep the API with camelcase events

  // Composition Events
  compositionEnd: 'compositionend',
  compositionStart: 'compositionstart',
  compositionUpdate: 'compositionupdate',
  // Keyboard Events
  keyDown: 'keydown',
  keyPress: 'keypress',
  keyUp: 'keyup',
  // Focus Events
  focusIn: 'focusin',
  focusOut: 'focusout',

  // Mouse Events
  contextMenu: 'contextmenu',
  dblClick: 'dblclick',
  dragEnd: 'dragend',
  dragEnter: 'dragenter',
  dragExit: 'dragexit',
  dragLeave: 'dragleave',
  dragOver: 'dragover',
  dragStart: 'dragstart',
  mouseDown: 'mousedown',
  mouseEnter: 'mouseenter',
  mouseLeave: 'mouseleave',
  mouseMove: 'mousemove',
  mouseOut: 'mouseout',
  mouseOver: 'mouseover',
// ...more and more...
}
```

We will to need an `AliasEventType` to correctly declare the all the fireEvent aliases (`fireEvent.dblClick`, `fireEvent.mouseEnter`, and so on). It seems we just shift the problem a little further.

The library API exposes camelcase events, which is more usable/readable for the end user IMHO. Maybe we should wait for the end of the TS migration before trying to simplify the code. For now, TS should be a "helper", not a constraint to build a clean API.

If you really want to see all the modifications i can spend more time on it this weekend. Let me know your opinions

smeijer: > If you really want to see all the modifications i can spend more time on it this weekend

Nah, let's not waste your time. I agree with:

> Maybe we should wait for the end of the TS migration before trying to simplify the code.

We've (you've) explored the available options. It looks like we have the best one for this moment.

So let get this thing merged 👍

kentcdodds:
I think we'll want to put a hold on this until #616 is merged (which should be early next week).

nickmccurdy:
Types can often help with big refactors like this, do you have any concerns about if merging this first would cause issues, or is it just that you'd prefer to prioritize the merge first?

kentcdodds:
It's just because I think we may not be quite ready to merge this. And I really want to prioritize merging in user event.

nickmccurdy:
Thanks I'm fine with that, I just wasn't sure about the status

types/event-map.d.ts
tylerthehaas: I'm not 100% sure I understand what you're wanting to do but tsc only runs on typescript files so if you want it to generate typescript definition files for only those that have been moved over that should happen naturally. I'm sure we could find what you're looking for in the [compiler docs](https://www.typescriptlang.org/docs/handbook/compiler-options.html). But if you clarify what you are looking to do I could do more research.

types/get-node-text.d.ts
tylerthehaas: shouldn't need to. Might need to add the `-d` flag when compiling or set that in your tsconfig.

nickmccurdy:
FYI: The `user-event` merge has been cancelled. We should be ready to work on this again. Let me know if you need help with anything.

ghostd:
Here are the rest of the migration (only "src", not "tests") before my vacation.

I think we can remove dtslint now the source uses TypeScript (and add the .d.ts files into .gitignore).

kentcdodds:
Thanks @ghostd.

Obviously there are a handful of merge conflicts which will need to be resolved before this is merged, but I'd like it if @testing-library/core-maintainers could review this, provide feedback, and we need to decide whether this is something we want to pursue.

marcosvega91:
> Thanks @ghostd.
> 
> Obviously there are a handful of merge conflicts which will need to be resolved before this is merged, but I'd like it if @testing-library/core-maintainers could review this, provide feedback, and we need to decide whether this is something we want to pursue.

I'm a great fun of Typescript but sometimes it requires more time to understand and use, especially when types used are very restrictive. It could help to avoid some errors but at the same time we have a lot of tests to avoid them.
Typescript is also known by less people than JavaScript so it could be more difficult for newcomers to contribute to this library.

kentcdodds:
> it could be more difficult for newcomers to contribute to this library.

That's precisely why I've not added TypeScript to any of my open source projects.

src/get-queries-for-element.ts
```diff
@@ -0,0 +1,55 @@
+import * as defaultQueries from './queries'
+
```
alexkrolick, line 2: This area seems the most complex in the migraiton... I know we've gone through a few iterations on the typings for this. Hopefully it doesn't need to change often.

src/matches.ts
```diff
@@ -1,4 +1,26 @@
-function fuzzyMatches(textToMatch, node, matcher, normalizer) {
+export type MatcherFunction = (content: string, element: HTMLElement) => boolean
+
+export type Matcher = string | RegExp | MatcherFunction
```
alexkrolick, line 3: The types for these options are quite helpful

src/matches.ts
```diff
@@ -13,7 +35,12 @@ function fuzzyMatches(textToMatch, node, matcher, normalizer) {
   }
 }
 
-function matches(textToMatch, node, matcher, normalizer) {
+function matches(
+  textToMatch: string | undefined,
```
alexkrolick, line 39: Can you have an undefined positional argument?

src/queries/display-value.ts
```diff
@@ -6,30 +6,37 @@ import {
...
+  return (element as any).value
+}
+
 function queryAllByDisplayValue(
-  container,
-  value,
-  {exact = true, collapseWhitespace, trim, normalizer} = {},
+  container: HTMLElement,
+  value: Matcher,
+  {exact = true, collapseWhitespace, trim, normalizer}: MatcherOptions = {},
```
alexkrolick, line 20: I wonder if at some point we can deprecated exact=false since we now support regex

```diff
- queryAllByDisplayValue(node, 'xyz', { exact: false })
+ queryAllByDisplayValue(node, /xyz/)
```

src/queries/label-text.ts
```diff
@@ -35,13 +38,19 @@ function queryAllLabelsByText(
     })
 
     return matcher(textToMatch, label, text, matchNormalizer)
-  })
+  }) as HTMLElement[]
```
alexkrolick, line 41: This isn't inferrable?

src/matches.ts
eps1lon: `matches(undefined, node, matcher, normalizer)`?

src/queries/label-text.ts
eps1lon: This is widening the type from `HTMLLabelElement[]`. Though I don't know why this is intended.

alexkrolick: Actually a nice constraint would be to add a queryAll function type that has `HTMLElement[]` as the return type. Then if we mess up the implementation to return something else, it would break.

eps1lon: Returning an `HTMLLabelElement` if `HTMLElement` is expected is perfectly fine though since `HTMLLabelElement extends HTMLElement`. The other way around would be problematic.

MichaelDeBoey:
`kcd-scripts` is now includes it's own `typecheck` script & auto-generates type files 🥳🎉

https://github.com/kentcdodds/kcd-scripts/releases/tag/v7.2.0

marcosvega91:
Maeby we can start migrating @testing-library/react that is smaller and than we can try with this ?

nickmccurdy:
I think it would still be best to start here, as most packages (including React Testing Library and User Event) depend on it. If we migrate dependent packages first, and then change the types emitted in DOM Testing Library, then we may have to update the dependent packages again.

Finishing migrating DOM Testing Library (including using this script and resolving merge conflicts) first may be more work up front, but I think it will be less work in the long run. Let me know if any help is needed.

kentcdodds:
I think it would be best to avoid doing the migration all at once and instead do it over time and only start generating the type defs when we're finished.

nickmccurdy:
Does that mean that we're updating or closing this pull request?

kentcdodds:
Probably much easier to close at this point

PR #616, opened 2020-06-11
feat(userEvent): build-in @testing-library/user-event

kentcdodds:
**What**: This merges `@testing-library/user-event` into `@testing-library/dom`.

**Why**: Most of the time people should be using user event.

**How**:

The first commit was: copy/paste + slightly modify.

This is NOT yet ready to be merged! I have several breaking changes, bug fixes, and new features to add before we'll be able to release this.

I've opened the PR up just to get discussion going. Would love to hear people's thoughts.

**Checklist**:

- [ ] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs)
- [ ] Types
- [x] Tests
- [ ] Ready to be merged

I'll mention that I know and accept that there will be limitations to user-event as we bring this in. But there will always be limitations. Despite that I think people are better off using `userEvent` rather than `fireEvent` for situations it works. Also, sometimes it will make more sense for people to give up on `userEvent` and instead use Cypress to test whatever it is they're trying to test.

## Current status (help wanted):

- [Docs Needed](https://github.com/testing-library/testing-library-docs/issues/490)
- [Types](https://github.com/testing-library/user-event/blob/master/typings/index.d.ts) need to be moved/updated


src/user-event/__tests__/type-modifiers.js
```diff
@@ -24,6 +24,7 @@ test('{esc} triggers typing the escape character', async () => {
     Events fired on: input[value=""]
 
     focus
+    select
```
kentcdodds, line 27: All of the `select` events are coming from calling `setSelectionRange` on the input...

Not sure how to avoid this but I think we should try... Will circle back on this when we do the audit.

kentcdodds:
Alright. Things are starting to look pretty good I think.

What we need now is to do the correctness audit and to add a few more tests to get to 100% coverage.

src/user-event/__tests__/select-options.js
```diff
@@ -0,0 +1,79 @@
...
+    click: Left (0) (bubbled from option[value="3"])
+    change
+  `)
+  const [o1, o2, o3] = options
+  expect(o1.selected).toBe(true)
+  expect(o2.selected).toBe(false)
+  expect(o3.selected).toBe(true)
+})
+
+test('sets the selected prop on the selected OPTION using nodes', async () => {
```
eps1lon, line 60: Is "using nodes" important here? I can't see how "nodes" are used in this test. Is this an implementation detail?

src/user-event/__tests__/tab.js
```diff
@@ -0,0 +1,329 @@
...
+  await userEvent.tab()
+
+  expect(number).toHaveFocus()
+
+  await userEvent.tab()
+
+  expect(radio).toHaveFocus()
+})
+
+test('should respect dom order when tabindex are all the same', async () => {
```
eps1lon, line 88: The indices in this test are not all the same.

kentcdodds:
I have some ideas on how we could simplify the test utils. Will work on that.

src/user-event/.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
+  "rules": {
+    // the wait in the loop is intentional and we don't want to parallelize
```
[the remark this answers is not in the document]
kentcdodds: I copy/pasted that comment from one situation, but there are several. I couldn't think of a single situation in this directory where we'd want the parallization so it made sense to me to just disable it for this directory.

package.json
```diff
@@ -58,7 +58,8 @@
       "import/prefer-default-export": "off",
       "import/no-unassigned-import": "off",
       "import/no-useless-path-segments": "off",
-      "no-console": "off"
+      "no-console": "off",
+      "no-func-assign": "off"
```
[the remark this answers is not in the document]
kentcdodds: This rule bugs me a bit. I prefer this:

```javascript
function foo() {
}
foo = wrap(foo)
```

Over this:

```javascript
const foo = wrap(function foo() { // don't want this to be anonymous because I want the name
})
```

Applies to all HOCs I use in React (like `React.memo`). In here we have this `wrapAsync` function which wraps all of our utilities in the config's `asyncWrapper` and I just like the way it looks better when we reassign the function.

src/user-event/__tests__/select-options.js
kentcdodds: The feature is that you can pass OPTION nodes to the function and it will select those options. That's an exposed feature so not an implementation detail.

src/user-event/__tests__/tab.js
kentcdodds: cc @benmonro, I think you wrote these tests. Could you take a look?

benmonro: ah yeah wording is a little confusing now that I read it again.  what about this:
```suggestion
test('should respect tab index order, then DOM order', async () => {
```

package.json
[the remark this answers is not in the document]
eps1lon: Why do you prefer the first one? Because I definitely didn't know you can re-assign function declarations. I wouldn't think this is valid which is exactly the rationale of this rule. The second pattern looks fine to me.

benmonro:
@kentcdodds might be nice to create a codeshift to help folks migrate their code.

package.json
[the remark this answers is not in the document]
eps1lon: In any case: If you like it why not disable it in kcd-scripts and explain the rationale? If you only have a personal, stylistic preference I'd always favor the default ruleset which makes onboarding easier.

src/user-event/.eslintrc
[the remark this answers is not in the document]
eps1lon: But this will also apply to any future file/addition. It's intentional for the code right now. But nobody will remember this rule exists if they add new code or it isn't intentional anymore.

src/user-event/__tests__/select-options.js
eps1lon:
```suggestion
test('sets the selected prop on the selected OPTION using option html elements', async () => {
```
?

kentcdodds:
A codemod would be super cool. I will not have the time to work on it unfortunately. But anyone's welcome to make one. So far, here are the differences:

1. All userEvent calls will be async
2. The `allAtOnce` option in `type` will be replaced by `userEvent.paste`
3. The imports will need to be updated (React Testing Library will re-export `userEvent`)

benmonro:
@kentcdodds Is the plan still that `fireEvent` will now be just using `userEvent`?

kentcdodds:
`fireEvent` will be unchanged through all of this. `userEvent` will use `fireEvent`. `fireEvent` will still be exposed as a lower-level primitive, but all examples and things that can be updated to `userEvent` should be.

benmonro:
oh... ok.  curious what the rationale was for that?  seems like it'd be a bit confusing to have 2 options there, no?

kentcdodds:
Because there are some things that we can't really do within `userEvent` that someone could only do with `fireEvent`.

benmonro:
would it make sense to have a lint rule to prevent the usage of fireEvent?

MichaelDeBoey:
@benmonro Something I suggested was adding a drag-event, which isn't a separate event at this moment, but could be implemented with `fireEvent`.

kentcdodds:
Perhaps a lint rule against things that *can* be done with `userEvent` (so if someone does `fireEvent.click` tell them to use `userEvent.click` for example). But there are plenty of events that you can fire that don't have a corollary `userEvent` yet.

package.json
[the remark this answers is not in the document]
kentcdodds: I've been meaning to disable it in eslint-config-kentcdodds for a while now 😅 Just haven't gotten around to it.

src/user-event/.eslintrc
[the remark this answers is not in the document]
kentcdodds: I honestly can't imagine a situation where we'd want parallelization. The async nature of this directory is to simulate the user's behavior (user's can't do two things at once). I think it's more sensible to disable this rule in this directory rather than add comments everywhere. They're just noise and make people feel like they may be doing something worng.

timdeschryver:
> A codemod would be super cool. I will not have the time to work on it unfortunately. But anyone's welcome to make one. 

I'll look into that

MichaelDeBoey:
@timdeschryver Don't know if it's possible, but I'm interested in learning how to create codemods and help you out on creating one

benmonro:
@MichaelDeBoey it's not too hard, astexplorer.net is pretty great for building out codemods, lint rules, and really anything AST based.

if you go there and choose 'jscodeshift' (which is what codemods are based on)

image

it will give you an example.  From there it's up to you to write all the jscodeshift code.

https://github.com/facebook/jscodeshift

kentcdodds:
FWIW, I've tried jscodeshift multiple times and never could figure it out, so I use a babel plugin with [babel-codemod](https://github.com/codemod-js/codemod) instead (still recommend astexplorer though).

package.json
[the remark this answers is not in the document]
alexkrolick: I didn't know that was possible either

src/user-event/__tests__/dblclick.js
```diff
@@ -1,16 +1,45 @@
...
+button#button[value=""] click
+
+button#button[value=""] pointerdown
+button#button[value=""] mousedown
+button#button[value=""] pointerup
+button#button[value=""] mouseup
+button#button[value=""] click
+button#button[value=""] dblclick
+
+*/
```
MichaelDeBoey, line 27: Seems like this can be removed? 🤔

```suggestion
```

kentcdodds: Haha, yeah, will remove that for sure.

kentcdodds:
Alrighty folks!

This is officially ready for review (thank you to everyone who's already been involved so far).

What remains is in the original comment (right now that's type definition updates and docs, both of which I'd really love help with!!).

Regarding the types, I'm thinking I'd prefer to get this merged in before #614. I think that would be quickest and easiest. This is a bit of a shake-up (though luckily *most* of the changes are in the brand new `user-event` folder).

If you'd like to help with types/docs/bugs, you can make a PR to this branch and I'll merge it in! I won't be working on anything this weekend (family time) but I may be able to slip in a review/merge here or there.

I'd love to get this released early next week. Thanks!

src/events.js
```diff
@@ -18,72 +18,78 @@ function fireEvent(element, event) {
   })
 }
 
-const createEvent = {}
+function createEvent(
```
kentcdodds, line 21: No substantial changes here. This was necessary to make it easier to create regular `window.Event` events that still had the niceties that `createEvent` had to offer.

src/event-map.js
```diff
@@ -1,350 +1,360 @@
 export const eventMap = {
```
kentcdodds, line 1: Formatting change only.

kentcdodds:
Tracking docs contribution: https://github.com/testing-library/testing-library-docs/issues/490

kentcdodds:
Sorry for the massive ping (feel free to unsubscribe from this issue), but I just want to give a huge shout-out to all the contributors of `@testing-library/user-event` over the year and a half (especially @Gpx). I was skeptical that userEvent could reasonably deliver on what it promises, and while it's not perfect and there are limitations, it's really quite good and I'm excited for it to take it's rightful place as the officially recommended and built-in way to interact with components under test in the Testing Library family.

Thank you contributors!

applause

👏👏👏👏👏👏👏👏👏👏👏👏👏👏👏👏👏👏👏👏

@Gpx @weyert @twhitbeck @MichaelDeBoey @michaellasky @shomalgan @calebeby @afontcu @skywickenden @bogdanbodnar @zperrault @FLGMwt @benmonro @GentlemanHal @YuanchengWu @maheshjag @jmcriffey @kandros @jordyvandomselaar @ilyamkin @klujanrosas @jsmapr1 @wachunga @bdh1011 @Meemaw @pomber @Raynos @skovy @vadimshvetsov @9still @rbusquet @dougbacelar @kayleighridd @malcolm-kee @kelvinlzhang @krzysztof-hellostudio @hontas @hudochenkov @nanivijay @tpict @nvh95 @nickmccurdy @timdeschryver (hope I didn't miss anyone 😅)

Raynos:
:+1: I use both packages in my tests and it makes sense for it to be a single package to install and import.

timdeschryver:
I think I got a working implementation for the codemod. It's written with `ts-morph` tho, does that matter? 

See https://github.com/timdeschryver/tl-user-event-mod for the repo.
@MichaelDeBoey feel free to help out, I'm sure there are some cases that I haven't covered yet, and adding more examples would be helpful.

You can already give this a shot by running.
If it does something unexpected, or isn't covering a case feel free to [create an issue](https://github.com/timdeschryver/tl-user-event-mod/issues/new). 

```bash
npx tl-user-event-mod
```

kentcdodds:
TL;DR, please test out the alpha version of `@testing-library/dom` and/or `@testing-library/react` as a replacement for `@testing-library/user-event`. In short:

```javascript
npm uninstall @testing-library/user-event
npm install @testing-library/react@alpha
```

Then find/replace `import userEvent from '@testing-library/user-event'` with `import {userEvent} from '@testing-library/react'` and test it out! Please take note of [the mentioned changes from `@testing-library/user-event`](https://github.com/testing-library/testing-library-docs/issues/490#issuecomment-643557354) (**Especially the bit that says now all `userEvent` methods are `async`, so you'll need to add `await` to the front of them). In addition to those there are bug fixes that may affect people.

Ok, now the long part, for @testing-library/all-maintainers:

```
npm publish --tag alpha --access public 
+ @testing-library/dom@7.15.0-alpha.1

npm publish --tag alpha --access public 
+ @testing-library/react@10.3.0-alpha.1
```

I just published an alpha version of this as well as `@testing-library/react`, please install and test this out in your projects!

For other @testing-library/all-maintainers, if your project relies on `@testing-library/dom`, please take this opportunity to publish an alpha version with the alpha of `@testing-library/dom`.

If you aren't sure how to do this, here's what I do:

1. Update your `package.json` to be whatever it would be if you published this today (probably a minor version) + `.alpha-1` So if you're current version is `10.2.1` then the new version should be `10.3.0-alpha.1` (note, because I'm using semantic-release, I don't actually commit this change)
2. `npm install @testing-library/dom@alpha`
3. Make sure you're re-exporting `userEvent` from `@testing-library/dom` (if you're doing `export * from '@testing-library/dom'` then you're good).
3. `npm test && npm run build`
4. `npm publish --tag alpha --access public`

Then let people know and they should be able to test it out in their own projects!

And report back on any issues/successes please! Thanks :)

kentcdodds:
Oh, and I tried it on https://github.com/kentcdodds/bookshelf and I just needed to add a few `await`s, change some tests to `async`, and remove one place where I was using `allAtOnce`.

SUCCESS!

nickmccurdy:
@kentcdodds I think you meant to ping `@testing-library/core-maintainers`

kentcdodds:
I probably could use help organizing the teams a little bit 😅 it's kinda hacked together over time.

kentcdodds:
Just hit one interesting case...

```javascript
  await userEvent.click(screen.getByLabelText(/search/i))
  await waitForElementToBeRemoved(() => screen.getByLabelText(/loading/i))
```

This will *sometimes* fail because the callback passed to `waitForElementToBeRemoved` will synchronously call the callback to make sure the element you're waiting to be removed is actually there to begin with (if it's not it'll throw an error).

Before our changes, `userEvent.click` was synchronous so getting the loading element happened synchronously, so this never failed. However, now, we have a race condition. If the async thing thing that happens finishes before the `userEvent.click` promise resolves, then the loading element will be removed from the screen and `waitForElementToBeRemoved` won't be able to find it.

A workaround for this is:

```javascript
  await userEvent.click(screen.getByLabelText(/search/i))
  await waitFor(() =>
    expect(screen.queryByLabelText(/loading/i)).not.toBeInTheDocument(),
  )
```

This is basically what `waitForElementToBeRemoved` does except without the existence check. I'm not happy with this workaround, but I'm unsure of what to recommend people do in situations like this (which I believe will not be uncommon).

I welcome suggestions.

nickmccurdy:
Could we work around this by just removing the callback form of `waitForElementToBeRemoved`? Then it would be up to the test author to make sure the element is loaded in the right order, giving them more control over race conditions.

kentcdodds:
You mean like this?

```javascript
  await userEvent.click(screen.getByLabelText(/search/i))
  const loading = screen.getByLabelText(/loading/i)
  await waitForElementToBeRemoved(loading)
```

That API currently exists actually, though it's not entirely useful because in the context of React, the framework may actually re-use that DOM node for the thing that gets rendered in it's place. (This was a real facepalm moment for me after I released that feature 🤦‍♂️)

Just saw your suggestion. The problem is the loading element doesn't appear until after `search` has been clicked.

But because APIs are mocked (with `msw`), it responds so fast that the loading disappears before the next line happens, so my example in this comment won't work either.

nickmccurdy:
My intention with removing the callback form was to force users to write tests that would have less race condition issues. Your example seems okay to me for vanilla JS.

For React, wouldn't relying on the DOM rendering order be too much of an implementation detail anyway? I feel like the user should be testing if DOM content does/doesn't exist rather than hoping that React doesn't reuse Nodes. If I'm understanding correctly, in that case `waitForElementTOBeRemoved` wouldn't even be necessary.

kentcdodds:
Yeah, it's more of `waitForThisQueryToFail`. I would argue that being able to pass a DOM node and saying `waitForThisDOMNodeToBeRemoved` is an implementation detail. The user doesn't care about DOM nodes appearing in the DOM or not. So really, the callback form is the non-implementation detail version of this.

It doesn't really make a difference anyway, the API isn't the problem, it's the race condition.

A potential solution would be to remove the initial check of existence, but then we're not getting confidence that the loading spinner shows up at all...

Here's a more complete workaround I think, that may bring some clarity:

```javascript
const clickPromise = userEvent.click(screen.getByLabelText(/search/i))
expect(await screen.findByLabelText(/loading/i)).toBeInTheDocument()
await clickPromise
await waitFor(() =>
  expect(screen.queryByLabelText(/loading/i)).not.toBeInTheDocument(),
)
```

That's giving us all the confidence that we need without any implementation details. But it's a sad excuse for a workaround 😬

nickmccurdy:
Could we just have two assertions in separate `waitFor` calls instead?
```js
  await userEvent.click(screen.getByLabelText(/search/i))
  await waitFor(() => screen.getByLabelText(/loading/i).toBeInTheDocument())
  await waitFor(() => screen.getByLabelText(/loading/i).not.toBeInTheDocument())
```

kentcdodds:
That would likely be worse because it extends the time between when the loading thing appears then disappears before we look for it.

nickmccurdy:
Is that just a performance issue or are you saying this also has race condition issues? If it's the latter, I wouldn't worry about it personally, I'd much rather have an accurate test that's easy to read than a fast test (similar reason why I'd rather use JSDOM and testing-library over mocking everything).

kentcdodds:
It's a race condition issue. The test will likely fail more often than not.

smeijer:
When I look at this test:

```js
await userEvent.click(screen.getByLabelText(/search/i))
await waitForElementToBeRemoved(() => screen.getByLabelText(/loading/i))
````

I don't think that this is a work around:

```js
const clickPromise = userEvent.click(screen.getByLabelText(/search/i))
expect(await screen.findByLabelText(/loading/i)).toBeInTheDocument()
await clickPromise
await waitFor(() =>
  expect(screen.queryByLabelText(/loading/i)).not.toBeInTheDocument(),
)
```

Let me explain. You approach it as `race condition` in the library. And although the term might be correct, it is what I would expect this test to do.

The component being tested as flow like:

```
1. click search button
2. show loading message
3. trigger network request
4. network response comes in
5. hide loading message
6. render results
```

And yet, the initial test includes `clicking` the button, and waiting for  `loading` to be removed. Shouldn't the tester also be waiting/checking for it to be rendered? 

```js
// click search
await userEvent.click(button);

// check if the loading msg was rendered
getByText('loading')

// wait for the loading msg to disappear
await waitForElementToBeRemoved(() => queryByText('loading'))
```

If it's about the cleanness of the API, `waitForElementToBeRemoved` could be extended to include the async waiting as well.

```js
// pseudo
function waitForElementToBeRemoved(text) {
  // get element sync, to ensure it exists
  let target = queryBy(text);

  // element doesn't exist, wait a while for it
  if (!target) {
    target = await findBy(text)
  }

  // still not exists :(
  if (!target) {
    throw Error(`
      element that should be removed, 
      didn't exist in first place
    `);
  } 
   
  // wait for removal
}
```

But all of this might give new trouble in the future, once Suspense gets adopted. The scenario that only shows the `fallback` (loading msg) when the network request is taking longer than `500 ms` (arbitrary number), changes this entirely. At that moment, it's by design that the spinner never appeared. Assuming the latency wasn't faked.

So, in my opinion, the initial test was flaky. We shouldn't wait for elements to be removed if they aren't there, to begin with.

kentcdodds:
Thanks for the feedback @jaredkhan. That technically works, but it's incomplete because the promise that's returned will not be handled (so lint rules and TS will have a problem with that I believe). Also, I think that lots of people wouldn't think to use `find*` queries after that click. They would assume that they should be able to use `get*` on the next line.

src/user-event/type.js
```diff
@@ -435,6 +448,11 @@ async function type(
           } else {
```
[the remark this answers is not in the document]
kentcdodds: Good point. Will update. Thanks!

afontcu:
Looks great! This is def a step forward 🙌 

I've been setting up an alpha version of Vue Testing Lib (not published yet. [here's the PR](https://github.com/testing-library/vue-testing-library/pull/147)). It was quite painless, but I'm running into some issues with `<textarea>` elements:

```diff
  Object {
    "rating": "3",
-   "review": "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
+   "review": "orem ipsum dolor sit amet, consectetur adipiscing elit.L",
    "title": "An Awesome Movie",
  }
```

only thing I changed was `fireEvent` -> `userEvent` (https://github.com/testing-library/vue-testing-library/pull/147/files#diff-2020a6f84827c2e834b00ff2d26d8347R34), and I even wrote a standalone test for a `<textarea>` which resulted in a Maximum call stack size exceeded error. Did anyone else experience weird inconsistent behaviours with textareas?

kentcdodds:
Interesting! I've not seen that, sorry. I noticed there code was calling fireEvent.update before. Is that something specific to vue?

afontcu:
> there code was calling fireEvent.update before. Is that something specific to vue?

yes! It is a small utility that helps to deal with Vue-specific issues. Nothing fancy: https://github.com/testing-library/vue-testing-library/blob/master/src/vue-testing-library.js#L157-L160

(With the merger of `userEvent`, `fireEvent.update` won't be needed, so I plan on removing it in a future release).

I ll try to dig a little deeper and see what I can come up with!

kentcdodds:
I'd really love @afontcu and @eps1lon and anyone else chime in on the race condition issue above.

I somehow missed @smeijer's comment. I think there's a general misunderstanding of what this issue really is, so I've made another example that runs in codesandbox (and locally): https://codesandbox.io/s/vibrant-chaum-xb73y?file=/src/__tests__/index.js

Here's what happens when I use `userEvent.click`:

As soon as that `click` is dispatched, `fireEvent` (which is wrapped in `act`) finishes and React flushes the effect synchronously. This means that as soon as the click is dispatched, we trigger the async function call and trigger a re-render of the component thanks to the state change which will also be synchronous. So by this time, we've synchronously started the async request *and* synchronously re-rendered with the loading state.

By now, we're still waiting for `userEvent.click` to resolve, *and* we now have another promise that needs to resolve. You might expect that the `userEvent.click` (which is wrapped in an async `act`) should resolve first because it was scheduled first and it does, sometimes... But other times the component's promise will resolve first. When it does, we get a state update and react re-renders the component without the loading state.

So the test passing depends on which promise resolves.

Taking a step back, this is absolutely what a manual tester would do to test this UI:

1. click the button
2. assert the loading spinner is there
3. assert the loading spinner is gone
4. continue...

That's what this code is doing:

```javascript
  await userEvent.click(button)
  await waitForElementToBeRemoved(() => screen.queryByLabelText(/loading/i))
```

But because of this race condition issue, we've got an issue.

This is a ship stopper. We need to get this fixed before merging. I'm not sure what to do and would love feedback and ideas.

src/user-event/click.js
```diff
@@ -0,0 +1,117 @@
...
+  switch (element.tagName) {
+    case 'LABEL':
+      await clickLabel(element, init, {clickCount})
+      break
+    case 'INPUT':
+      if (element.type === 'checkbox' || element.type === 'radio') {
+        await clickBooleanElement(element, init, {clickCount})
+        break
+      }
+    // eslint-disable-next-line no-fallthrough
```
eps1lon, line 102: I'd duplicate the call here. That rule has a strong rationale and duplicating once is really not an issue compared to reasoning about switches and fallthroughs.

eps1lon:
If I followed this correctly then the race condition happens in the following snippet, right?

>   await userEvent.click(screen.getByLabelText(/search/i))
  await waitForElementToBeRemoved(() => screen.getByLabelText(/loading/i))

I think this is similar to the issues we're going to face with concurrent react. A mocked scheduler/timers is probably the only complete solution so that we can control when we want updates to be flushed.

In this example we almost never want a `fetch` to complete on the same line, no?

Lukas-Kullmann:
> By now, we're still waiting for userEvent.click to resolve

I think this is the true underlying problem: `userEvent.xxx` being asynchronous functions. Could you maybe give a reasoning why they are async now? In the "old" user-event package, only `userEvent.type` was async, now all of them are. 

In my opinion, it does not make sense that `userEvent.xxx` are async (except `userEvent.type`). Them being async suggests that someone wants to do something like this:

```
const p = userEvent.click(element);
// do something
await p;
```

But what would you want to do between initializing the click and waiting for it to be done? I can't come up with anything.

So in the end, I would suggest removing the async nature from the userEvent functions (except `userEvent.type`). 
I tried this. It fixes the test you posted in @kentcdodds' comment.

eps1lon:
> In my opinion, it does not make sense that userEvent.xxx are async 

I think the underlying problem is that all dispatches are awaited sequentially instead of once. The dispatched events shouldn't be awaited since that wouldn't be the case in a browser anyway. In react we should wrap all dispatches in a single `act` instead of wrapping every single dispatch i.e.

```js
act(() => {
  fireEvent.mouseDown();
  element.focus();
  fireEvent.mouseUp();
  element.click();
});
```

instead of 
```js
act(() => void fireEvent.mouseDown())
act(() => void element.focus())
act(() => void fireEvent.mouseUp())
act(() => void element.click())
```

or for Promises:

```js
await Promise.all([fireEvent.mouseDown(), element.focus(), fireEvent.mouseUp(), element.click()]);
```
instead of
```js
await fireEvent.mouseDown();
element.focus();
await fireEvent.mouseUp();
element.click();
```

Lukas-Kullmann:
This problem is not related to the react testing library. It is an inherit problem of the dom testing library. You can reproduce the bug with just the dom testing library.

Here is a small code sandbox with @testing-library/dom only: https://codesandbox.io/s/lingering-firefly-rl9fn

In any case

> I think the underlying problem is that all dispatches are awaited sequentially instead of once

In that case, the dispatches being async is the underlying problem to this problem. Why do they need to be async?
Maybe the bug can be fixed if all dispatchers are awaited in parallel. But I think the easier solution would be to make this thing synchronous again.

marcosvega91:
maybe is not the right solution, but if we save a DOM diff for every event between start and end of the event? Then in `waitForElementToBeRemoved` we could check what happened before.
😕

kentcdodds:
The rational for making it all async (and making every fireEvent async) is that a user *cannot* fire a `mousedown` and then a `mouseup` synchronously after it. Those two events will happen on different ticks of the event loop. It's possible that the browser fires *some* events synchronously (maybe after the `mouseup`, the `click` is dispatched synchronously).

In any case, I don't think that awaiting between fireEvents is the problem here because the click event we're listening for is the last one that's fired. We don't await anything after that.

The only solution I can see as "working" is to make it all sync again, but then it's not really resembling the way the software is used and I'd like to do that...

eps1lon:
> The rational for making it all async (and making every fireEvent async) is that a user cannot fire a mousedown and then a mouseup synchronously after it

We cannot fire them sync as well. They are always consecutive. The issue right now is that we put each event in a separate tick of the (micro?)-task loop. We need an explainer for that. Does the ui-events spec define it that way?

kentcdodds:
I don't think the spec needs to. How could a user possibly click their mouse fast enough to fire mousedown and mouseup on the same tick off the event loop? They couldn't.

kentcdodds:
This may be one of the trade-offs we have to make on correctness to practicality though...

eps1lon:
> They couldn't.

*Maybe* for mousedown and mouseup. But right now you're putting every fired event in a separate tick without any justification. Just saying "you couldn't do it" is not an argument.

kentcdodds:
We'd have the same problem for anyone testing a similar situation for typing. So I'd a loading indicator pops up whole the user is typing, they would have this same problem with async type.

kentcdodds:
I realize that it's possible the browser fires some events in the same event loop tick. I'm not sure how to determine that. But it really doesn't matter. Once we have to make anything async, then we'll have this problem. Remember that the click event is the one that we're listening got in my example and that's the last event that fires in the sequence.

eps1lon:
I'm just now realizing the extent of all this async wrapping. Why is focus async? Why do we fire focusIn and await it? (btw the whole implementation is only required for JSDOM. It makes more sense to [upstream focusin/focusout](https://github.com/jsdom/jsdom/issues/2708)).

I don't feel comfortable merging user-event in its current state. There are just too many assumptions that may very well be correct but aren't documented in any way. This works for a personal library but not one where we want to share maintenance.

Lukas-Kullmann:
> but then it's not really resembling the way the software is used and I'd like to do that...

This is of course a valid design decision and a killing argument for every counter argument argument that goes into the direction of making it synchronous.

The thing that concerns me however is that tests never fully mimic the real world. The reason why we run into tgis bug here is because we have an unrealistically fast UI update from state A -> loading -> state B. In the real world, that would never happen in just one async tick. And of course, the bug is solved if you introduce a delay before generating the new random number of, say, 200ms.

But that's also how network mocking usually works. You resolve the network request after just one async frame (or at least a very shorter time than it would usually take in the production app). So as long as we resolve network requests so fast, we cannot really affort to have event phases that take so long (compared to all the other async tasks that you have in your app).

kentcdodds:
I'm beginning to feel that this needs more research before we can commit to bringing this into Dom testing library core. we can take the improvements that I've made in this PR and port them back to the original user event module and we'll wait on for the research before we bring it into core.

kentcdodds:
I love feedback on that plan. Just give a thumbs up to that comment if you agree or comment with a argument against it if you disagree.

Lukas-Kullmann:
I think this is not a good idea. Porting this to user-event does not make this type of bug go away. It will just lead to a very painful migration for the users of the library to the new version.

And I think this would also be a thing that quite some users would run into. And there is nothing that the library can do to nudge the user into the right direction for finding the cause of this bug. Let alone how to solve it.
I already find it quite painful find the cause of a failing test that says "Cannot find element with text/role/whatever".
At the moment you can most likely resolve this issue by replacing a "getBy" with a "findBy". With this change, the search space for the error extends "the element is not there yet" to also "the element is already gone". In my experience, the latter does not occur often at the moment. With bug still present, this would become quite usual.

kentcdodds:
Sorry, to be more clear, I'd move all my improvements I've made here back to user-event, without the "everything is async now" bit :)

kentcdodds:
Ok, I'm going to port back my improvements (sans the async-ify everything) and we'll close this here.

Lukas-Kullmann:
> Sorry, to be more clear, I'd move all my improvements I've made here back to user-event, without the "everything is async now" bit :)

Then the question becomes, why not removing the async stuff and leaving it here? 🙂

kentcdodds:
Fair point. I guess it's because I don't think that it's *right* if we're not doing async and I don't want it in core. It's just not ready yet I think.

kentcdodds:
Ok, it's official. We're not merging this right now. I've made all the same improvements that I worked on for the last two weeks to user-event itself. Except *everything* is sync (even `type` unless you pass a `delay`, in which case it's async, but you're opting into that behavior).

I'm disappointed about this (especially since I know several frameworks need everything to be async, so this means we're fragmented), but this is the best we can do.

kentcdodds:
Oh, and here's the PR: https://github.com/testing-library/user-event/pull/348

nickmccurdy:
Could we continue merging in `user-event` but without the `async` API changes? I'm not sure they both have to be done together, especially if we still plan on releasing `user-event`.

kentcdodds:
I really don't think we should. I'd prefer to make it async somehow. That's more correct.

But I'm out of bandwidth on this so I'm no longer going to be working on it 😫

PR #621, opened 2020-06-12
fix(TS): Recommend updating local type definitions

nickmccurdy:
**What**: Recommend updating local type definitions

**Why**: It seems like the types have moved back from DefinitelyTyped to this repository, so this repository should be used to maintain them instead of DefinitelyTyped (note that we also may need to remove them to reduce confusion, I'm not sure how though)

**How**: Updated pull request template

**Checklist**:

- [ ] Tests N/A
- [x] Ready to be merged


kentcdodds:
Note, only use "fix" when you want to trigger a release 👍

nickmccurdy:
Should it be chore or no prefix?

PR #624, opened 2020-06-12
chore: Add package lock to fix CI

nickmccurdy:
**What**: Enables npm's package lock (`package-lock.json`) and keeps it committed

**Why**: The CI build is currently failing because it is installing dependencies differently from my local machine, in such a way that for some reason `dtslint`'s version is causing dozens of errors. My local build validates just fine, so I uploaded my own `package-lock.json`, which is now working perfectly in CI. It's important that we keep it committed to avoid issues like this in the future. Fortunately npm obeys, updates, and even resolves merge conflicts in the lock file automatically, so we shouldn't have any issues with contributors understanding how to use it.

**How**: Removed npm config and committed `package-lock.json`

**Checklist**:

- [ ] Tests N/A
- [x] Ready to be merged


smeijer:
> so we shouldn't have any issues with contributors understanding how to use it.

About that last part, I have my doubts.

1. As I understand it, there is a difference between `npm i` and `npm ci`.

   `npm ci` removes your `node_modules` prior install and installs dependencies based on `package-lock` data. It will also never update `package.json` or `package-lock.json`.

2. Different versions of npm can create different versions of `package-lock.json`

  More than once I reviewed pull-requests that included an updated `package-lock`, even while there was no change in dependencies. Not a huge deal, but kind of annoying.

To prevent the issues from 2) contributors would need to use `npm ci` to install their dependencies, and stop using `npm i`. Only use `npm i` to add or update dependencies.

Besides that, I'm pro adding `package-lock.js`. So thanks for opening this PR :+1: :sunglasses:

kentcdodds:
I'm pretty sceptical of using lock files in libraries. But if one other maintainer gives their approval then we can merge this.

nickmccurdy:
> 1. As I understand it, there is a difference between `npm i` and `npm ci`.

@smeijer Since we're just talking about users manually installing to contribute here (and not CI), they should only be using `npm i(install)`, so the differences wouldn't affect them directly.

> 2. Different versions of npm can create different versions of `package-lock.json`
>    More than once I reviewed pull-requests that included an updated `package-lock`, even while there was no change in dependencies. Not a huge deal, but kind of annoying.
>    To prevent issues contributors would need to use `npm ci` to install their dependencies, and stop using `npm i`. Only use `npm i` to add or update dependencies.

@smeijer This sounds like a workaround I wouldn't recommend. This behavior is just there for automated installs. Users should always be keeping the format up to date. If the concern is more about how frequently the lock file format updates, we can lock the npm version or use Yarn which arguably has more stable lockfile semantics.

> I'm pretty sceptical of using lock files in libraries. But if one other maintainer gives their approval then we can merge this.

@kentcdodds npm actually refuses to publish `package-lock.json` to packages, so they're strictly used in development. However, keeping development dependencies consistent is just as important as keeping dependencies consistent for a production application, as scripts like CI will rely on their consistent behavior, and inconsistencies can cause subtle bugs to be published with the library if testing tools only run properly in some environments (fortunately in this case I was able to figure out the issue, but you're often out of luck if you forget the right package versions if you aren't using a lock file). And I'd be fine with Yarn too to be clear, I'm just trying to be consistent with our npm scripts and the org's preference to npm over Yarn.

smeijer:
> If the concern is more about how frequently the lock file format updates, we can lock the npm version or use Yarn (which arguably has more stable lockfile semantics as it adopted more locking features early on and has more community-driven RFC processes).

No, my concern is that if you use `npm` version `x` and I'm using `y`, we can create different `package-lock.json` files.

That means, that if we both keep our `package-lock`s "up to date", that would mean that we can keep submitting `package-locks` in our pull-requests all the time, even while the dependencies itself never changed. That is until we sync our `npm` versions.

Yes, `yarn` is better in this. But that brings the discussion to a whole other level. (do we really want this project to move from npm to yarn?)

I'm using `package-lock` in projects myself, but I haven't been a huge fan of it. I see the benefits. But it does have issues. So I'm not against adding it, but I doubt if it's worth the pain.

nickmccurdy:
I understand that lock files aren't perfect (especially npm because of how unstable the format is), but they solve this CI issue pretty easily, and I can't really think of a better alternative in this situation.

kentcdodds:
I definitely recommend lockfiles in apps. But for libraries, I'm mostly concerned with the stuff I publish to work with the latest versions (including transitive dependencies) of everything and if we use lockfiles then we have to regularly update those versions (no I don't want to use a bot to keep things updated, that's a nightmare). It's unfortunate that we can't use lockfiles for only our devDeps because that would be nice.

However, even in that case, @smeijer's arguments are solid. Any time I've tried to use a lockfile in an open source project that I get contributions to, people inevitably and inexplicably update the lockfile. That leads to confusion (and is potentially a security vulnerability if someone manually updates the lockfile to install a malicious package).

The alternative that I prefer is figure out why the latest resolved version of packages is breaking our build and fix that. If it means we need to remove a `^` in our package.json that's fine with me (though sometimes the issue is in a transitive dep and I really don't have a great answer for that). Most of the time such issues go away after a little while. It's never been a huge problem for me.

nickmccurdy:
Unfortunately it's hard to fix these issues without a package lock tracked in version control. I personally find it easier to just commit them and be cautious about when updates are needed.

kentcdodds:
I deleted the cache and re-ran the build and now it's passing.

PR #627, opened 2020-06-13
chore: add option to get specific query suggestions

smeijer:
### 1 - Add support for getting specific query method suggestions

**What**:
I've added an option to specify the desired method to the `getSuggestedQuery` method.

**Why**:
To enable tools (such as https://testing-playground.com) to show specific suggestions instead of only the most recommended one.

**How**:
`getSuggestedQuery` accepted the arguments `element, variant`, I've extended that to `element, variant, method`.

Method defaults to "best", but if specified, only the requested query will be suggested, or none at all.

This enables us to use it like:

```js
getSuggestedQuery(input, 'get', 'PlaceholderText'));
```

If that call results in a valid suggestion, the suggestion is returned otherwise it will return `undefined`. It will not fall back to alternative methods!

The default behavior of the method has been left unchanged. A call like:

```js
getSuggestedQuery(input, 'get'));
```

Still returns either the best query available, or `undefined`.

### 2 - added `getByTestId` suggestion

Sorry, I should learn to limit the scope of pull-requests. Even when they are tiny.

While working on the above, I noticed that a suggestion for `getByTestId` wasn't implemented. Testing-playground will render it red! But when there are really no other options, this query should still be suggested. It is part of the library after all.

So, I've also added the `getByTestId` suggestion, as last option available.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/suggestions.js
+++ b/src/__tests__/suggestions.js
@@ -353,3 +353,95 @@ test('getSuggestedQuery returns rich data for tooling', () => {
 
   expect(getSuggestedQuery(div).toString()).toEqual(`getByText(/cancel/i)`)
 })
+
+test('getSuggestedQuery can return specified methods in addition to the best', () => {
+  const {container} = render(`
+    <label for="username">label</label>
+    <input
+      id="username"
+      name="name"
+      placeholder="placeholder"
+      data-testid="testid"
+      title="title"
+      alt="alt"
+      value="value"
+      type="text"
+    />
+    <button>button</button>
+  `)
+
+  const input = container.querySelector('input')
+  const button = container.querySelector('button')
+
+  expect(getSuggestedQuery(input, 'get', 'Role')).toMatchObject({
+    queryName: 'Role',
+    queryMethod: 'getByRole',
+    queryArgs: ['textbox', {name: /label/i}],
+    variant: 'get',
+  })
+
+  expect(getSuggestedQuery(input, 'get', 'LabelText')).toMatchObject({
+    queryName: 'LabelText',
+    queryMethod: 'getByLabelText',
+    queryArgs: [/label/i],
+    variant: 'get',
+  })
+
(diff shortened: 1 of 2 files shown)


nickmccurdy:
Great idea, thank you! I'm interested if there's and followup work we can do to make this more useful, especially in the linter plugins.

smeijer:
> especially in the linter plugins.

I'm not entirely the right person to answer that question. But as far as I understand linters and understood from a Twitter thread, that will be a hard thing to do. Which is why `@testing-library/dom` can now throw when it detects better query options.

The thing is, linters verify on a line-by-line basis. They don't eval the code. And `@testing-library/dom` doesn't show the user with all the options, it only suggests the best. Which I believe is a good thing.

So, that's entirely why we have testing-playground (soon to be official! :heart:)

But maybe @benmonro has an idea on how to get more out of the suggestions?

marcosvega91:
Maybe we can add a new method that will return a list of suggestions ordered by relevance ?

smeijer:
Do you have a specific use case in mind?  

I'm not sure if a function like this belongs to `@testing-library/dom`. As this project only suggests the best query available, and not all queries that are possible. 

The base functionality that you describe is also quite trivial for other libraries to build themselves, by just using `getSuggestedQuery`. The more advanced case (that takes the entire dom context into account) is a bit harder, but also moves more out of the scope of this project.

marcosvega91:
I wrote that but I agree with you, testing library shouldn't do this

benmonro:
@nickmccurdy what did you have in mind?  As Stephan mentioned the lint plugin can only analyze code statically. In order to make a suggestion you need to inspect a live dom.  

I am still curious though because I do love writing lint rules.  🙂

benmonro:
@kentcdodds @smeijer  if we're adding test id to this shouldn't it check config to rather than hard coding to data-testid attribute?  Not that big of a deal but our team uses an alternate attribute

smeijer:
@benmonro, you're right. We should use the config for that.

PR #628, opened 2020-06-13
fix: canSuggest should not be case sensitive 

marcosvega91:
**What**:

Following up #627 I have updated `canSuggest` function because it only works when methods where exactly the same.

**Why**:

because I think that  `getSuggestedQuery` should return the same result for method 'Role' and method 'role'.

**How**:

I have compared lowecase strings.

**Checklist**:

- [X] Tests
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/suggestions.js
+++ b/src/__tests__/suggestions.js
@@ -373,7 +373,9 @@ test('getSuggestedQuery can return specified methods in addition to the best', (
   const input = container.querySelector('input')
   const button = container.querySelector('button')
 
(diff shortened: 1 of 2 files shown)


kentcdodds:
@all-contributors please add @marcosvega91 for code and tests

smeijer:
> I'm good with this

It's not a big deal, but how are we going to support this when adding typescript support? :innocent: 

```ts
type method = role | Role | getByText | getbytext | ....
```

Seems unwanted to me.

marcosvega91:
But in this case you will have an error so I think that we don't need to cover both cases in the type

PR #630, opened 2020-06-13
fix: typescript definitions for suggestions

marcosvega91:
**What**:

I have updated typescript definitions

**Why**:

As before you can't pass both variant and method parameters

**How**:

I have added two new types `Variant` and `Method` and used them as types for the respective params.
Method params is also optional

**Checklist**:

- [ ] Tests
- [X] Typescript definitions updated
- [X] Ready to be merged

### DIFF
--- a/types/suggestions.d.ts
+++ b/types/suggestions.d.ts
@@ -13,16 +13,24 @@ export type Variant =
 
 export type Method =
   | 'Role'
+  | 'role'
   | 'LabelText'
+  | 'labeltext'
   | 'PlaceholderText'
+  | 'placeholdertext'
   | 'Text'
+  | 'text'
   | 'DisplayValue'
+  | 'displayvalue'
   | 'AltText'
+  | 'alttext'
   | 'Title'
+  | 'title'
   | 'TestId'
+  | 'testtd'
 
 export function getSuggestedQuery(
   element: HTMLElement,
-  variant: Variant,
+  variant?: Variant,
   method?: Method,
 ): Suggestion | undefined
--- a/types/suggestions.d.ts
+++ b/types/suggestions.d.ts
@@ -3,4 +3,26 @@ export interface Suggestion {
   toString(): string
 }
 
-export function getSuggestedQuery(element: HTMLElement): Suggestion | undefined
+export type Variant =
+  | 'get'
+  | 'getAll'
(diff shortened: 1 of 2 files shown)


types/suggestions.d.ts
```diff
@@ -3,4 +3,26 @@ export interface Suggestion {
...
+  | 'PlaceholderText'
+  | 'Text'
+  | 'DisplayValue'
+  | 'AltText'
+  | 'Title'
+  | 'TestId'
+
+export function getSuggestedQuery(
+  element: HTMLElement,
+  variant: Variant,
```
smeijer, line 26: variant defaults to `get`

```suggestion
  variant?: Variant,
```

types/suggestions.d.ts
```diff
@@ -3,4 +3,26 @@ export interface Suggestion {
...
+
+export type Method =
+  | 'Role'
+  | 'LabelText'
+  | 'PlaceholderText'
+  | 'Text'
+  | 'DisplayValue'
+  | 'AltText'
+  | 'Title'
+  | 'TestId'
```
smeijer, line 22: This is incompatible with your change from #628? :innocent:

types/suggestions.d.ts
marcosvega91: yes I didn't see `function makeSuggestion(queryName, content, {variant = 'get', name}) {`

maybe is better to move the default value of variant to exported `getSuggestedQuery` ?

types/suggestions.d.ts
marcosvega91: As for the other thread, you will receive an error so is simpler than with javascript to understand that you have to pass `Role` instead of `role`.

smeijer: So we configure typescript to only accept `Role`, but at the same time we make javascript checks case insensitive by adding two `toLowerCase()` calls?

It's not a huge issue. Just feels wrong to me.

kentcdodds: FWIW, I'm fine with it, though I think it would be better to allow the lower case version here as well.

types/suggestions.d.ts
smeijer: I think it should be there as well. Both methods can have the default initializer.

PR #641, opened 2020-06-14
fix(userEvent): don't select option if select is disabled

nickmccurdy:
Migrated from testing-library/user-event#344 by @ben-dyer

**What**: Prevent a select's value from changing with `selectOptions` if it is disabled

**Why**: It's possible to change the value of a disabled select using `selectOptions`. For instance this is passing:
```
  const {
    container: {firstChild: select},
  } = render(
    <select disabled>
      <option>No value selected</option>
      <option value="1">1</option>
      <option value="2">2</option>
      <option value="3">3</option>
    </select>,
  )

  expect(select.selectedIndex).toBe(0)

  userEvent.selectOptions(select, '1')
  expect(select.selectedIndex).toBe(1)
```

**How**: Exit early in `selectOptions` if the element is disabled, similar to what is done in `clear` and `upload`

**Checklist**:

- [x] Tests
- [x] Ignore disabled options
- [x] Ready to be merged

### DIFF
--- a/src/user-event/__tests__/select-options.js
+++ b/src/user-event/__tests__/select-options.js
@@ -119,3 +119,17 @@ test('does not select anything if select is disabled', async () => {
   expect(o2.selected).toBe(false)
   expect(o3.selected).toBe(false)
 })
+
+test('does not select anything if options are disabled', async () => {
+  const {select, options, getEventSnapshot} = setupSelect({
+    disabledOptions: true,
+  })
+  await userEvent.selectOptions(select, '2')
+  expect(getEventSnapshot()).toMatchInlineSnapshot(
+    `No events were fired on: select[name="select"][value=""]`,
+  )
+  const [o1, o2, o3] = options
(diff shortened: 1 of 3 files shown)


src/user-event/__tests__/select-options.js
```diff
@@ -107,3 +107,19 @@ test('throws an error if multiple are passed but not a multiple select', async (
...
+    <select disabled>
+      <option>No value selected</option>
+      <option value="1">1</option>
+      <option value="2">2</option>
+      <option value="3">3</option>
+    </select>,
+  )
+
+  userEvent.selectOptions(select, '1')
+  expect(select.selectedIndex).toBe(0)
```
kentcdodds, line 124: Could we make sure no events are fired? Should probably use the setupSelect function like the others.

nickmccurdy: Yea I will, just want to finish migrating things before continuing to fix this

PR #645, opened 2020-06-15
feat(userEvent): Add paste API (fixes #640)

### LINKED ISSUE #640: Replace type allAtOnce with "paste"
It's not aligned with the goals of this lib, but the use case is understandable so we should include a simple "paste" event.

nickmccurdy:
Migrated from testing-library/user-event#343

**What**: Add paste API (fixes #640)

**Why**: The `allAtOnce` option of `type` doesn't represent real user interaction intuitively, and we need to fire `paste` events with the relevant `clipboardData`

**How**:

- [x] Extract `paste` event from old code for `type`'s `allAtOnce` option
- [x] Implement `clipboardData` support #585
- [ ] Add events
  - [ ] `paste` with `clipboardData`
  - [ ] `input` with matching value
  - [ ] `focus`
- [ ] `DataTransfer` mock objects

**Checklist**:

- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/user-event/paste.js
+++ b/src/user-event/paste.js
@@ -1,0 +1,122 @@
+import {
+  getConfig as getDOMTestingLibraryConfig,
+  fireEvent,
+} from '@testing-library/dom'
+
+// this needs to be wrapped in the asyncWrapper for React's act and angular's change detection
+async function paste(...args) {
+  let result
+  await getDOMTestingLibraryConfig().asyncWrapper(async () => {
+    result = await pasteImpl(...args)
+  })
+  return result
+}
+
+const getActiveElement = document => {
+  const activeElement = document.activeElement
+  if (activeElement.shadowRoot) {
+    return getActiveElement(activeElement.shadowRoot) || activeElement
+  } else {
+    return activeElement
+  }
+}
+
+// eslint-disable-next-line complexity
+function pasteImpl(
+  element,
+  text,
+  {initialSelectionStart, initialSelectionEnd} = {},
+) {
+  if (element.disabled) return
+
+  element.focus()
+
+  // The focused element could change between each event, so get the currently active element each time
+  const currentElement = () => getActiveElement(element.ownerDocument)
+  const currentValue = () => currentElement().value
+  const setSelectionRange = ({newValue, newSelectionStart}) => {
+    // if we *can* change the selection start, then we will if the new value
+    // is the same as the current value (so it wasn't programatically changed
+    // when the fireEvent.input was triggered).
+    // The reason we have to do this at all is because it actually *is*
+    // programmatically changed by fireEvent.input, so we have to simulate the
+    // browser's default behavior
+    if (
+      currentElement().selectionStart !== null &&
+      currentValue() === newValue
+    ) {
+      currentElement().setSelectionRange?.(newSelectionStart, newSelectionStart)
+    }
+  }
+
+  // by default, a new element has it's selection start and end at 0
+  // but most of the time when people call "paste", they expect it to paste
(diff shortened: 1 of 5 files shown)


src/user-event/paste.js
```diff
@@ -0,0 +1,120 @@
+import {getConfig as getDOMTestingLibraryConfig} from '../config'
+import {fireEvent} from '../events'
+
```
kentcdodds, line 3: Let's see if we can move some of the shared utilities from type until the utils and share them here.

src/user-event/paste.js
```diff
@@ -0,0 +1,120 @@
+import {getConfig as getDOMTestingLibraryConfig} from '../config'
+import {fireEvent} from '../events'
```
kentcdodds, line 2: Make sure to use the fireEvent from the utils module because it's async which is required for user event methods.

src/user-event/paste.js
nickmccurdy: I've extracted some of the functions, let me know if you have any more suggestions for abstracting this further

src/user-event/paste.js
```diff
@@ -0,0 +1,71 @@
...
+import {fireEvent, getActiveElement, calculateNewValue} from './utils'
+
+// this needs to be wrapped in the asyncWrapper for React's act and angular's change detection
+async function paste(...args) {
+  let result
+  await getDOMTestingLibraryConfig().asyncWrapper(async () => {
+    result = await pasteImpl(...args)
+  })
+  return result
+}
```
nickmccurdy, line 11: I originally copied this from `type`, but I don't understand why `type` is no longer using `getConfig` and `asyncWrapper`. Should I keep this here, remove it, or use some higher level abstraction to make the function asynchronous?

nickmccurdy:
> I left some comments, but I just had a thought... What if instead we make a `{paste}` modifier inside `type`? This would simplify the implemention quite a bit and honestly makes plenty of sense as an API to me personally.

Shouldn't the typing modifiers be primarily for individual key presses and modifiers? I feel like pasting has different enough responsibilities from typing semantically to be a separate function, just like how clear and tab are functions rather than modifiers. Either way, I'm interested if we could maybe make paste use type more directly to simplify the architecture, as I originally though that wouldn't be possible after removing the `allAtOnce` option.

kentcdodds:
> Shouldn't the typing modifiers be primarily for individual key presses and modifiers?

Not really. We already have `{selectall}`. I think `type` is for anything that the user would do with their keyboard which we feel they do enough to justify having a feature for it. I think it would be reasonable to support `{paste}` and also have a shortcut `userEvent.paste` which simply calls into it (similar in some ways to `userEvent.clear`.

nickmccurdy:
It still feels like a different kind of user interaction to me. At least `{selectall}` is directly related to editing some existing text, but users often paste without typing anything else, and as far as I understand this would be the only command that would input many characters immediately.

kentcdodds:
Fair point 👍 ok, let's proceed with sharing code instead.

kentcdodds:
Ok, because of the [decision to not merge user-event right now](https://github.com/testing-library/dom-testing-library/pull/616#issuecomment-644246162), I've back-ported all my improvements to user-event into https://github.com/testing-library/user-event/pull/348, and I included this as well. I hope you don't mind 😬

Thanks!

nickmccurdy:
This wasn't ready to release, can we revert it?

kentcdodds:
I made it ready 😁

nickmccurdy:
Tracking followup work back in testing-library/user-event#355

Issue #609, opened 2020-06-07
" Unable to find the "window" object for the given node." when node is a Promise


JoshuaKGoldberg:
- `@testing-library/dom` version: 7.10.1
- Testing Framework and version: `jest`@26.0.1 via `@testing-library/react`@10.2.1
- DOM Environment: Jest defaults

Relevant code or config

```js
fireEvent.click(new Promise(jest.fn());
```

What you did:

This is a contrived example; what I actually did was pass the result of `findByRole` instead of `getByRole`.

What happened:

```
 Unable to find the "window" object for the given node. Please file an issue with the code that's causing you to see this error: https://github.com/testing-library/dom-testing-library/issues/new

      40 |       await act(async () => {
    > 41 |         fireEvent.click(jest.fn());
         |                   ^
      42 |       });

```

Problem description:

`getWindowFromNode` doesn't account for the `node` not being a `Node`.

Suggested solution:

Give an explicit error message if it's an instance of a `Promise`.

zzzachzzz:
Yep, I had the same exact issue the other day where I tried to `fireEvent.click(screen. findBy*(...))` instead of `fireEvent.click(screen. getBy*(...))`, completely by accident. It took me way too long to figure out the problem. :sweat_smile:

kentcdodds:
I'm good with this. Let's just check if it has a `.then` function and if it does we can give a helpful error message.

Though you may want to use typescript to avoid this 😉

kentcdodds:
This is an invitation to make a pull request 😁

timdeschryver:
I think we could add this to the eslint repo. 
If I recall correctly we already have some helpers to find a Promise, so this should be not too hard to do.

kentcdodds:
That would definitely be cool

PR #646, opened 2020-06-15
Added explicit error message for Promises passed to getWindowFromNode

JoshuaKGoldberg:
**What**:

Added a nice explicit error message for the case in #609 of a `Promise` or Promise-like passed to a function that expects a DOM node.

**Why**:

Better error messages = happier users! 💝 

**How**:

Per https://github.com/testing-library/dom-testing-library/issues/609#issuecomment-643660994, checked if `node.then instanceof Function`.

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

I'm not attached to this particular error message, if a reviewer can think of something better that'd be nice.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -47,7 +47,6 @@ function getDocument() {
   return window.document
 }
 function getWindowFromNode(node) {
-  // istanbul ignore next I'm not sure what could cause the final else so we'll leave it uncovered.
   if (node.defaultView) {
     // node is document
     return node.defaultView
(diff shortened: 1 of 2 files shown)


src/helpers.js
```diff
@@ -57,8 +56,12 @@ function getWindowFromNode(node) {
   } else if (node.window) {
     // node is window
     return node.window
+  } else if (node.then instanceof Function) {
+    throw new Error(
+      `It looks like you passed a Promise object instead of a DOM node. Did you use a "find" search instead of a "get" or "query"?`,
```
kentcdodds, line 61: Could we change this message to be a little more clear with:

```suggestion
      `It looks like you passed a Promise object instead of a DOM node. Did you do something like \`fireEvent.click(screen.findBy...\` when you meant to do \`fireEvent.click(await screen.findBy...\`?`,
```

kentcdodds:
@all-contributors please add @JoshuaKGoldberg for code and tests

Issue #612, opened 2020-06-08
Jest has changed its timer implementation causing waitFor() not to work properly


jasonwilliams:
- `@testing-library/dom` version: 7.5.2
- Testing Framework and version: 10.0.4
  
- DOM Environment: Jest
Jest version: 25.1.0
Node v12

We had some code which uses fakeTimers, I noticed it was screwing up `waitFor()`.
So i needed to run `jest.useRealTimers()` just before calling `waitFor()` to fix the issue.

This shouldn't be needed, as testing-library tries to safe-guard against fakeTimer use, you can see here:
https://github.com/testing-library/dom-testing-library/blob/master/src/helpers.js#L5-L8

However, i noticed this line will always return `undefined` regardless of whether you have fakeTimers on or not. This is due to the fact that `globalObj.setTimeout._isMockFunction` is always undefined.
My guess is Jest used to mock setTimeout using its own utilities and this property once existed, but now Jest uses [@Sinon/MockTimers](https://github.com/sinonjs/fake-timers/) where this property is not added.

## When did this happen?
Looks like the change happened in Jest v26
https://github.com/facebook/jest/releases/tag/v26.0.0

there were also implementation changes in v25.1
https://github.com/facebook/jest/releases?after=v25.5.1

`_isMockFunction` is added here:
https://github.com/facebook/jest/blob/master/packages/jest-fake-timers/src/legacyFakeTimers.ts#L370

It was swapped out in v25, plan here:
https://github.com/facebook/jest/pull/7776#issuecomment-513552169

## Solution
The fix is to find another way to see if the timers are being mocked or not, as the current heuristics don't work.

zzzachzzz:
Awesome write-up. I also encountered some strange behavior when using Jest fakeTimers and trying to add a typing delay to `userEvents.type`, from [testing-library/user-event](https://github.com/testing-library/user-event).

kentcdodds:
Thank you for the detail! We could definitely fix this. I invite anyone to help figure out how to solve this. 👍

kgregory:
@kentcdodds I think [this condition (below)](https://github.com/testing-library/dom-testing-library/commit/167b4acc304b554f03cd16456b63001e9aa4adf1#diff-b38f7254be1a9f817fd2e33c294481a1R7-R8) will be true for someone using `@sinonjs/fake-timers` instead of `jest.useFakeTimers` because the timer functions will have a `clock` attribute in either case (if useFakeTimers was called with 'modern'):

[...]

PR #649, opened 2020-06-15
fix: runWithRealTimers to be compatible with new version of jest

marcosvega91:
**What**:

This change will fix issue #612.

**Why**:

With the version 26.0.0 of jest the `jest.useFakeTimers` function uses `@sinonjs/fake-timers`.
With this new package the property `_isMockFunction` is no more available.

**How**:

To fix the issue I have added a new check to `clock` property.
`@testing-library/dom` is now compatible with both version of `useFakeTimers`, the legacy and the modern one.

**Checklist**:

- [X] Tests
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/helpers.js
+++ b/src/__tests__/helpers.js
@@ -1,4 +1,11 @@
-import {getDocument, getWindowFromNode, checkContainerType} from '../helpers'
+import {
+  getDocument,
+  getWindowFromNode,
+  checkContainerType,
+  runWithRealTimers,
+} from '../helpers'
+
+const globalObj = typeof window === 'undefined' ? global : window
 
 test('returns global document if exists', () => {
   expect(getDocument()).toBe(document)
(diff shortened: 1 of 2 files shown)


src/__tests__/helpers.js
```diff
@@ -43,3 +50,19 @@ describe('query container validation throws when validation fails', () => {
...
+  const originalSetTimeout = globalObj.setTimeout
+
+  jest.useFakeTimers('legacy')
+  runWithRealTimers(() => {
+    expect(originalSetTimeout).toEqual(globalObj.setTimeout)
+  })
+
+  jest.useRealTimers()
+
+  jest.useFakeTimers('modern')
```
kentcdodds, line 64: Because of these tests, we'll need an `afterEach` that does `jest.useRealTimers()` to keep this isolated.

src/helpers.js
```diff
@@ -4,7 +4,8 @@ const globalObj = typeof window === 'undefined' ? global : window
 function runWithRealTimers(callback) {
   const usingJestFakeTimers =
     globalObj.setTimeout &&
-    globalObj.setTimeout._isMockFunction &&
+    (globalObj.setTimeout._isMockFunction ||
+      typeof globalObj.setTimeout.clock !== 'undefined') &&
```
kentcdodds, line 8: Sensible 👍

src/__tests__/helpers.js
marcosvega91: totally right I forgot it

src/helpers.js
marcosvega91: I have tried to find a different way but I didn't find

PR #655, opened 2020-06-20
fix: improve suggested message, can't await getBy

timdeschryver:
**What**:

- Improve suggestion when a Promise is passed to a `fireEvent` method

**Why**:

- A `getBy` query can't be awaited and the messages says `await screen.getBy`

**How**:

- Change the message (and the test)

**Checklist**:

- [x] Tests
- [x] Ready to be merged


timdeschryver:
If needed, I can undo the style changes - but it seems like it's correctly formatted now?

kentcdodds:
Thanks!

PR #662, opened 2020-06-23
feat(waitFor): add complete and transparent support for fake timers

### LINKED ISSUE #661: waitFor should work with fake timers
### Relevant code or config:

```js
const {waitFor} = require('@testing-library/dom')

beforeEach(() => jest.useFakeTimers())
afterEach(() => jest.useRealTimers())

test('example', async () => {
  const doAsyncThing = () =>
    new Promise((r) => setTimeout(() => r('data'), 300))
  let result
  doAsyncThing().then((r) => (result = r))

  await waitFor(() => expect(result).toBe('data'))
})
```

### What you did:

I expected `waitFor` to work even with jest fake timers enabled

### What happened:

It times out because there's nothing telling jest to advance timers

### Reproduction:

https://github.com/kentcdodds/testing-library-fake-timers

### Problem description:

When enabling fake timers with DOM Testing Library, the `waitFor` utility doesn't work (so none of the other async utils will either). The reason for this is there's nothing telling the clock to advance the timers.

### Suggested solution:

I'm pretty sure we actually *do* want the async utils to work seamlessly when fake timers are enabled.

Here's a `waitFor` function that will work only if fake timers are enabled. Maybe we can detect fake timers (we already have a function for this) and if they're enabled, then we could do this instead?

```javascript
async function waitForWithFakeTimers(cb) {
  let waiting = true
  while (waiting) {
    await act(() =>
      Promise.resolve()
        .then(() => jest.runAllTimers())
        .then(() => new Promise(resolve => setImmediate(resolve))),
    )
    try {
      cb()
      waiting = false
    } catch {}
  }
}
```

It's pretty basic, it doesn't support timeout etc. But it's a good start.

**Mostly, I'm looking for feedback**. Should we do something like this if fake timers are enabled? Or is there another approach we can take?

kentcdodds:
**What**: feat(waitFor): add complete and transparent support for fake timers

**Why**: Closes #661

**How**:

Basically, we detect if we're in an environment that's faking out timers and if we are then we do a `while` loop that advances the jest fake timers by the interval until the callback passes. We do still support a timeout so it can time out eventually and won't loop forever.

**Checklist**:

~- [ ] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs) N/A~
- [x] Tests
~- [ ] Typescript definitions updated N/A~
- [x] Ready to be merged


tannerlinsley:
Elegant :)

kentcdodds:
Build is busted thanks to: https://github.com/testing-library/dom-testing-library/issues/663

I'm planning on doing a beta release of this locally so I can test it out.

src/__tests__/wait-for-element-to-be-removed.js
```diff
@@ -89,55 +78,6 @@ test('after successful removal, fullfills promise with empty value (undefined)',
...
-  })
-})
-
-test("doesn't change jest's timers value when importing the module", () => {
-  jest.useFakeTimers()
-  importModule()
-
-  expect(window.setTimeout._isMockFunction).toEqual(true)
-})
-
```
kentcdodds, line 140: We don't need these tests because `waitForElementToBeRemoved` is built on top of `waitFor` and it has its own tests.

src/helpers.js
```diff
@@ -117,4 +125,5 @@ export {
   setTimeoutFn as setTimeout,
   runWithRealTimers,
   checkContainerType,
+  jestFakeTimersAreEnabled,
```
kentcdodds, line 128: Pretty much all the changes in this file are refactors to facilitate this export.

kentcdodds:
This is published as `@testing-library/dom@7.16.2-alpha.1`

kentcdodds:
And I published these changes in `@testing-library/react@10.3.1-alpha.1` to try it out in a practical situation.

kentcdodds:
Works like a charm! :chefs-kiss:

kentcdodds:
Dang it.

> Ran 100000 timers, and there are still more! Assuming we've hit an infinite recursion and bailing out...

🤦‍♂️  investigating....

src/wait-for.js
```diff
@@ -35,22 +36,50 @@ function waitFor(
...
+
+    const usingFakeTimers = jestFakeTimersAreEnabled()
+    if (usingFakeTimers) {
+      // this is a dangerous rule to disable because it could lead to an
+      // infinite loop. However, eslint isn't smart enough to know that we're
+      // setting finished inside `onDone` which will be called when we're done
+      // waiting or when we've timed out.
+      // eslint-disable-next-line no-unmodified-loop-condition
+      while (!finished) {
+        jest.runAllTimers()
```
kentcdodds, line 56: Whoops! This was supposed to be `jest.advanceTimersByTime(interval)` 😅

kentcdodds:
I think this is ready. I made a handful of changes to how we test some of these utilities (especially the deprecated ones).

src/wait-for.js
```diff
@@ -34,23 +37,46 @@ function waitFor(
...
+
     const overallTimeoutTimer = setTimeout(onTimeout, timeout)
-    const intervalId = setInterval(checkCallback, interval)
 
-    const {MutationObserver} = getWindowFromNode(container)
-    const observer = new MutationObserver(checkCallback)
-    runWithRealTimers(() =>
-      observer.observe(container, mutationObserverOptions),
-    )
-    checkCallback()
```
kentcdodds, line 48: We no longer need to run this with real timers because we only run it when we're not using fake timers.

kentcdodds:
Note: I can't guarantee that these changes won't break people who had to work around the limitations in their tests. So if someone was using our async utilities with fake timers, it's possible that we'll break them depending on how they worked around things.

I'll add this to the release notes, but if you're here reading this because your tests broke as a result of this change, just know that you shouldn't have to worry about advancing the fake timers manually anymore.

PR #664, opened 2020-06-23
fix: Don't recommend ByRole if role is `generic`

### LINKED ISSUE #663: Fix default branch CI build
I think when I merged [this PR](https://github.com/testing-library/dom-testing-library/pull/656) I had assumed the build passed because of [codecov's comment](https://github.com/testing-library/dom-testing-library/pull/656#issuecomment-647007727) which shouldn't appear until we have a successful build. I'm still not sure what happened there to be honest.

But in any case, since that PR was merged (which really just updated `aria-query`) our tests have been failing. Mostly the suggestion tests, but some others as well.

Could someone dive in and help. You don't have to do it all. Even fixing a few would be helpful.

I think we could especially use help from @benmonro who wrote most of the suggestion tests. In particular, aria-query now gives us a role of `generic` for generic elements. I'm not 100% certain I want to recommend `screen.getByRole('generic', {name: /bob/i})` over `screen.getByText(/bob/i)`... So I'm thinking we should probably skip that suggestion if the role is `generic`.

marcosvega91:
fix #663 
**What**:

I have fixed all the tests that failed

**Why**:

Because after the update of `aria-query` something didn't work

**How**:

I have made two changes in this PR:

1) With the new aria specifications `input` without a role has a `textbox` as default one.
    I' have replaced all the input used in tests that not require a role with input of `type=password` that doesn't have a 
   default role @benmonro 

2) I have skipped generic role because we prefer to suggest others query 
    we prefer `screen.getByText(/bob/i)` instead of `screen.getByRole('generic', {name: /bob/i})`
 

**Checklist**:

- [X] Tests
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -533,14 +533,14 @@ test('queryAllByRole returns semantic html elements', () => {
   expect(queryAllByRole(/listitem/i)).toHaveLength(3)
   // TODO: with https://github.com/A11yance/aria-query/pull/42
   // the actual value will match `toHaveLength(2)`
-  expect(queryAllByRole(/textbox/i)).toHaveLength(1)
+  expect(queryAllByRole(/textbox/i)).toHaveLength(2)
   expect(queryAllByRole(/checkbox/i)).toHaveLength(1)
   expect(queryAllByRole(/radio/i)).toHaveLength(1)
   expect(queryAllByRole('row')).toHaveLength(3)
   expect(queryAllByRole(/rowgroup/i)).toHaveLength(2)
   // TODO: with https://github.com/A11yance/aria-query/pull/42
   // the actual value will match `toHaveLength(3)`
-  expect(queryAllByRole(/(table)|(textbox)/i)).toHaveLength(2)
+  expect(queryAllByRole(/(table)|(textbox)/i)).toHaveLength(3)
   expect(queryAllByRole(/img/i)).toHaveLength(1)
   expect(queryAllByRole('meter')).toHaveLength(1)
   expect(queryAllByRole('progressbar')).toHaveLength(0)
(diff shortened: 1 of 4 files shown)


src/__tests__/suggestions.js
```diff
@@ -74,7 +74,9 @@ test(`should not suggest if the suggestion would give different results`, () =>
 })
 
 test('should suggest by label over title', () => {
-  renderIntoDocument(`<label><span>bar</span><input title="foo" /></label>`)
+  renderIntoDocument(
+    `<label><span>bar</span><input type="password" title="foo" /></label>`,
```
eps1lon, line 78: To be accurate: The spec related to aria didn't change. This was a bug in earlier versions of `aria-query`. `<input />` with no or an invalid `type` attribute is always considered `<input type="text" />` which is a `role="textbox"`.

src/__tests__/element-queries.js
```diff
@@ -533,14 +533,14 @@ test('queryAllByRole returns semantic html elements', () => {
...
   // TODO: with https://github.com/A11yance/aria-query/pull/42
   // the actual value will match `toHaveLength(2)`
-  expect(queryAllByRole(/textbox/i)).toHaveLength(1)
+  expect(queryAllByRole(/textbox/i)).toHaveLength(2)
   expect(queryAllByRole(/checkbox/i)).toHaveLength(1)
   expect(queryAllByRole(/radio/i)).toHaveLength(1)
   expect(queryAllByRole('row')).toHaveLength(3)
   expect(queryAllByRole(/rowgroup/i)).toHaveLength(2)
   // TODO: with https://github.com/A11yance/aria-query/pull/42
   // the actual value will match `toHaveLength(3)`
```
afontcu, line 542: I guess these comments (this one and the one in lines 534-535) can be removed now 🎉

eps1lon: https://github.com/testing-library/dom-testing-library/pull/665

kentcdodds:
@marcosvega91, and all reviewers... You are my heros!

I really didn't have time to work on this. Thank you!

PR #667, opened 2020-06-23
fix(waitFor): handle odd timing issue with fake timers

kentcdodds:
**What**: fix(waitFor): handle odd timing issue with fake timers

**Why**: It's super hard to explain and I honestly don't really understand the issue completely myself. All I know is, before moving the `checkCallback` function call above the promise flush line, I had issues with tests impacting each other. After this change, that issue goes away.

I also added a comment explaining why we shouldn't use `jest.advanceTimersToNextTimer` along with a test to demonstrate the issue.

**How**: Change the order of operations

**Checklist**:

~- [ ] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs) N/A~
- [x] Tests
~- [ ] Typescript definitions updated N/A~
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/fake-timers.js
+++ b/src/__tests__/fake-timers.js
@@ -9,14 +9,14 @@ afterAll(() => {
   jest.useRealTimers()
 })
 
-async function runWaitFor() {
+async function runWaitFor({time = 300} = {}, options) {
   const response = 'data'
   const doAsyncThing = () =>
-    new Promise(r => setTimeout(() => r(response), 300))
+    new Promise(r => setTimeout(() => r(response), time))
   let result
   doAsyncThing().then(r => (result = r))
 
-  await waitFor(() => expect(result).toBe(response))
+  await waitFor(() => expect(result).toBe(response), options)
 }
 
 test('real timers', async () => {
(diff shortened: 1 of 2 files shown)


eps1lon:
> All I know is, before moving the checkCallback function call above the promise flush line, I had issues with tests impacting each other. After this change, that issue goes away.

@kentcdodds Do you remember which tests were impacting each other? It's totally fine if this only reproduces in a full test suite. But some repro would be nice since this change is quite odd considering we're scheduling more work from `waitFor` **after** the `waitFor` promise is resolved.

kentcdodds:
If this change is causing issues, then let's change it back. I'm pretty sure this was in [the bookshelf app](https://github.com/kentcdodds/bookshelf), but I don't have time to reproduce things and the latest version of react-query eliminates all the problems that I experienced with those tests anyway so we're probably ok to revert this.

eps1lon:
https://github.com/testing-library/dom-testing-library/pull/1073 has a test that breaks with the behavior introduced in this PR (https://github.com/testing-library/dom-testing-library/pull/667).

I can fix the problematic behavior with a less invasive fix. But I can also just revert this change. I'm more comfortable with just reverting this change.

In any case, I'll try both fixes in the bookshelf app so reduce likelyhood of unintended breakage. Thanks for the pointer!

eps1lon:
See https://github.com/testing-library/dom-testing-library/pull/1073#issuecomment-965627327 for conclusion of the thread started with https://github.com/testing-library/dom-testing-library/pull/667#issuecomment-965091530

PR #668, opened 2020-06-24
fix: add getConfig to types

timdeschryver:
**What**:

- adds the `getConfig` type to the TS typings

**Why**:

- otherwise it seems that `getConfig` isn't exported

**How**:

- add it to type types file

**Checklist**:

- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/types/config.d.ts
+++ b/types/config.d.ts
@@ -1,14 +1,15 @@
 export interface Config {
-    testIdAttribute: string;
-    asyncWrapper(cb: (...args: any[]) => any): Promise<any>;
-    eventWrapper(cb: (...args: any[]) => any): void;
-    asyncUtilTimeout: number;
-    defaultHidden: boolean;
-    throwSuggestions: boolean;
+  testIdAttribute: string
+  asyncWrapper(cb: (...args: any[]) => any): Promise<any>
+  eventWrapper(cb: (...args: any[]) => any): void
+  asyncUtilTimeout: number
+  defaultHidden: boolean
+  throwSuggestions: boolean
 }
 
 export interface ConfigFn {
-    (existingConfig: Config): Partial<Config>;
+  (existingConfig: Config): Partial<Config>
 }
 
-export function configure(configDelta: Partial<Config> | ConfigFn): void;
+export function configure(configDelta: Partial<Config> | ConfigFn): void
+export function getConfig(): Config


timdeschryver:
I was looking into extending the config for Angular Testing Library, to add default imports. This way, Angular Testing Library retrieves the config with `getConfig` to include these default imports to the Dependency Injection.

For example:

```ts
beforeEach(() => {
  configure({
    defaultImports: [ReactiveFormsModule],
  });
});

test('test 1', async () => {
 await render(MyComponent)
})

test('test 2', async () => {
 await render(MyComponent)
})

// VS

test('test 1', async () => {
 await render(MyComponent, { imports: [ReactiveFormsModule] })
})

test('test 2', async () => {
 await render(MyComponent, { imports: [ReactiveFormsModule] })
})
```

An alternative could be that Angular Testing Library provides its own `angularConfigure` method?
What do you think?

kentcdodds:
Interesting. I'm not sure you need this method to do that. You could simply make your own configure wrapper that takes custom config properties and forwards along the rest. I think it would be dangerous to set the DOM Testing Library config with properties it doesn't support.

That said, we use getConfig in user-event so it's not just exposed by accident and should probably be typed for that reason.

eps1lon:
> I was looking into extending the config for Angular Testing Library, to add default imports.

I'd be careful with adding additional properties to the config object from `@testing-library/dom`. This could create conflicts if we decide to add properties with names you're already using.

Have you considered implementing your own `configure` that forwards excess properties to the `configure` from DTL and has a separate property for the config of the angular testing library (or vice-versa) e.g.

```js
import { configure } from '@testing-library/angular';

configure({ property: 'foo', someFlag: false, dom: domTestingLibraryConfig })
```

I guess this is currently problematic for any library extending the config of DTL.

> That said, we use getConfig in user-event so it's not just exposed by accident and should probably be typed for that reason.

A library using a property does not justify that something is public. You could reach into all sorts of internals. I want to make sure this is actually intended on our side and why considering we don't have any documentation for it.

timdeschryver:
I have not, but I think I'm going to go with your proposition @eps1lon 👍.

The current implementation looks like this, but I'm not too happy about it.

```ts
function configure(config: Partial<Config>) {
  dtlConfigure({
    defaultImports: config.defaultImports,
  } as Partial<dtlConfig>);
}
```


> That said, we use getConfig in user-event so it's not just exposed by accident and should probably be typed for that reason.

@kentcdodds feel free to close or merge this PR.

PR #670, opened 2020-06-24
feat: add allowed roles to role query types

### LINKED ISSUE #666: Intellisense with getByRole

I got the feeling that we could make it easier to use the `getByRole` queries.
From what I can see not all developers know (all of) the WAI-ARIA Roles.
The suggestions and the playground are definitely helping, but I think it would be nice if we can type the roles so that the IDE can pick them up and suggest them (shorter feedback loop).

### Suggested implementation:

Add a `ByRoleMatcher` that includes all of the ARIA Roles, and use this type in the `byRole` queries.

```ts
export type ByRoleMatcher =  'button' | 'listitem' | 'textbox' | String | RegExp | MatcherFunction

export type AllByRole = (container: HTMLElement, role: ByRoleMatcher, options?: ByRoleOptions) => HTMLElement[];
export type GetByRole = (container: HTMLElement, role: ByRoleMatcher, options?: ByRoleOptions) => HTMLElement;
export type QueryByRole = (container: HTMLElement, role: ByRoleMatcher, options?: ByRoleOptions) => HTMLElement | null;
```

This won't be a breaking change because the user would still have the possibility to use a role that isn't included in the type.

This looks like this:

image

It's important to use `String` while defining the type, otherwise it will either result that the passed role must be one of the options, or that there won't be any intellisense. That's also the reason why it isn't possible to reuse the existing `Matcher` type (e.g. `export type ByRoleMatcher = Matcher | 'button' | 'listitem' | 'textbox'`)

See the [TS playground](https://www.typescriptlang.org/play/?ssl=18&ssc=21&pln=18&pc=12#code/PTAEAsEMGdQSwHYBcCmAbNdoodgNKAPYJoCeokAxpSgA5KzEpEBmoS4z0SATogOagArgjjFoAKBYjKSMQlDRIpABLo0hAIwAKRLSFIAXKADknDIROgAPqYBGxAFaEhPK7bOQLJgJSgA3hKgoJTihGgoAHQa-LoI+kg+ANwSAL4SEiAQMKAIhPDI6lg4+BTUdAwUGIq8AqAAbl5CKJLSCLLyispqFgBMcQnGZuqWNvZOLm5jnt5j3HwI-H6BwaG44VExAwbJaRlZULCIqBjFuCgEVDT0VWg1C4KNaM2tMnLEXaojAMzbRqb8QiEAAmdlIKHcpkgQlAPBQ9UIcCmHiUzyQkIAyrVFssgiEwhFooRYnodil0hIlF8LDoTPwDDh2JB+L4UlSehp+nSGQokMzWZTuj9tNzULz+bsJEA) for examples.

timdeschryver:
**What**:

- adds a string union type for supported aria roles
- closes #666 

**Why**:

- shorten the feedback loop for a better developer experience

**How**:

- Used the aria roles from [aria-query/ariaLiteralRoles](https://github.com/A11yance/aria-query/blob/main/src/etc/roles/ariaLiteralRoles.js#L87), [aria-query/ariaAbstractRoles](https://github.com/A11yance/aria-query/blob/main/src/etc/roles/ariaAbstractRoles.js#L17), [aria-query/ariaDpubRoles](https://github.com/A11yance/aria-query/blob/main/src/etc/roles/ariaDpubRoles.js#L44)

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

@eps1lon is there already a aria roles union type available that could be reused?


types/matches.d.ts
```diff
@@ -1,6 +1,145 @@
 export type MatcherFunction = (content: string, element: HTMLElement) => boolean
 export type Matcher = string | RegExp | MatcherFunction
 
+// important to use String here
+// this gives us intellisense but it can also accept values
+// that are not included in the union types
+export type ByRoleMatcher = AllowedRoles | String | RegExp | MatcherFunction
+export type AllowedRoles =
```
kentcdodds, line 8: Should we include a link for the source of these so if there are new ones added we know where to go to get them?

eps1lon: We should leverage `@types/aria-query` instead. This would automatically sync the types with the supported version of aria.

eps1lon: Though we might want to wait for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45657.

Someone else could re-open that one and add themselves as an owner. Then we can fast track it as soon as I approve it 😄

timdeschryver: Great, thanks @eps1lon .

eps1lon: Would still help if someone could review DefinitelyTyped/DefinitelyTyped#45657 since we'll be using those changes in the end.

eps1lon: `@types/aria-query@^4.2.0` should have all the roles for aria 1.2

types/matches.d.ts
```diff
@@ -1,6 +1,14 @@
+import { ARIARole } from 'aria-query'
+
 export type MatcherFunction = (content: string, element: HTMLElement) => boolean
 export type Matcher = string | RegExp | MatcherFunction
 
+// important to use String here
+// this gives us intellisense but it can also accept values
+// that are not included in the union types
+// tslint:disable-next-line:ban-types
+export type ByRoleMatcher = ARIARole | String | RegExp | MatcherFunction
```
eps1lon, line 10: Does this work with `string` as well?

```suggestion
export type ByRoleMatcher = ARIARole | string | RegExp | MatcherFunction
```

The comment explains that `String` is important but it's not obvious from the comment why `string` wouldn't work. This would be the more "natural" choice and get rid of the tslint warning.

timdeschryver: `string` doesn't give us the intellisense we want:

As `String`:

image

As `string`:

image

timdeschryver: Your comment made me thinking, and I changed the type implementation in the last commit.
WDYT?

It has the same result

image

timdeschryver: Seems like the linter doesn't like this ...

kentcdodds:
Seems legit. Looks like travis is failing due to some linting issues in the types?

timdeschryver:
You're right @kentcdodds .
Both approaches aren't making the linter happy, let's wait on @eps1lon to see which approach he likes more.
Personally, I think this approach is more clean than the first attempt.

types/tslint.json
```diff
@@ -3,6 +3,7 @@
   "rules": {
     "no-useless-files": false,
     "no-relative-import-in-test": false,
-    "semicolon": false
+    "semicolon": false,
+    "whitespace": false
```
timdeschryver, line 7: Disabled whitespace checks as this clashes with our prettier configuration.

timdeschryver:
Should be good to go on my end, can you have another look at it @eps1lon and share your thoughts?

eps1lon:
@timdeschryver Nice work!

PR #671, opened 2020-06-24
feat(waitFor): add onTimeout which adds DOM output to timeout errors

### LINKED ISSUE #559: automatically call screen.debug() if there's a timeout in waitFor
This would be particularly nice for `waitForElementToBeRemoved` (which would benefit from this being added to `waitFor`) because then people could easily see what the DOM looks like and maybe why the element has not been removed.

kentcdodds:
**What**: add onTimeout which adds DOM output to timeout errors

**Why**: Closes #559 (nicer error messages)

**How**: add the parameter and have it use prettyDOM to attach the nice message to the error

**Checklist**:

- [x] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs): https://github.com/testing-library/testing-library-docs/pull/512
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/base-queries-warn-on-invalid-container.js
+++ b/src/__tests__/base-queries-warn-on-invalid-container.js
@@ -112,7 +112,7 @@ describe('asynchronous queries throw on invalid container type', () => {
     ['findAllByTestId', findAllByTestId],
   ])('%s', (_queryName, query) => {
     const queryOptions = {}
-    const waitOptions = {timeout: 1}
+    const waitOptions = {timeout: 1, onTimeout: e => e}
     return expect(
       query(
         'invalid type for container',
--- a/src/wait-for.js
+++ b/src/wait-for.js
@@ -10,6 +10,7 @@ import {
   clearTimeout,
 } from './helpers'
 import {getConfig, runWithExpensiveErrorDiagnosticsDisabled} from './config'
+import {prettyDOM} from './pretty-dom'
 
 // This is so the stack trace the developer sees is one that's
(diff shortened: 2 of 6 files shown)


src/__tests__/base-queries-warn-on-invalid-container.js
```diff
@@ -112,7 +112,7 @@ describe('asynchronous queries throw on invalid container type', () => {
     ['findAllByTestId', findAllByTestId],
   ])('%s', (_queryName, query) => {
     const queryOptions = {}
-    const waitOptions = {timeout: 1}
+    const waitOptions = {timeout: 1, onTimeout: e => e}
```
kentcdodds, line 115: I added `onTimeout: e => e` to several places in the tests because otherwise the snapshots got formatted really weird and would fail due to formatting issues 🤷‍♂️

So I added a special test to handle this case specifically and the rest just do nothing to change the error.

src/wait-for.js
```diff
@@ -25,6 +26,10 @@ function waitFor(
     showOriginalStackTrace = getConfig().showOriginalStackTrace,
     stackTraceError,
     interval = 50,
+    onTimeout = error => {
+      error.message = `${error.message}\n\n${prettyDOM(container)}`
```
kentcdodds, line 30: The reason we do this rather than simply calling `screen.debug()` here:

1. It could be that the `container` isn't bound to the same `document` as `screen`, so we should use the container given
2. If we do `screen.debug()` then the DOM output will appear above the error which I think would be confusing.

PR #672, opened 2020-06-25
fix: use the first label for LabelText query suggestions

marcosvega91:
**What**:  Fix a little bug 

**Why**:  As reported [here](https://github.com/testing-library/testing-playground/issues/214) when `getSuggestedQuery` tried to get label text through `aria-labelledby` it threw an exception when the id had spaces or special chars

**How**: Surrounding the selector with `" "`

**Checklist**:

- [X] Tests
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/suggestions.js
+++ b/src/__tests__/suggestions.js
@@ -485,3 +485,29 @@ test('getSuggestedQuery does not create suggestions for script and style element
   expect(getSuggestedQuery(script, 'get', 'TestId')).toBeUndefined()
   expect(getSuggestedQuery(style, 'get', 'TestId')).toBeUndefined()
 })
+
+test('should not thrown when getting query byLabelText if ID has special chars or spaces', () => {
+  const {container, rerender} = renderIntoDocument(`
+    <div id="not usual id">One</div>
+    <input
+      type="text"
+      aria-labelledby="not usual id"
+    />
+  `)
+
+  expect(() =>
+    getSuggestedQuery(container.querySelector('input'), 'get', 'labelText'),
+  ).not.toThrow()
+
+  rerender(`
+    <div id="why you need to use \\/|"£$ ?">One</div>
+    <input
+      type="text"
+      aria-labelledby="why you need to use \\/|"£$ ?"
+    />
+  `)
+
+  expect(() =>
+    getSuggestedQuery(container.querySelector('input'), 'get', 'labelText'),
+  ).not.toThrow()
+})
--- a/src/__tests__/suggestions.js
+++ b/src/__tests__/suggestions.js
@@ -486,28 +486,25 @@ test('getSuggestedQuery does not create suggestions for script and style element
   expect(getSuggestedQuery(style, 'get', 'TestId')).toBeUndefined()
 })
 
-test('should not thrown when getting query byLabelText if ID has special chars or spaces', () => {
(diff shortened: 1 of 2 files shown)


marcosvega91:
yes about spec you're right.

I don't know which solution is better then the other, both solution are not correct because of #545

smeijer:
True, both fixes are just hacks to stop the query from throwing errors, while making suggestions.

The difference is that that by wrapping the query in quotes, the query won't find a matching element, and thereby won't make a suggestion. It feels wrong to me to build a selector that we know is incorrect and won't return any elements.

By taking only the first id, the query still finds a matching element and will make a valid suggestion. It won't give the best suggestion possible, but at least it will suggest something close that does return an element.

The best fix would be to contact the labels, but as I understand, that isn't currently supported?

If the partial match doesn't result in a valid query either (I haven't tested it), I propose to just return `undefined` instead:

```js
const ariaLabelledBy = element.getAttribute('aria-labelledby')
    if (ariaLabelledBy) {
      // compound labels are currently unsupported, fall back to different query options
      label = ariaLabelledBy.includes(' ') ? undefined : document.querySelector(`[id="${ariaLabelledBy}"]`)
    }
```


<details>
 <summary>concat labels</summary>

```js
function getLabelTextFor(element) {
  let label =
    element.labels &&
    Array.from(element.labels).find(el => Boolean(normalize(el.textContent)))

  if (label) {
    return label;
  }

  // non form elements that are using aria-labelledby won't be included in `element.labels`
  const ariaLabelledBy = element.getAttribute('aria-labelledby');

  if (ariaLabelledBy) {
    const selector = ariaLabelledBy.split(' ').map(id => `[id="${id}"]`).join(', ');
    const labels = document.querySelectorAll(selector);
    return Array.from(labels).map(label => label.textContent).join(' ')
  }


  return undefined
}
```

</details>

marcosvega91:
Yes the function should work even if the logic is more complex, look at [this](https://github.com/chromium/chromium/blob/582ef42acc19d347fe7c96f28d3dfc502f7e5f62/ui/accessibility/extensions/chromevoxclassic/common/dom_util.js#L468)

The problem then is that the query will not work because `screen. getByLabelText` not support label concatenation.
@delca85 created a PR for the (issue)[https://github.com/testing-library/dom-testing-library/pull/607]

smeijer:
Yes, I see. So my [proposal](https://github.com/testing-library/dom-testing-library/pull/672#pullrequestreview-437203095), to just only take the first label and use that instead. Sounds like a good fix to me.

Let's take the markup:

```html
<div id="a">One</div>
<div id="b">Two</div>
<input
  type="text"
  aria-labelledby="a b"
/>
```

The full label is `One Two`.

I see 3 options:

1. Combine the labels to a single label, and use that. But this isn't supported (see #607)

2. Wrap the query in quotes, as this PR is currently doing. No element is matched ('id["a b"]' doesn't exist), so no suggestion is being made. Potentially resulting in suggesting a query that's a worse option (testids?), as a `getByLabelText` couldn't be created.

3. Only use the first label to generate the suggestion. This would result in a query suggestion like: `screen.getByLabelText(/one/i)`. This is a valid query, that will result in returning the `input`.

As 1 isn't currently possible, that's not a real option. So only 2 and 3 are left.

marcosvega91:
Yes I can implement the third option.

Issue #448, opened 2020-02-04
querying aria-role superclass hierarchies


alexkrolick:
`getByRole` should be able to understand superclass hierarchies such as:

```
input      [abstract, debatable whether this should be queryable?]
  checkbox
    switch   (should be able to get switch by switch, checkbox, or input)
  option
  radio
  slider
  spinbutton
  textbox
```

Looks related to #446 

Examples: https://codesandbox.io/s/young-tree-68kbl

kentcdodds:
Not sure whether to label this as a bug or enhancement, but either way, this would be great :)

kentcdodds:
Actually, I feel like this should be something that aria-query or dom-accessibility-api should handle 🤔

It's definitely an enhancement though :)

alexkrolick:
Sounds like aria-query provides this functionality

delca85:
If it sounds good to all of you, I could try to make this enhancement!

kentcdodds:
That sounds great to me @delca85!

kentcdodds:
We've decided to not implement this.

PR #675, opened 2020-06-27
Querying aria role superclass

delca85:
**What**:

`*ByRole` queries are now able to retrieve roles according to super class roles hierarchies.

**Why**:

According to #448 is a needed enhancement.

**How**:

- add an helper method to retrieve super class roles given a role
- use this helper method in `queryAllByRole`
- fix previous tests

**Checklist**:

- [x] add an helper method to retrieve super class roles given a role
- [x] use this helper method in `queryAllByRole`
- [x] fix previous tests

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/ariaAttributes.js
+++ b/src/__tests__/ariaAttributes.js
@@ -63,12 +63,17 @@ test('`selected: true` matches `aria-selected="true"` on supported roles', () =>
   ).toEqual(['selected-native-columnheader', 'selected-columnheader'])
 
   expect(getAllByRole('gridcell', {selected: true}).map(({id}) => id)).toEqual([
+    'selected-rowheader',
+    'selected-native-columnheader',
+    'selected-columnheader',
+    'selected-native-rowheader',
     'selected-gridcell',
   ])
 
   expect(getAllByRole('option', {selected: true}).map(({id}) => id)).toEqual([
     'selected-native-option',
     'selected-listbox-option',
+    'selected-treeitem',
   ])
 
   expect(
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -521,11 +521,11 @@ test('queryAllByRole returns semantic html elements', () => {
     </form>
   `)
 
-  expect(queryAllByRole(/table/i)).toHaveLength(1)
-  expect(queryAllByRole(/tabl/i, {exact: false})).toHaveLength(1)
+  expect(queryAllByRole(/table/i)).toHaveLength(2)
+  expect(queryAllByRole(/tabl/i, {exact: false})).toHaveLength(2)
   expect(queryAllByRole(/columnheader/i)).toHaveLength(1)
   expect(queryAllByRole(/rowheader/i)).toHaveLength(1)
-  expect(queryAllByRole(/grid/i)).toHaveLength(1)
+  expect(queryAllByRole(/grid/i)).toHaveLength(3)
   expect(queryAllByRole(/form/i)).toHaveLength(0)
   expect(queryAllByRole(/button/i)).toHaveLength(1)
   expect(queryAllByRole(/heading/i)).toHaveLength(6)
@@ -536,7 +536,7 @@ test('queryAllByRole returns semantic html elements', () => {
   expect(queryAllByRole(/radio/i)).toHaveLength(1)
   expect(queryAllByRole('row')).toHaveLength(3)
   expect(queryAllByRole(/rowgroup/i)).toHaveLength(2)
-  expect(queryAllByRole(/(table)|(textbox)/i)).toHaveLength(3)
+  expect(queryAllByRole(/(table)|(textbox)/i)).toHaveLength(4)
   expect(queryAllByRole(/img/i)).toHaveLength(1)
   expect(queryAllByRole('meter')).toHaveLength(1)
   expect(queryAllByRole('progressbar')).toHaveLength(0)
--- a/src/__tests__/role-helpers.js
+++ b/src/__tests__/role-helpers.js
@@ -3,6 +3,7 @@ import {
   logRoles,
   getImplicitAriaRoles,
   isInaccessible,
+  getRolesAndSuperClassByRoles,
 } from '../role-helpers'
 import {render} from './helpers/test-utils'
 
@@ -181,6 +182,48 @@ test.each([
 ])('shouldExcludeFromA11yTree for %s returns %p', (html, expected) => {
   const {container} = render(html)
   container.firstChild.appendChild(document.createElement('button'))
-
   expect(isInaccessible(container.querySelector('button'))).toBe(expected)
 })
+
+test.each([
+  ['', []],
+  ['roletype', ['roletype']],
+  ['switch', ['roletype', 'widget', 'input', 'checkbox', 'switch']],
+  ['input', ['roletype', 'widget', 'input']],
+  [
+    'select',
+    [
+      'roletype',
+      'widget',
+      'composite',
+      'structure',
+      'section',
+      'group',
+      'select',
+    ],
+  ],
+  [
+    ['switch', 'input'],
+    ['roletype', 'widget', 'input', 'checkbox', 'switch'],
+  ],
+  [
+    ['switch', 'select'],
+    [
+      'roletype',
+      'widget',
+      'input',
+      'checkbox',
+      'composite',
+      'structure',
+      'section',
+      'group',
+      'switch',
+      'select',
+    ],
+  ],
+])(
+  'getSuperClassByRole returns expected roles as super class for given roles',
+  (roles, expected) => {
+    expect(getRolesAndSuperClassByRoles(roles).sort()).toEqual(expected.sort())
+  },
+)
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -346,7 +346,9 @@ Here are the accessible roles:
 test('has no useful error message in findBy', async () => {
   const {findByRole} = render(`<li />`)
 
-  await expect(findByRole('option', {timeout: 1})).rejects.toThrow('Unable to find role="option"')
+  await expect(findByRole('option', {timeout: 1})).rejects.toThrow(
+    'Unable to find role="option"',
+  )
 })
 
 test('explicit role is most specific', () => {
@@ -376,6 +378,46 @@ Here are the accessible roles:
   />
 </body>"
 `)
+})
+
+test('superclass roles are retrieved when are abstract roles', () => {
+  const {getAllByRole} = render(`
+   <>
+      <input type="text" />
+      <input type="checkbox" />
+      <textarea />
+    </>`)
+  expect(getAllByRole('textbox')).toHaveLength(2)
+  expect(getAllByRole('input')).toHaveLength(3)
+})
+
+test('superclass roles are retrieved for explicitly defined roles', () => {
+  const {getAllByRole} = render(`
+    <>
+      <input type="checkbox" />
+      <span role="switch" aria-checked="true" />
+    </>
+  `)
+  expect(getAllByRole('switch')).toHaveLength(1)
+  expect(getAllByRole('checkbox')).toHaveLength(2)
+})
+
+test('superclass list has found when the defined role is feed', () => {
+  const {getAllByRole} = render(`
+    <>
+      <ul>
+        <li>abc</li>
+      </ul>
+      <div role="feed">
+        <article role="listitem">xyz</article>
+      </div>
+      <ul role="feed">
+        <li role="article">def</li>
+      </ul>
+    </>
+  `)
+  expect(getAllByRole('list')).toHaveLength(3)
+  expect(getAllByRole('feed')).toHaveLength(2)
 })
 
 describe('configuration', () => {
--- a/src/queries/role.js
+++ b/src/queries/role.js
@@ -6,6 +6,7 @@ import {
   prettyRoles,
   isInaccessible,
   isSubtreeInaccessible,
+  getRolesAndSuperClassByRoles,
 } from '../role-helpers'
 import {wrapAllByQueryWithSuggestion} from '../query-helpers'
 import {checkContainerType} from '../helpers'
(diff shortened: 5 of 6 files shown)


src/__tests__/ariaAttributes.js
```diff
@@ -63,12 +63,17 @@ test('`selected: true` matches `aria-selected="true"` on supported roles', () =>
   ).toEqual(['selected-native-columnheader', 'selected-columnheader'])
 
   expect(getAllByRole('gridcell', {selected: true}).map(({id}) => id)).toEqual([
+    'selected-rowheader',
```
eps1lon, line 66: Appreciate the work but I'm not convinced this will actually be an improvement and not cause more confusion. I'd get behind including subclasses for abstract roles (not really though) but for concrete roles it's a bit confusing that this role would also be considered a `gridcell`. Because it actually isn't.

In it's current form this will also be a breaking change. This is evident from the changes in existing tests.

delca85: I understand your point.
The *byRole queries could get only the abstract super classes if it sounds more reasonable. Otherwise an additional parameter could be added to decide if the hierarchies should be navigated or not.

nickmccurdy: Wouldn't this change make testing library behave more like a real browser would? I don't consider bug fixes to be breaking changes. Either way I think it's worth it as long as this models browser behavior effectively.

eps1lon: > Wouldn't this change make testing library behave more like a real browser would?

I'm not aware of any behavior browsers are adding for aria roles. What do you think would `getByRole('gridcell')` accomplish if it would return `<th>` or `getByRole('input')` for `<input type="search" />`?

nickmccurdy: As far as I understand, this part of the WAI ARIA spec requires that subclasses implement properties of superclasses. https://www.w3.org/WAI/PF/aria/roles#inheritedattributes

In that case it seems to me like these changes are important for matching browser behavior. Additionally, not supporting this would violate the Liskov Substitution Principle (replacing a type with a subclass type needs to still retain the interfaces of the original superclass type).

kentcdodds: I don't know the solution to the problem Nick brings up 😬

nickmccurdy: I feel like the best approach is being very explicit about warning the user how the functionality is changing. For example we could have a bold warning in the changelog, and if a ByRole query has more than one result, it can warn that the role may be ambiguous because of subclasses.

Alternatively as I mentioned above, we can treat this as getting the explicit role attribute value rather than checking if a role matches. In that case I think we should rename the mat her to make it more clear what it does. We could even have a generic `ByAttribute` matcher that takes a property and value (similarly to Jest’s `toHaveProperty).

eps1lon: > In that case it seems to me like these changes are important for matching browser behavior. 

I want to ask again: What behaviors are browsers adding if they encounter aria roles? As far as I'm aware only assistive technology cares about this. I'd like to understand how they treat sub/super classes and how this is relevant to us. So far I've only seen theoretical arguments about how aria is specified. I don't want to burden users more with role taxonomy unless there is a concrete use case for it that justifies this tradeoff.

> For example we could have a bold warning in the changelog

That will be missed by the audience we target with this warning.

> This could be a confusing change in behavior of the tests, but as far as I understand is not actually a semver breaking change because it is a bug fix.

This is not a bug fix. We always talked about concrete roles. Class hierarchy was never part of the equation. `<th>` being returned from a query that previously returned `<td>` does not sound like something that should've been the case all along.

nickmccurdy: Browsers add tons of features based on ARIA roles, and one of our most important goals is to test user behavior in a way that encourages accessible.

For example, take this example of the `switch` from from MDN: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role

If I use this in VoiceOver, I hear "Speaker power, on, switch". This is very helpful information, and the `switch` role makes it more clear to me as a potentially blind user (who can't see that this is a switch and not a checkbox) that this is an on/off state and not a checked/unchecked state. If I just replace this role with `checkbox`, I hear "Interact with the title of checkbox", which is not helpful at all as I do not know what the checkbox is for and I don't know that it represents an on/off state.

While the UI distinction between `switch` and `checkbox` is important (which is why they are separate roles), the JavaScript logic here can work in a very similar way. I think it's very important that we model the correct browser behavior here to encourage realistic tests and accessible implementations of those tests. It's also possible that a user of testing library could refactor a `checkbox` to a `switch`, in which case we should still match it as a `checkbox` as it may have similar functionality as far as our tests are concerned, and we don't want to unnecessarily break tests (this is one of the main reasons why testing library was created as an alternative to Enzyme).

alexkrolick: My main idea when proposing this was pretty much the example Nick gave. There are some things, like checkboxes, that have subclasses that you could look for specifically, but that could also be an implementation detail at higher levels.

nickmccurdy: Technically relying on the value of an HTML attribute is also an implementation detail. Either way, supporting subclass hierarchy is more accurate to the spec and will make it behave more like browser implementations.

eps1lon: > Browsers add tons of features based on ARIA roles, and one of our most important goals is to test user behavior in a way that encourages accessibility.

Could you give me concrete examples? I'm asking you a third time now and I find it a bit rude that you handwave my questions again. I tried searching for these examples but can't find any. All I know of is that assistive technology e.g. screen readers read concrete aria roles. I could not find any examples for behavior of abstract roles. It's possible that I'm completely missing my point but if someone does not know something I'd rather share knowledge than insist that I'm right. I think this is a more constructive way of collaboration.

Even more concerning: I found statements that user agents should specifically not map abstract roles: https://www.w3.org/TR/wai-aria-implementation/#mapping_role

> It's also possible that a user of testing library could refactor a checkbox to a switch, in which case we should still match it as a checkbox as it may have similar functionality as far as our tests are concerned

I don't understand. A switch shares some functionality but has more constraints on the type of interactions. It would make sense to return the checkbox if the switch would simply add more functionality but here it removes some (here: the mixed state). So if you query a checkbox then you expect that the user is able to put the checkbox in a mixed state. A switch is not able to do that.

Another case I'd like some clarification on: If I have role="switch" and the screen reader does not support this role: Does is fall back to a checkbox or does it bail out completely? If it bails out then returning the element for `ByRole('checkbox')` is even wors.

In summary: You appeal to "how a browser behaves" but have no concrete examples for it. The only concrete argument you're making is a technical one and I don't find it convincing if the user is our priority.

kentcdodds: I think I'm with @eps1lon here.

I think what I'd like to see is an example of a test that someone can't easily right without this change and the workaround they have to perform because we don't support this.

nickmccurdy: > Could you give me concrete examples?

My apologies for any frustration, my intention was for `switch`/`checkbox` to be an example of this. I thought you were asking about how subclassed roles would be spoken differently (as it is a browser's responsibility to figure how to implement these roles with screenreader APIs). However this example doesn't address how not inferring a role could actually break behavior in a test logically (besides using `getByRole` with a superclass, which is just a problem with the test itself). I'll try to better examples of this (hopefully with failing tests), but I need to do more research first.

> All I know of is that assistive technology e.g. screen readers read concrete aria roles.

As far as I understand, screen readers do not directly read ARIA roles. Each browser is responsible for implementing the ARIA spec and mapping concrete ARIA roles to screenreader APIs, which can lead to different behavior depending on the browser, operating system, and screenreader.

> I could not find any examples for behavior of abstract roles.

Most [abstract roles](https://www.w3.org/TR/wai-aria/#abstract_roles) represent categories of related behaviors, rather than specific behaviors, so they're mostly not relevant to us. However, I did notice that the behavior of `composite` roles may require changes to `tab` in `user-event`, so I opened testing-library/user-event#394.

> Even more concerning: I found statements that user agents should specifically not map abstract roles: https://www.w3.org/TR/wai-aria-implementation/#mapping_role

I believe this is referring to mapping from abstract ARIA roles in the DOM to accessibility APIs (not from superclass ARIA roles to subclass ARIA roles). In other words, an abstract role does not represent a relationship to existing accessibility APIs, it's just a way to group ARIA roles that have related behaviors. You're still right that we don't need to handle mapping abstract classes though, as they don't affect behavior of `dom-testing-library` functions in tests.

> So if you query a checkbox then you expect that the user is able to put the checkbox in a mixed state.

Good point, I suppose if we do need to implement subclass hierarchies, it would be less confusing (and more type safe in TypeScript) to have a separate function or mode with different semantics. By default, we could just warn when additional elements are found with subclass roles. This wouldn't break the tests on its own, but if a test failed because you were assuming that changing a role to a subclass role in the DOM would still cause it to be matched by testing library, you'd know that you'd have to change the `getByRole` call to use the concrete role instead.

> Another case I'd like some clarification on: If I have role="switch" and the screen reader does not support this role: Does is fall back to a checkbox or does it bail out completely?

The different switch behavior is optional, so if a browser decides to not implement it, it can just fall back tot he checkbox role implementation.

nickmccurdy: # Alternatives
These address disadvantages of not handling subclass roles semantically without breaking changes:
1. Separate `ByInheritedRole` functions
1. Separate `queryInheritance` flag (similar to `queryFallbacks` but for subclasses instead of fallbacks)
1. Don't add APIs, but have `ByRole` warn when additional elements are found with subclass roles
1. Avoid code changes, but explicitly mention in the documentation that inheritance isn't used for `ByRole`

kentcdodds: I'd still like to get a practical example of what users have to do because we don't have support for subclasses 🤔

nickmccurdy: As far as I can tell, there isn't one beyond modelling the user's expectations for the semantics of whether subclasses are matched when using `ByRole`

kentcdodds: In that case, and in light of the arguments against this I see this as a situation where the benefits (technically correct, but that's all we get) are outweighed by the costs (as iterated by @eps1lon above).

nickmccurdy: What do you think of my ideas for [alternatives](https://github.com/testing-library/dom-testing-library/pull/675#discussion_r449523726)? None of them require breaking changes, and as far as I'm aware they should be low cost as most of the implementation work for looking up superclasses is done here.

kentcdodds: The only option that makes sense to me is 3.

nickmccurdy: I'd be okay with it, though I don't think it's as user friendly as an automated warning.

nickmccurdy:
I've updated the docs in testing-library/testing-library-docs#520 as an alternative to this, and it's been merged. Do we want to close this?

kentcdodds:
Yeah, as much as it pains me to close @delca85's PR, I think it's better this way (so sorry Bianca!! we'll get a PR from you merged in here eventually I swear).

src/__tests__/ariaAttributes.js
Dreamsorcerer: Well, I'm rather late to this discussion, but would like to add some remarks as I think abstract roles should still be supported.

> Even more concerning: I found statements that user agents should specifically not map abstract roles: https://www.w3.org/TR/wai-aria-implementation/#mapping_role

I think you may have misinterpreted that. I read that as an element should be mapped to a concrete role (e.g. textbox), but a textbox is a subclass of input, so if you query for input you should find that element (otherwise, what do you think the abstract role _is_ for?).

But, you shouldn't map the element directly to an input instead of a textbox.

> Could you give me concrete examples?

I don't know the implementation details, but how would a user navigate a form without the input role? I assume that screen readers and probably the default tab behaviour in browsers work by iterating over the input elements.

Without input roles, do you expect them to get a list of checkboxes, a list of textboxes etc. and then present them to the user in a seemingly random order?

> I think what I'd like to see is an example of a test that someone can't easily right without this change and the workaround they have to perform because we don't support this.

Here's a simple test that can't be done easily today:
```
expect(within(form).getAllByRole("input").map((e) => e.value)).toEqual(["Foo", "Bar", 5]);
```

I would also look at the number of closed issues and stackoverflow questions from users asking how to select inputs with ByRole(), it appears to be a very frequent request.

Dreamsorcerer: Another test would be to test the labels, although I don't think there's a convenience attribute to get it (something else that would nice to have):
```
expect(within(form).getAllByRole("input").map((e) => e.labelText)).toEqual(]"ID", "Name", "Number"]);
```

The important thing that both of these tests do is validate the order of elements presented to the user. I believe it is actually _impossible_ to write these tests with ByRole() today.

kentcdodds: I'm going to need to see a real use case. And please address the other concerns brought up in the comments of this PR

Dreamsorcerer: I think the other concerns are mostly related to backwards compatibility. Abstract roles have no impact on that. If only abstract roles are included, then it seems like it should be easy to include while providing 90+% of what users actually want out of it.

> I'm going to need to see a real use case.

A real use case for testing? The above examples are actual tests I tried to write for my application, so that is a real use case.

The application is dynamically generated from a schema, so validating that the inputs appear in the expected order is an important detail.

Dreamsorcerer: > The application is dynamically generated from a schema, so validating that the inputs appear in the expected order is an important detail (while also saving me writing several different expect()s, one for each input).

e.g. Here are exactly the same kind of tests on a table when viewing the data (rather than editing in a form):
https://github.com/aio-libs/aiohttp-admin/blob/ee9498171d0f608574c5c80fb7dc80c0dac51af9/admin-js/tests/simple.test.js#L19

kentcdodds: I'm open to looking at a new PR which handles abstract roles.

Can you show me how your test would change with this new feature?

Dreamsorcerer: > Can you show me how your test would change with this new feature?

The first test examples above, which operate on inputs of a form, would work. Today they don't (the last linked test is for the view mode, where data is in a table). My in-progress tests locally currently look like:
```
expect(within(edit).getByLabelText("Id *")).toHaveValue(1);
expect(within(edit).getByLabelText("Num *")).toHaveValue(5);
expect(within(edit).getByLabelText("Optional Num")).toHaveValue(null);
expect(within(edit).getByLabelText("Value *")).toHaveValue("first");
```
Which fails to test ordering while being very verbose. This would become:
```
expect(within(edit).getAllByRole("input").map((e) => e.value)).toEqual([1, 5, null, "first"]);
```
Probably with a second line to assert labels as mentioned above.

kentcdodds: That makes sense.

I don't actively maintain this project, but I think a PR would be the next step.

PR #677, opened 2020-06-28
Adds matcher name to multiple elements error

JacobParis:
**What**:
The getByRole queries are designed to return a single element whose role matches the provided role argument and whose accessible name matches the provided name argument.

When multiple elements are matched, the language in the error message only mentions the role and not the accessible name.

**Sample query**
```js
screen.getByRole('definition', { name: /value/i })
```

**Current error message**
```
TestingLibraryElementError: Found multiple elements with the role "definition"
```
**New error messages**
```
TestingLibraryElementError: Found multiple elements with the role "definition" and name "value"
```
```
TestingLibraryElementError: Found multiple elements with the role "definition" and name `/value/i`
```

**Why**:

When you have several queries against the same role but for different accessible names, the error messages produced by them are identical. The only way to figure out which name is finding duplicates is to look at the code trace.

**How**:

The information including the accessible name parameter was already provided to the error message handler. The logic for displaying the name with or without quotes depending on its type was already present in the `getMissingError` handler, so I copied it to the `getMultipleError` handler

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -346,7 +346,9 @@ Here are the accessible roles:
 test('has no useful error message in findBy', async () => {
   const {findByRole} = render(`<li />`)
 
-  await expect(findByRole('option', {timeout: 1})).rejects.toThrow('Unable to find role="option"')
+  await expect(findByRole('option', {timeout: 1})).rejects.toThrow(
+    'Unable to find role="option"',
+  )
 })
 
 test('explicit role is most specific', () => {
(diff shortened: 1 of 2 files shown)


kentcdodds:
Love it. Thanks! Could you add a test for this as well?

JacobParis:
There did not appear to be a test explicitly covering the multiple element error in the first place so I've added a new one

kentcdodds:
Looks like we're not quite there on coverage. Let me know if you need help identifying what's left.

JacobParis:
Just pushed up the missing test

kentcdodds:
@all-contributors please add @JacobParis for code and tests

PR #681, opened 2020-07-03
get by label concat values

delca85:
**What**:

- `input` elements with not empty value are retrieved by `getAllLabelsByText`
- `getByLabelText` gets labels even when it receives a label composed of a concatenation

**Why**:

Issue #545 finds out that `input` elements are not recognized as labels and  labelling by concatenation is not got by `getByLabelText`.

**How**:
1. `queryAllLabelsByText` now makes a query not only on elements of type 
 label but on input elements too. The empty input are excluded.
2. after evaluating labels node through `matcher`, the labels that doesn't match are combined through an helper function. If a combination matches with the received text than the labels is returned by `queryAllLabelsByText`. 

**Checklist**:

- [x] make `queryAllLabelsByText` look for `input` elements too
- [x] create a `queryAllLabels` function that retrieves all labels without any regards about text to be matched
- [x] find labels matching 
- [x] create `combination` function that concatenate all the labels and returns the node that through some combination matches with the received text
- [x] `queryAllLabelsByText` returns now the labels matching by itself and the ones matching through concantenation
 

- [x] Tests
- [x] Ready to be merged
      

I have already opened #607 to fix this behavior but maybe not in the right way. I hope this idea is better than the one I have submitted before.

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -188,6 +188,7 @@ test('can get form controls by label text', () => {
   expect(getByLabelText('5th two').id).toBe('fifth-id')
   expect(getByLabelText('6th one').id).toBe('sixth-id')
   expect(getByLabelText('6th two').id).toBe('sixth-id')
+  expect(getByLabelText('6th one 6th two').id).toBe('sixth-id')
   expect(getByLabelText('6th one 6th two 6th three').id).toBe('sixth-id')
   expect(getByLabelText('7th one').id).toBe('seventh-id')
 })
@@ -341,6 +342,61 @@ test('label with no form control', () => {
 <div>
   <label>
     All alone
+  </label>
+</div>"
+`)
+})
+
+test('label with no form control and fuzzy matcher', () => {
+  const {getByLabelText, queryByLabelText} = render(
+    `<label>All alone label</label>`,
+  )
+  expect(queryByLabelText('alone', {exact: false})).toBeNull()
+  expect(() => getByLabelText('alone', {exact: false}))
+    .toThrowErrorMatchingInlineSnapshot(`
+"Found a label with the text of: alone, however no form control was found associated to that label. Make sure you're using the "for" attribute or "aria-labelledby" attribute correctly.
+
+<div>
+  <label>
+    All alone label
+  </label>
+</div>"
+`)
+})
+
+test('label with children with no form control', () => {
+  const {getByLabelText, queryByLabelText} = render(`
+  <label>
+    All alone but with children
+    <textarea>Hello</textarea>
+    <select><option value="0">zero</option></select>
+  </label>`)
+  expect(queryByLabelText(/alone/, {selector: 'input'})).toBeNull()
+  expect(() => getByLabelText(/alone/, {selector: 'input'}))
+    .toThrowErrorMatchingInlineSnapshot(`
+"Found a label with the text of: /alone/, however no form control was found associated to that label. Make sure you're using the "for" attribute or "aria-labelledby" attribute correctly.
+
+<div>
+  
+  
+  <label>
+    
+    All alone but with children
+    
+    <textarea>
+      Hello
+    </textarea>
+    
+    
+    <select>
+      <option
+        value="0"
+      >
+        zero
+      </option>
+    </select>
+    
+  
   </label>
 </div>"
 `)
@@ -960,4 +1016,27 @@ test('can get a select with options', () => {
     </label>
   `)
   getByLabelText('Label')
+})
+
+test('can get an element with aria-labelledby when label has a child', () => {
+  const {getByLabelText} = render(`
+    <div>
+      <label id='label-with-textarea'>
+        First Label
+        <textarea>Value</textarea>
+      </label>
+      <input aria-labelledby='label-with-textarea' id='1st-input'/>
+      <label id='label-with-select'>
+        Second Label
+        <select><option value="1">one</option></select>
+      </label>
+      <input aria-labelledby='label-with-select' id='2nd-input'/>
+    </div>
+  `)
+  expect(getByLabelText('First Label', {selector: 'input'}).id).toBe(
+    '1st-input',
+  )
+  expect(getByLabelText('Second Label', {selector: 'input'}).id).toBe(
+    '2nd-input',
+  )
 })
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -171,7 +171,8 @@ test('can get form controls by label text', () => {
       <div>
         <input id="sixth-label-one" value="6th one"/>
         <input id="sixth-label-two" value="6th two"/>
-        <input aria-labelledby="sixth-label-one sixth-label-two" id="sixth-id" />
+        <label id="sixth-label-three">6th three</label>
+        <input aria-labelledby="sixth-label-one sixth-label-two sixth-label-three" id="sixth-id" />
       </div>
     </div>
   `)
@@ -183,6 +184,7 @@ test('can get form controls by label text', () => {
   expect(getByLabelText('5th two').id).toBe('fifth-id')
   expect(getByLabelText('6th one').id).toBe('sixth-id')
   expect(getByLabelText('6th two').id).toBe('sixth-id')
+  expect(getByLabelText('6th one 6th two 6th three').id).toBe('sixth-id')
 })
 
 test('can get elements labelled with aria-labelledby attribute', () => {
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -174,6 +174,10 @@ test('can get form controls by label text', () => {
         <label id="sixth-label-three">6th three</label>
         <input aria-labelledby="sixth-label-one sixth-label-two sixth-label-three" id="sixth-id" />
       </div>
+      <div>
+        <span id="seventh-label-one">7th one</span>
+        <input aria-labelledby="seventh-label-one" id="seventh-id" />
+      </div>
     </div>
   `)
   expect(getByLabelText('1st').id).toBe('first-id')
@@ -185,6 +189,7 @@ test('can get form controls by label text', () => {
   expect(getByLabelText('6th one').id).toBe('sixth-id')
   expect(getByLabelText('6th two').id).toBe('sixth-id')
   expect(getByLabelText('6th one 6th two 6th three').id).toBe('sixth-id')
+  expect(getByLabelText('7th one').id).toBe('seventh-id')
 })
 
 test('can get elements labelled with aria-labelledby attribute', () => {
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -168,6 +168,11 @@ test('can get form controls by label text', () => {
         <label id="fifth-label-two">5th two</label>
         <input aria-labelledby="fifth-label-one fifth-label-two" id="fifth-id" />
       </div>
+      <div>
+        <input id="sixth-label-one" value="6th one"/>
+        <input id="sixth-label-two" value="6th two"/>
+        <input aria-labelledby="sixth-label-one sixth-label-two" id="sixth-id" />
+      </div>
     </div>
   `)
   expect(getByLabelText('1st').id).toBe('first-id')
@@ -176,6 +181,8 @@ test('can get form controls by label text', () => {
   expect(getByLabelText('4th').id).toBe('fourth.id')
   expect(getByLabelText('5th one').id).toBe('fifth-id')
   expect(getByLabelText('5th two').id).toBe('fifth-id')
+  expect(getByLabelText('6th one').id).toBe('sixth-id')
+  expect(getByLabelText('6th two').id).toBe('sixth-id')
 })
 
 test('can get elements labelled with aria-labelledby attribute', () => {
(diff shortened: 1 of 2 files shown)


src/queries/label-text.js
```diff
@@ -12,30 +12,73 @@ import {
 } from './all-utils'
 import {queryAllByText} from './text'
 
+function getCombinations(labels, matcher) {
+  const combs = [[]]
+  const matching = []
+  for (const label of labels) {
+    const copy = [...combs] // See note below.
```
kentcdodds, line 19: I'm not sure what note this is referring to. Also, if I'm following this logic correctly, the for loop on the next line will only be run only once (for this iteration of the nested loop) and the value of the `prefix` will be `[]` every time. Any reason we need a loop here?

src/queries/label-text.js
```diff
@@ -12,30 +12,73 @@ import {
...
+      combs.push(combination)
+      if (matcher(combination.join(' '), label.node)) {
+        matching.push(label.node)
+      }
+    }
+  }
+  return matching
+}
+function queryAllLabels(container) {
+  return Array.from(container.querySelectorAll('label,input'))
```
kentcdodds, line 31: It occurs to me that an element can be labeled by *anything*, not just `label` and `input`. So while this may be solving for the input case, it doesn't solve for everything else 🤔

That said, I'm not sure how to handle "everything else." It seems to me that rather than searching for all possible elements that can label anything, we should be searching for everything that's labeled first, determine the contents of their label, and then return the elements whose label matches the query.

src/queries/label-text.js
```diff
@@ -12,30 +12,73 @@ import {
...
+      matcher(textToMatch, node, text, matchNormalizer),
+    )
+    .map(({node}) => node)
+  const labelsNotMatchingTextAndNotEmpty = textToMatchByLabels.filter(
+    ({node, textToMatch}) =>
+      Boolean(textToMatch) &&
+      nodesByLabelMatchingText.findIndex(nodeByLabelByText =>
+        nodeByLabelByText.isEqualNode(node),
+      ) === -1,
+  )
```
kentcdodds, line 74:
```suggestion
  const labelsNotMatchingTextAndNotEmpty = textToMatchByLabels.filter(
    ({node, textToMatch}) =>
      textToMatch && nodesByLabelMatchingText.some(n => n.isEqualNode(node)),
  )
```

src/queries/label-text.js
delca85: I am not sure I am totally getting your point, are you telling that [this](https://github.com/testing-library/dom-testing-library/blob/732a8abf5560b13cfba976f5f801048e7018e2d4/src/queries/label-text.js#L20) loop runs only once inside the [outer](https://github.com/testing-library/dom-testing-library/blob/732a8abf5560b13cfba976f5f801048e7018e2d4/src/queries/label-text.js#L18) loop?

src/queries/label-text.js
delca85: I am not sure I am totally getting your point, are you telling that [this](https://github.com/testing-library/dom-testing-library/blob/732a8abf5560b13cfba976f5f801048e7018e2d4/src/queries/label-text.js#L20) loop runs only once inside the [outer](https://github.com/testing-library/dom-testing-library/blob/732a8abf5560b13cfba976f5f801048e7018e2d4/src/queries/label-text.js#L18) loop?

kentcdodds: Correct. Because `copy` is `[[]]` and so it's an array with only one element in it. It'll never be....

Wait. I guess that's only the case on the first iteration of the outer loop. Future iterations could have more... Hmmm... I need to take more time to understand what's going on here...

src/queries/label-text.js
delca85: For what it's worth, I think that searching for labelled items and then for their labels is more feasible than querying for all elements that could be labels. Like you said above!

kentcdodds: If you'd like to tack a whack at that, I think it would solve a lot of problems.

src/queries/label-text.js
delca85: Yes, I think it's only the first iteration of the outer loop that makes the inner one run only once.

src/queries/label-text.js
delca85: I think the right change should be:
```javascript
const labelsNotMatchingTextAndNotEmpty = textToMatchByLabels.filter(
    ({node, textToMatch}) =>
      textToMatch && !nodesByLabelMatchingText.some(n => n.isEqualNode(node)),
  )
```
because I would like to retrieve all the labels in `textToMatchByLabels ` but not empty and not included in `nodesByLabelMatchingText`

src/queries/label-text.js
delca85: For sure I would like to do it!

delca85:
Hey @kentcdodds !
I have pushed some changes that follow the suggestion you gave me yesterday.

Now [`queryAllByLabelText`](https://github.com/testing-library/dom-testing-library/blob/f3a3c65d0f8551ce2b85892a585110a30a330d19/src/queries/label-text.js#L80) at the beginning looks for labelled element and then checks their labels value. 
I still makes a [research](https://github.com/testing-library/dom-testing-library/blob/f3a3c65d0f8551ce2b85892a585110a30a330d19/src/queries/label-text.js#L141) of all the label elements to get all the elements labelled but without `aria-labelledby` and `aria-label` attribute.

Right now I am still using the "old" `queryAllLabelsByText` in order to be able to send a specific error message in case of the searched label is found but none form control is associated to it. I think this is not totally right because that `queryAllLabelsByText` is looking just for `label` and `input` element not everything while everything could be a label (as you told).

I suppose my job is not finished yet but this is what I achieved.

src/__tests__/element-queries.js
```diff
@@ -332,6 +347,61 @@ test('label with no form control', () => {
...
+`)
+})
+
+test('label with children with no form control', () => {
+  const {getByLabelText, queryByLabelText} = render(`
+  <label>
+    All alone but with children
+    <textarea>Hello</textarea>
+    <select><option value="0">zero</option></select>
+  </label>`)
```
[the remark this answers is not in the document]
delca85: Maybe I am not understanding, you are saying that the right test to be done here (at line 374) was that the labeled element is `textarea` instead of that no `input` element labeled was found?

`getByLabelText(/alone/)` would have found the `textarea` element, I added the selector to make it return null.

Please tell me if I am missing something.

kentcdodds:
I apologize it's taken me so long to give feedback here. I've been doing more family stuff than open source stuff recently.

I was thinking about a much bigger reworking of the logic for this function.

Right now, here's what we try to do:

1. Find all matching labels and then find all the elements those labels are used to label
2. Find all other matching elements and see if they're also used to label anything

The #545 should be found via the `#2` option there, but it's difficult because the search query matches across multiple labels, so we won't find it without some additional complex logic.

What I was thinking is we could do this instead:

1. Find all labeled elements (everything with a `.label` property or an `aria-labeledby` attribute)
2. Filter those elements to the ones whose labels match the search text

I think this would allow us to more easily solve #545 and would result in fewer bugs. I expect that this would be a non-breaking change.

Does that make sense?

delca85:
> I apologize it's taken me so long to give feedback here. I've been doing more family stuff than open source stuff recently.

Don't be sorry at all, family stuff is always first class stuff!

> Right now, here's what we try to do:
> 
> 1. Find all matching labels and then find all the elements those labels are used to label
> 2. Find all other matching elements and see if they're also used to label anything
> 
> The #545 should be found via the `#2` option there, but it's difficult because the search query matches across multiple labels, so we won't find it without some additional complex logic.

Totally agree with you, specially regarding the additional complex logic required in order to solve #545 .

> What I was thinking is we could do this instead:
> 
> 1. Find all labeled elements (everything with a `.label` property or an `aria-labeledby` attribute)
> 2. Filter those elements to the ones whose labels match the search text
> 
> I think this would allow us to more easily solve #545 and would result in fewer bugs. I expect that this would be a non-breaking change.
> 
> Does that make sense?

It makes a lot of sense and it is what I have tried to do in the code I pushed [here](https://github.com/testing-library/dom-testing-library/pull/681/commits/f3a3c65d0f8551ce2b85892a585110a30a330d19) :
1. at [this](https://github.com/testing-library/dom-testing-library/blob/f3a3c65d0f8551ce2b85892a585110a30a330d19/src/queries/label-text.js#L90) line I make a query to get all the elements with `aria-labelledby` and then I look for their value to check if it matches with the desired one;
2. at [this](https://github.com/testing-library/dom-testing-library/blob/f3a3c65d0f8551ce2b85892a585110a30a330d19/src/queries/label-text.js#L139) line I retrieve the element with `aria-label` attribute and check if their value match with the desired one
3. finally I get all the `label` element because I thought that some labels still be missed by the first two sets retrieved by the steps 1 and 2.

In comparison to what you wrote, I check the value of the elements I found with a label immediately after I found them, because I thought it is more performant (I save a loop doing this instead of checking the value of all the labeled elements after I found all of them).

Like I told you [here](https://github.com/testing-library/dom-testing-library/pull/681#issuecomment-653814805) I am still using `queryAllLabelsByText` and this is the reason (and an issue related):

> Right now I am still using the "old" `queryAllLabelsByText` in order to be able to send a specific error message in case of the searched label is found but none form control is associated to it. I think this is not totally right because that `queryAllLabelsByText` is looking just for `label` and `input` element not everything while everything could be a label (as you told).

I hope I explained what I did in a clear way, if I am still miss something please tell me.
Anyway, thanks for your time.

src/queries/label-text.js
```diff
@@ -45,73 +58,94 @@ function queryAllByLabelText(
...
-        // we're using this notation because with the # selector we would have to escape special characters e.g. user.name
-        // see https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector#Escaping_special_characters
-        // <label for="someId">text</label><input id="someId" />
-
-        // .control support has landed in jsdom (https://github.com/jsdom/jsdom/issues/2175)
-        elementsForLabel.push(
-          container.querySelector(`[id="${label.getAttribute('for')}"]`),
+
+  const matchingLabelledElements = Array.from(container.querySelectorAll('*'))
+    .filter(element => element.hasAttribute('aria-labelledby'))
```
kentcdodds, line 65: I think we might be able to clean up some stuff (and reduce the number of loops) by changing this filter to include elements with the `.label` property:

```suggestion
    .filter(element => element.label || element.hasAttribute('aria-labelledby'))
```

This would require the `reducer` function handle the `.label` elements a little differently, but I think it would allow us to remove some other code as well (specifically all the code for `matchingElementsByLabels`).

delca85: I am going to try what you suggested as soon as possible.

delca85: Hey @kentcdodds !
For sure I am missing something but so far I have not been able to make the change you suggested me. 
Trying to include the elements with the `.label` property in the filter you highlighted it is not working because the element labeled by a `label` element put around them have the `.label` property undefined.
I have tried to show it in [this](https://codesandbox.io/s/nostalgic-benz-fnxsm?file=/src/__tests__/index.js) sandbox.

Again, I am aware there is something I am not doing in the right way but I am not able to get it.

kentcdodds: Oh, sorry, that's a typo. It should be `.labels`. This is a `NodeList` of all the labels. If there are multiple, then it looks like the Chrome Accessibility DevTools combines them:

image

https://jsbin.com/pudalunaxu/1/edit?html,output

```html
<div>
  <label for="my-input">One</label>
  <label for="my-input">Two</label>
  <input id="my-input" />
</div>
```

So I think that's probably what we should do as well. Combine the text for all labels in the NodeList, then compare that against the given TextMatch.

Does that make sense?

delca85: Yes, it makes a lot of sense.
I have just pushed some code using `.labels`.

I am not so sure I have done all the work because I am not obtaining all the labels in the NodeList through a permutation but just join their values (and join values in the NodeList removing an item a time). I don't know if you meant this with `combine`.

src/queries/label-text.js
```diff
@@ -45,74 +69,69 @@ function queryAllByLabelText(
...
+  const matchingLabelledElements = Array.from(container.querySelectorAll('*'))
+    .filter(
+      element => element.labels || element.hasAttribute('aria-labelledby'),
+    )
+    .reduce((labelledElements, labelledElement) => {
+      const labelsId = labelledElement.getAttribute('aria-labelledby')
+        ? labelledElement.getAttribute('aria-labelledby').split(' ')
+        : []
+      const labelsValue = labelsId.length
+        ? labelsId.map(labelId => {
```
kentcdodds, line 83: What happens if an element had both a .labels property and the aria-labelledby? I don't know if that's valid or not.

eps1lon: aria-labelledby has precedence: https://www.w3.org/TR/accname-1.1/#step2B

2B is for aria-labelledby, 2D for <label>

kentcdodds: In that case, I think this PR is good to go! Anyone have other thoughts?

kentcdodds:
@all-contributors please add @delca85 for tests and code

kentcdodds:

delca85:
Thank you so much @kentcdodds !
I have learnt a lot working on this PR and it's totally a pleasure being part of this team.
Thanks!

smeijer:
Thank you @delca85. This is PR is really awesome!

delca85:
> Thank you @delca85. This is PR is really awesome!

Thanks @smeijer , I really hope you are right!

PR #683, opened 2020-07-07
fix(types): allow bound function to accept string and regex matcher

### LINKED ISSUE #680: Signature for BoundFunction<getByRole> does not accept RegExp in 7.20.0
- `@testing-library/dom` version: 7.20.0, broken by https://github.com/testing-library/dom-testing-library/pull/670

### Relevant code or config:

Based on [the tests in `@testing-library/vue` on Definitely Typed](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/testing-library__vue/testing-library__vue-tests.ts):

```js
import { queries, BoundFunction } from '@testing-library/dom';
declare var y: BoundFunction<typeof queries.getByRole>
y(/some text/, {exact: true, trim: false, collapseWhitespace: true, normalizer: x => x}); // $ExpectType HTMLElement
```

As of 7.20.0, this gives an error on the regex `/some text/`. It also gives an error on the string "some text".

This is because `BoundFunction` is a conditional type that does not understand multiple overloads, but #670 introduced multiple overloads for `getByRole` in order to improve completions for ARIARole. Getting rid of the overloads fixes the problem.
has a bad interaction between the overloads of `getByRole` in `queries.d.ts` and the conditional type `BoundFunction` in get-queries-for-element. But the types were so complex that I didn't have time to finish debugging it.

### What you did:

Running tests for Definitely Typed, including `@testing-library/vue`: https://dev.azure.com/definitelytyped/DefinitelyTyped/_build/results?buildId=53614&view=logs&j=b2dd9be2-fb32-52f8-c4c9-7dc0bc384587&t=f2dea62c-b696-5a51-78ad-206032e0a320&l=22288

### What happened:

Error in testing-library__vue
Error: /home/vsts/work/1/s/DefinitelyTyped/types/testing-library__vue/testing-library__vue-tests.ts:69:21
ERROR: 69:21  expect  TypeScript@4.0 compile error: 
Argument of type 'RegExp' is not assignable to parameter of type 'ARIARole'.
  Type 'RegExp' is not assignable to type '"alertdialog"'.

### Problem description:

Previously, arbitrary strings, regexes and matcher functions were allowed as the first argument to `BoundFunction<getByRole>`. Now only literal string types that are part of ARIARole are allowed.

### Suggested solution:

Just getting rid of the second overload didn't fix the problem for me; I had to create a new type alias like queries.d.ts used to have.

On the other hand, I don't know why the ComponentHarness interface in @testing-library/vue inherits from BoundFunctions in:

```ts
export interface ComponentHarness extends BoundFunctions<typeof queries> {
  container: HTMLElement;
  baseElement: HTMLBodyElement;
  debug(el?: HTMLElement): void;
  unmount(): void;
  isUnmounted(): boolean;
  html(): string;
  emitted(): { [name: string]: any[][] };
  updateProps(props: object): Promise<void>;
}
```

So it's possible that's not a reasonable thing to do. Maybe @cimbul can explain that &mdash; I couldn't find any documentation on BoundFunctions anywhere.

timdeschryver:
**What**:

- Currently it isn't possible to use bound `byRole` queries (for example with `screen`) in combination with a string or regexp matcher, `screen.getByRole('foo')` or `screen.getByRole(/head/)`. The types only allow an ARIARole, e.g. `screen.getByRole('button')`
- Closes #680 

**Why**:

- #670 introduced an unexpected breaking change
- We want our users to use strings and regexp matchers

**How**:

- Removed the method overload on the `getByRole` queries 
- Creating a `ByRoleMatcher` that uses the `String` type instead of the `string`  type we can have intellisense while not introducing a breaking change (we are still able to use a string that isn't included in the `AriaRole` union). 
- In https://github.com/testing-library/dom-testing-library/commit/3938e9224eb7f3504d2cf9488712da49e9a558d5 I tried to explicit type the bound functions, but this also introduced a breaking change to several parts of the API. 

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

I think for now, using `String` is the best solution. We can provide intellisense while not introducing a breaking change. We could revisit the types when we release a new major version.

### DIFF
--- a/types/matches.d.ts
+++ b/types/matches.d.ts
@@ -3,12 +3,9 @@ import {ARIARole} from 'aria-query'
 export type MatcherFunction = (content: string, element: HTMLElement) => boolean
 export type Matcher = string | RegExp | MatcherFunction
 
-// Allow `String` to have intellisense for roles while writing ByRole queries
-// If `string` is used it overrules the ARIARole string union types
-// resulting in no intellisense
-// We still want to support roles not included in the union type
-/* tslint:disable:ban-types */
-export type ByRoleMatcher = ARIARole | String | RegExp | MatcherFunction
+// Get autocomplete for ARIARole union types, while still supporting another string
+// Ref: https://github.com/microsoft/TypeScript/issues/29729#issuecomment-505826972
+export type ByRoleMatcher = ARIARole | (string & {}) | RegExp | MatcherFunction
 
 export type NormalizerFn = (text: string) => string
 
--- a/types/matches.d.ts
+++ b/types/matches.d.ts
@@ -14,6 +14,32 @@ export interface MatcherOptions {
   suggest?: boolean
 }
 
+export interface ByRoleOptions extends MatcherOptions {
+  /**
+   * If true includes elements in the query set that are usually excluded from
+   * the accessibility tree. `role="none"` or `role="presentation"` are included
+   * in either case.
+   */
+  hidden?: boolean
+  /**
+   * If true only includes elements in the query set that are marked as
+   * selected in the accessibility tree, i.e., `aria-selected="true"`
+   */
+  selected?: boolean
+  /**
+   * Includes every role used in the `role` attribute
+   * For example *ByRole('progressbar', {queryFallbacks: true})` will find <div role="meter progressbar">`.
+   */
+  queryFallbacks?: boolean
+  /**
+   * Only considers  elements with the specified accessible name.
+   */
+  name?:
+    | string
+    | RegExp
+    | ((accessibleName: string, element: Element) => boolean)
+}
+
 export type Match = (
   textToMatch: string,
   node: HTMLElement | null,
--- a/types/matches.d.ts
+++ b/types/matches.d.ts
@@ -14,32 +14,6 @@ export interface MatcherOptions {
   suggest?: boolean
 }
 
-export interface ByRoleOptions extends MatcherOptions {
-  /**
-   * If true includes elements in the query set that are usually excluded from
-   * the accessibility tree. `role="none"` or `role="presentation"` are included
-   * in either case.
-   */
-  hidden?: boolean
-  /**
-   * If true only includes elements in the query set that are marked as
-   * selected in the accessibility tree, i.e., `aria-selected="true"`
-   */
-  selected?: boolean
-  /**
-   * Includes every role used in the `role` attribute
-   * For example *ByRole('progressbar', {queryFallbacks: true})` will find <div role="meter progressbar">`.
-   */
-  queryFallbacks?: boolean
-  /**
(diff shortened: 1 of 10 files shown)


types/matches.d.ts
```diff
@@ -1,6 +1,15 @@
...
+
 export type MatcherFunction = (content: string, element: HTMLElement) => boolean
 export type Matcher = string | RegExp | MatcherFunction
 
+// Allow `String` to have intellisense for roles while writing ByRole queries
+// If `string` is used it overrules the ARIARole string union types
+// resulting in no intellisense
+// We still want to support roles not included in the union type
+/* tslint:disable:ban-types */
+export type ByRoleMatcher = ARIARole | String | RegExp | MatcherFunction
```
eps1lon, line 11: This use case is tracked in https://github.com/microsoft/TypeScript/issues/29729.

I'd prefer we go with the approach that is documented in that issue: `'literal' | (string & {})` (https://github.com/microsoft/TypeScript/issues/29729#issuecomment-505826972). Then we at least follow community best practices. And we should probably link the issue in the comment in case it gets resolved.

timdeschryver: 🤯 I was looking for a workaround but I couldn't find any.
Thanks for also giving this a look @eps1lon , this is simply great!

benmonro:
@timdeschryver looks 7.20.0 also broke @testing-library/testcafe as well with the following failure:

```sh
FAIL ./selectors.test.ts
[test:unit:monorepo]   ● Test suite failed to run
[test:unit:monorepo] 
[test:unit:monorepo]     ../../src/index.ts:71:14 - error TS2349: This expression is not callable.
[test:unit:monorepo]       Each member of the union type '{ (container: HTMLElement, role: Matcher, options?: ByRoleOptions | undefined): HTMLElement[]; (container: HTMLElement, role: ARIARole, options?: ByRoleOptions | undefined): HTMLElement[]; } | ... 13 more ... | FindAllByBoundAttribute' has signatures, but none of those signatures are compatible with each other.
[test:unit:monorepo] 
[test:unit:monorepo]     71       return window.TestingLibraryDom[query](
[test:unit:monorepo]                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[test:unit:monorepo] 
[test:unit:monorepo] Test Suites: 1 failed, 1 total
```

Here's the link w/ all the details: https://github.com/testing-library/testcafe-testing-library/pull/204/checks?check_run_id=830176994

seems related, but maybe i'm wrong...

timdeschryver:
@benmonro this fix should also resolve that issue.

timdeschryver:
I have rebased this branch, should be good to go for another review.

kentcdodds:
Anyone with rights can feel free to merge.

PR #690, opened 2020-07-11
Removed duplicated test

keiya01:
**What**:

I removed duplicated test in wait-for test.

**Why**:

I think duplicated test makes us confuse.
However, if you have intended this test, could you teach me the reason ?

**How**:

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged


eps1lon:
Thanks but these tests test different things (see description). Just because we can move everything into a single test doesn't mean we have to.

PR #692, opened 2020-07-12
support `{checked: true}` for checkbox / radio

### LINKED ISSUE #691: Passing `{ selected: true }` to `getByRole` of a radio / checkbox throws an error

Currently trying to find checked radio / checkbox throws an error `"aria-selected" is not supported on role "radio".`
I'd expect it to work for natively selected elements.

### Suggested implementation:

Just try to find by checked first.

Currently to get those, we need to `filter` a `getAllBy*` query:
```js
const selected = getAllByLabelText('preference').filter(control => control.checked);
```

The WAI-ARIA spec separates `aria-checked` from `aria-selected` so maybe we can introduce this change too.
Or just update the docs to link to both.
Suggested tab example after implementing the change:
> you can get the "Native"-tab by calling `getByRole('tab', { selected: true })`. To learn more about the selected state and which elements can have this state see [ARIA `aria-selected`](https://www.w3.org/TR/wai-aria-1.2/#aria-selected) or [ARIA `aria-checked`](https://www.w3.org/TR/wai-aria-1.2/#aria-checked).

idanen:
Closes #691 

**What**:

Allow querying checkboxes and radios that are checked using the `selected` option in `ByRole` queries.

**Why**:

Better experience than filtering for them

**How**:

Checking for both `aria-selected` _and_ `checked` property.

**Checklist**:

- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/ariaAttributes.js
+++ b/src/__tests__/ariaAttributes.js
@@ -40,6 +40,28 @@ test('`checked: true|false` matches `checked` elements with proper role', () => 
   expect(getByRole('checkbox', {checked: false})).toBeInTheDocument()
 })
 
+test('`checked: true|false` does not match element in `indeterminate` state', () => {
+  const {queryByRole, getByLabelText} = renderIntoDocument(
+    `<div>
+      <span role="checkbox" aria-checked="mixed">not so much</span>
+      <input type="checkbox" checked aria-label="indeteminate yes" />
+      <input type="checkbox" aria-label="indeteminate no" />
+    </div>`,
+  )
+  getByLabelText(/indeteminate yes/i).indeterminate = true
+  getByLabelText(/indeteminate no/i).indeterminate = true
+
+  expect(
+    queryByRole('checkbox', {checked: true, name: /indeteminate yes/i}),
+  ).toBeNull()
+  expect(
+    queryByRole('checkbox', {checked: true, name: /indeteminate no/i}),
+  ).toBeNull()
+  expect(
+    queryByRole('checkbox', {checked: true, name: /not so much/i}),
+  ).toBeNull()
+})
+
 test('`selected: true` matches `aria-selected="true"` on supported roles', () => {
   const {getAllByRole} = render(`
 <select>
--- a/src/__tests__/ariaAttributes.js
+++ b/src/__tests__/ariaAttributes.js
@@ -9,26 +9,26 @@ test('`selected` throws on unsupported roles', () => {
   )
 })
 
-test('`selected: true` matches `checked` checkboxes', () => {
+test('`checked: true|false` matches `checked` checkboxes', () => {
   const {getByRole} = renderIntoDocument(
     `<div>
       <input type="checkbox" checked />
       <input type="checkbox" />
     </div>`,
   )
-  expect(getByRole('checkbox', {selected: true})).toBeInTheDocument()
-  expect(getByRole('checkbox', {selected: false})).toBeInTheDocument()
+  expect(getByRole('checkbox', {checked: true})).toBeInTheDocument()
+  expect(getByRole('checkbox', {checked: false})).toBeInTheDocument()
 })
 
-test('`selected: true` matches `checked` elements with proper role', () => {
+test('`checked: true|false` matches `checked` elements with proper role', () => {
   const {getByRole} = renderIntoDocument(
     `<div>
       <span role="checkbox" aria-checked="true">✔</span>
       <span role="checkbox" aria-checked="false">𝒙</span>
     </div>`,
   )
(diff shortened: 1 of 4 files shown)


src/__tests__/ariaAttributes.js
```diff
@@ -9,6 +9,11 @@ test('`selected` throws on unsupported roles', () => {
   )
 })
 
+test('`selected: true` matches `checked` checkboxes', () => {
+  const {getByRole} = renderIntoDocument(`<input type="checkbox" checked />`)
+  expect(getByRole('checkbox', {selected: true})).toBeInTheDocument()
+})
```
kentcdodds, line 15: Could we get a test for the aria-checked version as well?

And it may help to make this test more complete if we include a second checkbox that is not checked.

nickmccurdy:
I'd prefer if this introduced a new option like `checked` (as suggested in https://github.com/testing-library/dom-testing-library/issues/691#issuecomment-657208117)

src/role-helpers.js
```diff
@@ -185,8 +185,37 @@ function computeAriaSelected(element) {
   if (element.tagName === 'OPTION') {
     return element.selected
   }
+  if ('checked' in element) {
```
eps1lon, line 188: Can we remove the logic for checking checked in `computeAriaSelected` now?

idanen:
Good point.
According to the [spec](https://html.spec.whatwg.org/multipage/input.html#checkbox-state-(type=checkbox)):
> The control is never a true tri-state control, even if the element's indeterminate IDL attribute is set to true. The indeterminate IDL attribute only gives the appearance of a third state.

Also from the spec:
> If the element's indeterminate IDL attribute is set to true, then the control's selection should be obscured as if the control was in a third, indeterminate, state.

So I don't really know how to move forward..
Any suggestions?

idanen:
Another point: the spec uses "positive selection" and "negative selection" when describing `checkedness` state, so maybe that's not such a bad idea to use the same API (`selected`)?

eps1lon:
> So I don't really know how to move forward..

How is `aria-checked` mapped from and to `<input type="checkbox" />`?

idanen:
From what I went over (`checked` spec, and WAI-ARIA for `aria-checked`), it's implied that:
The WAI-ARIA spec describes it like this:
> - When checked, the checkbox element has state aria-checked set to true.
> - When not checked, it has state aria-checked set to false.
> - When partially checked, it has state aria-checked set to mixed.

There's no direct "mapping", but from how I understand it should map like so:
- When checked, the checkbox element's `checked` property is set to `true`.
- When not checked, the checkbox element's `checked` property is set to `false`.
- When partially checked, the checkbox element's `checked` property is set to `false` and `indeterminate` property is set to `true`.

So when using `ByRole` with `{checked: true}` or `{checked: false}` it's pretty obvious, non should include `mixed`.
Do we want to add also `{indeterminate: true}`? Make the `checked` option tri-state?

This discussion makes me lean more towards `{selected: true}`, since it really helps users find controls by state, without diving too deep into the HTML spec.
If a label like "Please select one or more of these" is semantically correct, than querying with "selected" should also be correct.
WDYT?

eps1lon:
> checkbox role, with the aria-checked state set to "mixed" if the element's indeterminate IDL attribute is true, or "true" if the element's checkedness is true, or "false" otherwise

-- https://www.w3.org/TR/html-aam-1.0/#details-id-56

So `indeterminate=true` takes presedence.

src/role-helpers.js
```diff
@@ -198,6 +198,9 @@ function computeAriaChecked(element) {
   // implicit value from html-aam mappings: https://www.w3.org/TR/html-aam-1.0/#html-attribute-state-and-property-mappings
   // https://www.w3.org/TR/html-aam-1.0/#details-id-56
   // https://www.w3.org/TR/html-aam-1.0/#details-id-67
+  if ('indeterminate' in element && element.indeterminate) {
+    return false
```
eps1lon, line 202: This seems confusing to me. A mixed checkbox is neither check nor unchecked. Right now we consider it unchecked as far as I can tell.

idanen: I agree that it's confusing, but that's the spec.
A checkbox in indeterminate state is neither checked not unchecked.
What do you suggest?

idanen: Sorry, missed the second part of your comment.
You're right, I had a mistake in the test.
Pushing a fix

eps1lon: It's fine to not be able to query mixed checkboxes for now. But returning indeterminate ones for `checked: false` would be incorrect.

kentcdodds:
@all-contributors please add @idanen for code and tests

kentcdodds:

PR #701, opened 2020-07-16
fix: Ignore .idea folder

getspooky:
What ? 
WebStorm creates an .idea folder as you start a project. This folder contains some important settings and configurations, like syntax highlighting settings, VCS mapping, debug configuration, and so on. Sometimes user details such as username and password also come up in the .idea folder.
It's important to ignore this folder 

- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@ node_modules
 coverage
 dist
 .DS_Store
-
+.idea
 # these cause more harm than good
 # when working with contributors
 package-lock.json


smeijer:
Thanks for the effort.

The thing is, this can easily get fixed by using the global `.gitignore` file. If we would make an exception for the `.idea` folder, then we end up doing the same for all other editors that place some configuration in the project folders as well. Or even special operating files such as `thumbs.db` on windows?

For what it's worth, I've done the same in the past (see 617).

As we won't be merging this, I'm going to close it. Thanks again for taking the effort to contribute though. Much appreciated!

Issue #707, opened 2020-07-20
getByLabelText does not work with some IDs


ryuuji3:
- `@testing-library/dom` version: 7.21.2
- `@testing-library/react` version: 10.2.1
- Testing Framework and version: jest 24.8.0
  
- DOM Environment: default
  

### Relevant code or config:
This was originally discovered in react testing library, so here's example usage

Edit: I discovered it wasn't the id/for combination but aria-labelledby that was the problem. I updated the experiment

```js
const { findByLabelText } = render(
   <div>
        <label id="eighth.label">8rd</label>
        <input aria-labelledby="eighth.label" id="eighth.id" />
    </div>
)

expect(getByLabelText(/test/i).toHaveDisplayValue("2") // SyntaxError: '[id=failing.test]' is not a valid selector
```

### What you did:

Attempting to find an input by its label to assert on it, but the ID would normally need to be wrapped in double quotes since its not compatible with # selector syntax.

### What happened:

A SyntaxError about an invalid CSS selector

### Problem description:

This is a regression from the behavior that worked in 7.20.2 (that I am locally running).

### Suggested solution:
https://github.com/testing-library/dom-testing-library/blob/44e1d5c36f0cade02dac1275d7eabb14d3a743d2/src/queries/label-text.js#L84

```js
const labellingElement = container.querySelector(`[id=${labelId}]`)
```
should become
```js
const labellingElement = container.querySelector(`[id="${labelId}"]`)
```

I might be able to submit a PR but you might get to it faster than I can, since I am not available to tonight

ryuuji3:
Working on a PR now. Never contributed before so this is a good first issue

PR #708, opened 2020-07-20
Issue 707- getByLabelText does not work with some IDs when using aria-labelledby

ryuuji3:
**What**: When using aria-labelledby, getByLabelText should account for a selector that might need quotes around it.

**Why**: In such a case, the selector is invalid causing unexpected test failures 

**How**: Just a one line change to wrap quotes around the id in the selector.

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

Closes https://github.com/testing-library/dom-testing-library/issues/707


kentcdodds:
@all-contributors please add @ryuuji3 for code and tests

ryuuji3:
Sweet, thank you very much!

Issue #717, opened 2020-07-25
"Cannot read property 'test' of undefined." when undefined passed to matches


JoshuaKGoldberg:
- `@testing-library/dom` version: 7.19.0
- Testing Framework and version: `jest`@26.1.0 and `@testing-library/react`@10.4.7
- DOM Environment: `jsdom`@11.12.0

### Relevant code or config:

```js
const view = render(<div />);

view.getByRole(undefined);
```

### What you did:

In a project that does not use TypeScript _(alas!)_, I passed `undefined` instead of a valid matcher to `getByRole`.

### What happened:

```
TypeError: Cannot read property 'test' of undefined

       9 |     const view = render(<div />);
      10 | 
    > 11 |     view.getByRole(undefined);
         |          ^
      13 | 

      at matches (../../node_modules/@testing-library/dom/dist/matches.js:39:20)
      at ../../node_modules/@testing-library/dom/dist/queries/role.js:78:47
          at Array.some (<anonymous>)
      at ../../node_modules/@testing-library/dom/dist/queries/role.js:78:26
          at Array.filter (<anonymous>)
      at queryAllByRole (../../node_modules/@testing-library/dom/dist/queries/role.js:57:54)
      at ../../node_modules/@testing-library/dom/dist/query-helpers.js:89:17
      at ../../node_modules/@testing-library/dom/dist/query-helpers.js:64:17
      at getByRole (../../node_modules/@testing-library/dom/dist/query-helpers.js:108:19)
```

### Reproduction:

https://codesandbox.io/s/react-testing-library-getbytext-undefined-r6ygs for a version showing `null` and `getByText` causing the same issue.

### Problem description:

For all matchers that eventually call to `matches`, including `getByText` and `getByRole`, passing an invalid matcher such as `null` or `undefined` crashes with a confusing error message. Users likely did not mean to pass an invalid matcher to the search.

https://github.com/testing-library/dom-testing-library/blob/639294c6e24ee9f06f214c57383e202b4c3b14b4/src/matches.js#L27

### Suggested solution:

Similar to #609, I'd like to send a PR that adds an explicit error message for this case.

kentcdodds:
Sure 👍 thanks!

PR #718, opened 2020-07-25
Added explicit error message for null or undefined matchers

JoshuaKGoldberg:
**What**:

Added a nice explicit error message for the case in #717 of `null` or `undefined` as a `matcher`.

**Why**:

Better error messages = happier users! 💝

**How**:

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

As before: I'm not attached to this particular error message, if a reviewer can think of something better that'd be nice.

### DIFF
--- a/src/matches.js
+++ b/src/matches.js
@@ -1,7 +1,17 @@
+function assertNotNullOrUndefined(matcher) {
+  if (matcher == null) {
+    throw new Error(
+      `It looks like you passed ${matcher} instead of a matcher. Did you do something like getByText("${matcher}")?`,
+    )
+  }
+}
+
 function fuzzyMatches(textToMatch, node, matcher, normalizer) {
   if (typeof textToMatch !== 'string') {
     return false
   }
+
+  assertNotNullOrUndefined(matcher)
 
   const normalizedText = normalizer(textToMatch)
   if (typeof matcher === 'string') {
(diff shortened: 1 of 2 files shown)


src/matches.js
```diff
@@ -1,8 +1,18 @@
+function assertNotNullOrUndefined(matcher) {
+  if (matcher == null) {
+    throw new Error(
+      `It looks like you passed ${matcher} instead of a matcher. Did you do something like getByText("${matcher}")?`,
```
kentcdodds, line 4: Let's do what we can to avoid language that could be aggravating for sometime struggling to make something work. Thanks!

```suggestion
      `It looks ${matcher} was passed instead of a matcher. Did you do something like getByText("${matcher}")?`,
```

JoshuaKGoldberg: Ooh, great tone suggestion, thank you! I love little improvements like that 😄

nickmccurdy: I prefer the grammar of the original, how about this?
```suggestion
      `It looks like ${matcher} was passed instead of a matcher. Did you do something like getByText("${matcher}")?`,
```

src/matches.js
```diff
@@ -1,8 +1,18 @@
+function assertNotNullOrUndefined(matcher) {
+  if (matcher == null) {
+    throw new Error(
+      `It looks like ${matcher} was passed instead of a matcher. Did you do something like getByText("${matcher}")?`,
```
nickmccurdy, line 4: Shouldn't this be without quotes?
```suggestion
      `It looks like ${matcher} was passed instead of a matcher. Did you do something like getByText(${matcher})?`,
```

kentcdodds:
@all-contributors please add @JoshuaKGoldberg for code and tests

Issue #716, opened 2020-07-25
getByLabelText stopped returning div associated with label via the "for" attribute


ipap360:
- `@testing-library/dom` version: > 7.21.1
- Testing Framework and version: jest 25.5.4
- DOM Environment: jsdom v16
- Node.JS: v12.18.1

### Relevant code or config:

![form-control](https://user-images.githubusercontent.com/11017666/88461659-4e799680-ce9d-11ea-9cd6-53ef794443d1.PNG)

### What you did:

getByLabelText("Data product:")

### What happened:

TestingLibraryElementError: Found a label with the text of: Data product:, however no form control was found associated to that label. Make sure you're using the "for" attribute or "aria-labelledby" attribute correctly.

### Reproduction:

https://codesandbox.io/s/react-testing-library-demo-v8wzb?file=/src/hello.js
Tried to reproduce this here but I couldn't; it seems to be working. Interestingly, this happened to us as well when the sub-dependency testing-library/dom that testing-library/react depends on was stuck to a lower version.

### Problem description:

Semantically the "for" attribute associates labels with form controls (input, select etc) however in practice, we use several more advanced controls, often wrapped into div containers (e.g. react-select). 

Until v7.21.1, testing-library/dom supported this "non-standard" use of the label which is convenient.

### Suggested solution:

If making this more semantically accurate is a design choice by the authors, it is of course understandable. But we'll need to change the code for these advanced components to use "aria-labelledby" instead (which creates some inconvenience)

Thanks in advance

timdeschryver:
This seems like a duplicate of #714 ?

kentcdodds:
Perhaps we could add a more helpful error message for this?

ipap360:
Apologies, I thought I searched the entire internet and missed one of the most recent issues!

kentcdodds:
No worries. It's a tricky one to Google for. Would you like to help by adding a more helpful error for this case?

ipap360:
Sure, it'd be my pleasure!
I presume that it should cover labels associated with any non-labellable elements and not just with divs, right?

kentcdodds:
Yep 👍

eps1lon:
Closed in #720

PR #720, opened 2020-07-27
fix(ByLabelText): improve error message when label is associated with…

ipap360:
… non-labellable elements (#716)

**What**: 

Add a more specific error message whenever getByLabelText/getAllByLabelText cannot find an associated element because the element is non-labellable.

**Why**:

The generic error message wasn't clear on why it was failing and whether this was changed on purpose (given that it was previously working).

**How**:

Whenever a label exists but no form control association is found, do an additional check for any other elements associated with the label via the "for" attribute. If there are, print specific error messages on why the query is failing and how to resolve this issue.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

My very first open source contribution 😊 
Really happy that it is for a great project that I rely so much on 🥳

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -407,6 +407,114 @@ test('label with children with no form control', () => {
 `)
 })
 
+test('label with non-labellable element', () => {
+  const {getByLabelText, queryByLabelText} = render(`
+  <div>
+    <label for="div1">Label 1</label>
+    <div id="div1">
+      Hello
+    </div>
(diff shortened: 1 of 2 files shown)


ipap360:
> Looks great to me! Just need to figure out the coverage issue. Looks like we may be missing a branch of code in the tests.

Sorry about that. Changed an existing test slightly to cover this branch too. 😇

kentcdodds:
@all-contributors please add @ipap360 for code and tests

Issue #721, opened 2020-07-28
Configure type missing documented fields


thepelkus-too:
- `@testing-library/dom` version: 7.18.1
- Testing Framework and version: jest 24.9
  
- DOM Environment: jsdom 16
  

### Relevant code or config:

```js
import { configure } from '@testing-library/dom';

configure({
  getElementError: (message, container) => {
    const error = new Error(message);
    error.name = 'TestingLibraryElementError';
    return error;
  },
});
```

### What you did:

I was attempting to customize error reporting by way of swapping out the getElementMethod on the configuration.

### What happened:

Getting the following typescript error:

```
Argument of type '{ getElementError: (message: any, container: any) => Error; }' is not assignable to parameter of type 'Partial<Config> | ConfigFn'. 
Object literal may only specify known properties, and 'getElementError' does not exist in type 'Partial<Config> | ConfigFn'.",
```

### Reproduction:

https://codesandbox.io/s/react-testing-library-demo-6q10i

### Problem description:

Typescript is reporting errors on configuration fields that should be valid according to the documentation.

### Suggested solution:

I'd be happy to open a PR to add the documented fields! Just wanted to open this issue first to make sure it was valid.

thepelkus-too:
Closed by PR #724

PR #724, opened 2020-07-29
Add missing config fields to type

thepelkus-too:
**What**:

Adding fields to the Config type to support the options that are documented.

**Why**:

Without adding these fields to the types, using them in projects which support types will cause errors or require exceptions to type enforcement. Example provided in issue #721 

**How**:

Adding fields to existing Config type.

**Checklist**:

- [N/A] Tests
- [x] Ready to be merged
      

I'll caveat this with saying that I'm not sure exactly how the type tests work. Would it make sense to add a function there that attempts to set each of the fields (or all of the fields at once) on the configuration to verify that there aren't type errors there?


.opt-in
```diff
@@ -0,0 +1 @@
+pre-commit
```
nickmccurdy, line 1: I don't think this file should be committed

thepelkus-too: Oof. Messy commit on my part. Thanks for catching that.

PR #725, opened 2020-07-30
Fixed syntax error for node version 8

StevenMatchett:
**Node 8 syntax error fix**:

**Breaks Node 8 as it see not having a parm at the end of a except block as a syntax error**:

**How**:

Added a parm on the end of a catch

**Checklist**:

-  [N/A] Tests
-  [N/A] Typescript definitions updated
- [ ] Ready to be merged

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -40,7 +40,7 @@ function getJestFakeTimersType() {
       // jest.getRealSystemTime is only supported for Jest's `modern` fake timers and otherwise throws
       jest.getRealSystemTime()
       return 'modern'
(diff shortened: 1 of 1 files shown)


eps1lon:
This should be something that is fixed upstream in kcd-scripts since we should rather transpile it to support older environments. Though we're not supporting node 8 anymore. Does this cause problems in environments other than node 8?

kentcdodds:
Thanks for this but I'm afraid we don't support Node 8 as it has reach its end of life and is no longer supported by the node team.

PR #726, opened 2020-08-02
fix(ByLabelText): support IE and other legacy browsers

### LINKED ISSUE #723: ByLabelText does not work with IE 11
- `@testing-library/dom` version: 7.21.6
- Testing Framework and version: any
- DOM Environment: IE11 or Firefox 52

### Relevant code or config:

```html
<label for="rating-test-2">2 Stars</label>
<input id="rating-test-test" type="radio" />
```

### What you did:

`getByLabelText('2 Stars')`

### What happened:

```bash
  Found a label with the text of: 2 Stars, however the element associated with this label (<input />) is non-labellable [https://html.spec.whatwg.org/multipage/forms.html#category-label]. If you really need to label a <input />, you can use aria-label or aria-labelledby instead.
```

### Reproduction:

[`55e72b2` (#21944)](https://github.com/mui-org/material-ui/pull/21944/commits/55e72b2dbedff899da7e966bb277e1a06c40d44f)

### Problem description:

[`HTMLInputElement.labels`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/labels) is not supported in IE 11 or prior to Firefox 56

### Suggested solution:

Use CSS selector. It's even used later in the code and it's not clear why we use two different approaches.

maxnewlands:
**What**: 
`*ByLabelText` is updated to support IE11 and other legacy browsers.

**Why**:
 Closes #723.

**How**:
`HTMLInputElement.prototype.labels` is polyfilled. [Existing test covering primary use cases of getByLabelText](https://github.com/testing-library/dom-testing-library/blob/2c57f06cf76a3f388e943628de7aba6eb5d6afe1/src/__tests__/element-queries.js#L147) is copied into a new test, where the `labels` property is mocked to simulate absence of support.

**Checklist**:

- [ ] ~Documentation added to the
      [docs site](https://github.com/testing-library/testing-library-docs)~ N/A
- [x] Tests
- [ ] ~Typescript definitions updated~ N/A
- [x] Ready to be merged
      

This PR stops short of polyfilling `HTMLLabelElement.prototype.control`, which [appears to be well-supported](https://developer.mozilla.org/en-US/docs/Web/API/HTMLLabelElement/control#Browser_compatibility) in older browsers.

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -1186,7 +1186,7 @@ test('return a proper error message when no label is found and there is an aria-
 // https://github.com/testing-library/dom-testing-library/issues/723
 it('gets form controls by label text on IE and other legacy browsers', () => {
   // Simulate lack of support for HTMLInputElement.prototype.labels
-  const inputLabelsSpy = jest
+  jest
     .spyOn(HTMLInputElement.prototype, 'labels', 'get')
     .mockReturnValue(undefined)
 
@@ -1241,6 +1241,4 @@ it('gets form controls by label text on IE and other legacy browsers', () => {
   expect(getByLabelText('6th one 6th two 6th three').id).toBe('sixth-id')
   expect(getByLabelText('7th one').id).toBe('seventh-id')
   expect(getByLabelText('8th one').id).toBe('eighth.id')
-
-  inputLabelsSpy.mockRestore()
 })
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -1191,54 +1191,10 @@ it('gets form controls by label text on IE and other legacy browsers', () => {
     .mockReturnValue(undefined)
 
   const {getByLabelText} = renderIntoDocument(`
-    <div>
-      <label>
-        1st<input id="first-id" />
-      </label>
-      <div>
-        <label for="second-id">2nd</label>
-        <input id="second-id" />
-      </div>
-      <div>
-        <label id="third-label">3rd</label>
-        <input aria-labelledby="third-label" id="third-id" />
-      </div>
-      <div>
-        <label for="fourth.id">4th</label>
-        <input id="fourth.id" />
-      </div>
-      <div>
-      <div>
-        <label id="fifth-label-one">5th one</label>
-        <label id="fifth-label-two">5th two</label>
-        <input aria-labelledby="fifth-label-one fifth-label-two" id="fifth-id" />
-      </div>
-      <div>
-        <input id="sixth-label-one" value="6th one"/>
-        <input id="sixth-label-two" value="6th two"/>
-        <label id="sixth-label-three">6th three</label>
-        <input aria-labelledby="sixth-label-one sixth-label-two sixth-label-three" id="sixth-id" />
-      </div>
-      <div>
-        <span id="seventh-label-one">7th one</span>
-        <input aria-labelledby="seventh-label-one" id="seventh-id" />
-      </div>
-      <div>
-        <label id="eighth.label">8th one</label>
-        <input aria-labelledby="eighth.label" id="eighth.id" />
-      </div>
-    </div>
+    <label>
+      Label text
+      <input id="input-id" />
+    </label>
   `)
-  expect(getByLabelText('1st').id).toBe('first-id')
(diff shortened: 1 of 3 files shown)


src/__tests__/element-queries.js
```diff
@@ -1182,3 +1182,65 @@ test('return a proper error message when no label is found and there is an aria-
...
+  expect(getByLabelText('5th one').id).toBe('fifth-id')
+  expect(getByLabelText('5th two').id).toBe('fifth-id')
+  expect(getByLabelText('6th one').id).toBe('sixth-id')
+  expect(getByLabelText('6th two').id).toBe('sixth-id')
+  expect(getByLabelText('6th one 6th two').id).toBe('sixth-id')
+  expect(getByLabelText('6th one 6th two 6th three').id).toBe('sixth-id')
+  expect(getByLabelText('7th one').id).toBe('seventh-id')
+  expect(getByLabelText('8th one').id).toBe('eighth.id')
+
+  inputLabelsSpy.mockRestore()
```
kentcdodds, line 1245: If this test fails, then the mock won't be restored and other tests could be impacted.

What we should do instead, is add `jest.restoreAllMocks` in place of the line in our setup here: https://github.com/testing-library/dom-testing-library/blob/54aebf7ae2d70b8b4a187e44a31360bc8a130826/tests/setup-env.js#L40

That should handle this case and everything else.

src/__tests__/element-queries.js
```diff
@@ -1182,3 +1182,65 @@ test('return a proper error message when no label is found and there is an aria-
 </div>"
 `)
 })
+
+// https://github.com/testing-library/dom-testing-library/issues/723
+it('gets form controls by label text on IE and other legacy browsers', () => {
+  // Simulate lack of support for HTMLInputElement.prototype.labels
+  const inputLabelsSpy = jest
+    .spyOn(HTMLInputElement.prototype, 'labels', 'get')
+    .mockReturnValue(undefined)
```
kentcdodds, line 1191: This is a great idea!

src/__tests__/element-queries.js
```diff
@@ -1182,3 +1182,65 @@ test('return a proper error message when no label is found and there is an aria-
...
+
+  const {getByLabelText} = renderIntoDocument(`
```
kentcdodds, line 1193: I'm not sure how necessary it is to copy/paste this from the other test. I think a single label that situation that covers the code path we're testing should suffice.

eps1lon: Ideally we'd use the same test that is run with a test matrix covering different environments i.e. run with 
- `.labels` and `.control` (modern browser
- no `.labels` but `.control` (IE 11)
- neither `.lables` nor `.control` (edge 17)

I guess we can set something up like this in another PR where we leverage different jest environments that mock out problematic DOM methods we're aware of.

src/__tests__/element-queries.js
maxnewlands: Got it! I didn't realize this approach could hurt test isolation, but it makes plenty of sense. 🌠

src/__tests__/element-queries.js
maxnewlands: Thanks, but all credit goes to @eps1lon's [reference implementation](https://github.com/eps1lon/dom-accessibility-api/pull/352/files#diff-227df8dcfa07fb56cfb8e394fb09ada7R443) 🙂

src/__tests__/element-queries.js
maxnewlands: > neither .lables nor .control (edge 17)

Shoot, I overlooked this. Appreciate you calling it out, @eps1lon ! 

I'll push my changes that reduce this to one test case for now. I also like the idea of running a comprehensive set of tests in different environments, and if it's preferable to take that approach instead of what's in this PR, please feel free to close this.

kentcdodds: Nah, let's get this in then we can add that in a future PR.

maxnewlands: Sure thing. Then if there are no other comments, this just needs an approval. 🙏

kentcdodds:
@all-contributors please add @maxnewlands for code and tests

PR #733, opened 2020-08-08
Pr/440 show line in debug

### LINKED ISSUE #440: show line in debug()
I'm not sure if this is even possible, but figured I'd throw it out there. When I use `console.log('foo')` the console will include the line number where the console log was made. However, when using debug() it's always pointing to the line in dom-testing-library (see screenshot).  
image

I'm not even sure if its possible to do this, but it would be so great if instead showed the line in *my code* where debug was called.

victorandcode:
**What**:
Add the call frame to the output of `screen.debug` before the prettified DOM

**Why**:
For easier debugging. 
Resolves #440 

**How**:
- Create an empty Error and filter it's stack frames to keep only the one we care about
- Coloring the message with chalk

**Before**
Screenshot from 2020-08-08 21-00-53

**After**
Screenshot from 2020-08-08 20-53-46

Here's the code for the test
Screenshot from 2020-08-08 20-54-56

**Checklist**:

- [X] Tests
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/get-user-code-frame.js
+++ b/src/__tests__/get-user-code-frame.js
@@ -69,10 +69,6 @@ test('it returns only user code frame when node code frames are present afterwar
 })
 
 test("it returns empty string if file from code frame can't be read", () => {
-  const consoleWarnSpy = jest
-    .spyOn(global.console, 'warn')
-    .mockImplementationOnce(jest.fn)
-
   // Make fire read purposely fail
   fs.readFileSync.mockImplementationOnce(() => {
     throw Error()
@@ -83,8 +79,4 @@ test("it returns empty string if file from code frame can't be read", () => {
   globalErrorMock.mockImplementationOnce(() => ({stack}))
 
   expect(getUserCodeFrame(stack)).toEqual('')
-  expect(consoleWarnSpy).toHaveBeenCalledTimes(1)
-  expect(consoleWarnSpy).toHaveBeenCalledWith(
-    `Couldn't read file /sample-error/error-example.js for displaying the code frame`,
-  )
 })
--- a/src/__tests__/get-user-code-frame.js
+++ b/src/__tests__/get-user-code-frame.js
@@ -17,8 +17,7 @@ jest.mock('fs', () => ({
   ),
 }))
 
-const userStackFrame =
-  'at somethingWrong (/home/john/projects/sample-error/error-example.js:7:14)'
+const userStackFrame = 'at somethingWrong (/sample-error/error-example.js:7:14)'
 
 let globalErrorMock
 
@@ -31,16 +30,16 @@ afterEach(() => {
   global.Error.mockRestore()
 })
 
-test('it returns only client code frame when code frames from node_modules are first', () => {
+test('it returns only user code frame when code frames from node_modules are first', () => {
   const stack = `Error: Kaboom
-      at Object.<anonymous> (/home/john/projects/projects/sample-error/node_modules/@es2050/console/build/index.js:4:10)
+      at Object.<anonymous> (/sample-error/node_modules/@es2050/console/build/index.js:4:10)
       ${userStackFrame}
   `
   globalErrorMock.mockImplementationOnce(() => ({stack}))
   const userTrace = getUserCodeFrame(stack)
 
   expect(userTrace).toMatchInlineSnapshot(`
-    "/home/john/projects/sample-error/error-example.js:7:14
+    "/sample-error/error-example.js:7:14
       5 |         document.createTextNode('Hello world')
       6 |       )
     > 7 |       screen.debug()
@@ -49,18 +48,18 @@ test('it returns only client code frame when code frames from node_modules are f
   `)
 })
 
-test('it returns only client code frame when node code frames are present afterwards', () => {
+test('it returns only user code frame when node code frames are present afterwards', () => {
   const stack = `Error: Kaboom
-      at Object.<anonymous> (/home/john/projects/projects/sample-error/node_modules/@es2050/console/build/index.js:4:10)
+      at Object.<anonymous> (/sample-error/node_modules/@es2050/console/build/index.js:4:10)
       ${userStackFrame}
-      at Object.<anonymous> (/home/user/Documents/projects/sample-error/error-example.js:14:1)
+      at Object.<anonymous> (/sample-error/error-example.js:14:1)
       at internal/main/run_main_module.js:17:47
   `
   globalErrorMock.mockImplementationOnce(() => ({stack}))
   const userTrace = getUserCodeFrame()
 
   expect(userTrace).toMatchInlineSnapshot(`
-    "/home/john/projects/sample-error/error-example.js:7:14
+    "/sample-error/error-example.js:7:14
       5 |         document.createTextNode('Hello world')
       6 |       )
     > 7 |       screen.debug()
@@ -86,6 +85,6 @@ test("it returns empty string if file from code frame can't be read", () => {
   expect(getUserCodeFrame(stack)).toEqual('')
   expect(consoleWarnSpy).toHaveBeenCalledTimes(1)
   expect(consoleWarnSpy).toHaveBeenCalledWith(
-    `Couldn't read file /home/john/projects/sample-error/error-example.js for displaying the code frame`,
+    `Couldn't read file /sample-error/error-example.js for displaying the code frame`,
   )
 })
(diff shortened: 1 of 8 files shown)


src/get-user-trace.js
```diff
@@ -0,0 +1,21 @@
...
+}
+
+function getUserTrace() {
+  const err = new Error()
+  const firstClientCodeFrame = err.stack
+    .split('\n')
+    .slice(1) // Remove first line which has the form "Error: TypeError"
+    .find(frame => !frame.includes('node_modules/')) // Ignore frames from 3rd party libraries
+
+  return chalk.gray(`${getFrameLocation(firstClientCodeFrame)}\n`)
```
eps1lon, line 18: How does this look like in a browser? If I understand https://github.com/chalk/chalk/issues/1 correctly then `chalk` doesn't work in a browser. Would be nice to have confirmation that `@testing-library/dom` still works outside of `node`.

victorandcode:
After doing some research based on the feedback, I believe we have 2 options to implement this feature :factory_worker: 

## If we only want to display the frame like `location/to/my/file.js:10:2`

**This can only be achieved in node** since we are counting on ignoring external packages inside node_modules. Also, we need the environment (jest), to calculate correctly the line of the error with source mapping.

## If we want to display the code frame as suggested by @kentcdodds 

This feature can be achieved via `@babel/code-frame` + node's `fs` module. I don't see any alternative to do this on the browser, therefore, **it would be node exclusive like the previous option**.

In order to achieve this, we would need to do some tweaking in the build config to mock those 2 modules I mentioned so these don't cause any trouble in the browser.


Let me know if you feel like any of those 2 options is worth pursuing :)

kentcdodds:
I like option 2. I don't think we'd need anything special in config. We could do something like this: https://github.com/testing-library/react-testing-library/blob/master/src/flush-microtasks.js#L28

src/get-user-code-frame.js
```diff
@@ -0,0 +1,67 @@
...
+let codeFrameColumns = null
+
+try {
+  const nodeRequire = module && module.require
+  readFileSync = nodeRequire.call(module, 'fs').readFileSync
+  codeFrameColumns = nodeRequire.call(module, '@babel/code-frame')
+    .codeFrameColumns
+} catch {
+  // We're in a browser environment
+  /* istanbul ignore next */
```
victorandcode, line 12: Given that these lines run while the module is loaded, I don't know of a practical way to mock them for a single test to force them to throw. That's why I'm ignoring coverage for now. If you know of a way to do this, I'd appreciate the guidance :pray:

victorandcode:
Option 2 is now implemented and code frames are now displayed :tada: Please take a look when you have some time. I also tested the UMD build in chrome and it's working fine.

Here's an example:
Screenshot from 2020-08-16 11-18-02
Screenshot from 2020-08-16 11-18-24

kentcdodds:
This is fantastic! I'll give the code a deeper look later, but one thing: could we put the codeframe at the bottom?

kentcdodds:
Oh and the line above the codeframe appears to be listing the wrong file, it should show the one that is shown in the codeframe right?

victorandcode:
> Oh and the line above the codeframe appears to be listing the wrong file, it should show the one that is shown in the codeframe right?

So actually that line comes from jest :(. I have looked into it but I haven't found a way to prevent it.

victorandcode:
> Oh and the line above the codeframe appears to be listing the wrong file, it should show the one that is shown in the codeframe right?

This has now been fixed. Here's the new version:
Screenshot from 2020-08-16 15-50-31

kentcdodds:
Super! Let's add another line between the debug output and the line listing where the file is, and let's also make that line muted like the console.log line. Thanks!

victorandcode:
> Super! Let's add another line between the debug output and the line listing where the file is, and let's also make that line muted like the console.log line. Thanks!

This is now fixed and looks like this :eyes: 

I still need to take a look at the CI error, it shouldn't be failing since locally it doesn't with the same script :thinking: 

Screenshot from 2020-08-16 20-48-25

src/__tests__/get-user-code-frame.js
```diff
@@ -0,0 +1,91 @@
...
+  globalErrorMock = jest.spyOn(global, 'Error')
+})
+
+afterEach(() => {
+  global.Error.mockRestore()
+})
+
+test('it returns only client code frame when code frames from node_modules are first', () => {
+  const stack = `Error: Kaboom
+      at Object.<anonymous> (/home/john/projects/projects/sample-error/node_modules/@es2050/console/build/index.js:4:10)
```
kentcdodds, line 36: This says `john/projects/projects`, but the snapshot just says `john/projects/` 🤔

src/__tests__/screen.js
```diff
@@ -1,6 +1,10 @@
 import {screen} from '..'
 import {renderIntoDocument} from './helpers/test-utils'
 
+jest.mock('../get-user-code-frame', () => ({
+  getUserCodeFrame: () => '',
+}))
+
```
kentcdodds, line 7: Why is this mocked here and not all the other tests? What's special about this one?

src/get-user-code-frame.js
```diff
@@ -0,0 +1,70 @@
...
+    frameLocationElements[0],
+    parseInt(frameLocationElements[1], 10),
+    parseInt(frameLocationElements[2], 10),
+  ]
+
+  let rawFileContents = ''
+  try {
+    rawFileContents = readFileSync(filename, 'utf-8')
+  } catch {
+    console.warn(`Couldn't read file ${filename} for displaying the code frame`)
```
kentcdodds, line 38: Let's ditch this warning and change it to a comment instead:

```suggestion
    // Couldn't read file for some reason
```

src/get-user-code-frame.js
```diff
@@ -0,0 +1,70 @@
...
+  codeFrameColumns = nodeRequire.call(module, '@babel/code-frame')
+    .codeFrameColumns
+  chalk = nodeRequire.call(module, 'chalk')
+} catch {
+  // We're in a browser environment
+  /* istanbul ignore next */
+  console.warn(
+    'Printing the user trace is not supported in a browser environment',
+  )
+}
```
kentcdodds, line 19: Let's not bother warning. This is a nice-to-have and folks wouldn't expect this feature anyway.

```suggestion
} catch {
  // We're in a browser environment
}
```

src/__tests__/get-user-code-frame.js
victorandcode: On this particular test, that frame should actually be ignored since it's inside node_modules. The one we care about is one line after `${userStackFrame}`.
However, the frame that is ignored does have a typo (`/projects/projects`). I've corrected this and simplified the paths used for all tests to avoid confusion. They all now start with `/sample-error` instead of `/home/john/<etc>`

src/__tests__/screen.js
victorandcode: Given that the test `exposes debug method` internally calls `logDOM`, which in turn calls `getUserCodeFrame`, we need to mock it so it doesn't add clutter to the output of the screen's tests snapshots

kentcdodds: Gotcha. Let's write a comment explaining that please. Thanks!

victorandcode: Done :)

src/get-user-code-frame.js
victorandcode: This is now fixed

src/get-user-code-frame.js
victorandcode: This one as well

src/get-user-trace.js
victorandcode: These changes are now working in the browser as well since we're loading modules conditionally if we're in node :partying_face:

victorandcode:
Hey @kentcdodds, please let me know if the latest changes look ok.

All comments have been addressed but the [travis error](https://travis-ci.org/github/testing-library/dom-testing-library/jobs/722441803) still puzzles me. It fails in a seemingly unrelated file (role.js) but passes locally.

I also tried temporarily rolling back my changes, creating a [very simple commit ](https://github.com/testing-library/dom-testing-library/commit/d6f7f8a77e7963f84fc0eacde9ebdfe6ac0ac338) but the error still happened.

I was wondering if this had happened before

kentcdodds:
Yeah, that's odd. I'm not sure what's up with.

Anyone else want to investigate?

kentcdodds:
Tried it locally myself and everything worked fine. So I'm going to just hope this is an issue with CI for this PR and we'll go ahead and merge it. I'll deal with it if it persists beyond this PR CI issue.

kentcdodds:
Looks like there was a weird issue in CI. Cleared cache and ran again and everything's fine now.

kentcdodds:
@all-contributors please add @victorandcode for code and tests

victorandcode:
> Looks like there was a weird issue in CI. Cleared cache and ran again and everything's fine now.

Thanks for taking a look :)

Issue #744, opened 2020-08-18
<li> is not an accessible element


nullromo:
- `@testing-library/dom` version: 7.22.0
- Testing Framework and version: jest version 24.9.0
- DOM Environment: jsdom version 11.12.0

### Relevant code or config:

```js
// spinnerButton.test.js
import React from 'react';
import { render, screen } from '@testing-library/react';
import SpinnerButton from '../spinnerButton.js';

test("renders a spinner when it's loading", () => {
    // render the button
    render(<SpinnerButton text={''} waiting={true} />);

    // expect a spinner to exist
    expect(screen.getByRole('status')).toBeInTheDocument();
});
```

```js
// spinnerButton.js
import React from 'react';
import Spinner from 'react-bootstrap/Spinner';
import './styles/spinnerButton.css';

class SpinnerButton extends React.Component {
    render = () => {
        const buttonProps = { ...this.props };
        delete buttonProps.waiting;
        delete buttonProps.text;
        delete buttonProps.spinnerClassName;
        return (
            <button {...buttonProps}>
                {this.props.waiting ? (
                    <Spinner
                        as='li'
                        size='sm'
                        animation='border'
                        role='status'                     /* <= RELEVANT LINE OF CODE */
                        className={this.props.spinnerClassName}
                        aria-hidden='true'
                    >
                        <span></span>
                    </Spinner>
                ) : (
                    this.props.text
                )}
            </button>
        );
    };
}

export default SpinnerButton;
```

### What you did:

I have a React component, `SpinnerButton` that shows some text before it has been clicked and a spinning wheel after it has been clicked. I gave a DOM node in that button a `role` called `'status'`.

Then I tried to create a test that would find that element by its role, but it cannot be found.

### What happened:

The test output looks like this:

```
 renders a spinner when it's loading

    TestingLibraryElementError: Unable to find an accessible element with the role "status"

    Here are the accessible roles:

      button:

      Name "":
      <button />

      --------------------------------------------------

    <body>
      <div>
        <button>
          <li
            aria-hidden="true"
            class="spinner-border spinner-border-sm"
            role="status"
          >
            <span />
          </li>
        </button>
      </div>
    </body>

      29 |
      30 |     // expect a spinner to exist
    > 31 |     expect(screen.getByRole('status')).toBeInTheDocument();
         |                   ^
      32 | });
      33 |
```

So I know that the `<li>` element is there and that it has `role="status"` set.

[...]

PR #745, opened 2020-08-19
fix: suggest hidden:true with getByRole if element is not accessible 

marcosvega91:
**What**: If an element is inaccessible `getSuggestedQuery` should suggest to use `hidden:true` with `getByRole` 

**Why**: Answering to this issue https://github.com/testing-library/dom-testing-library/issues/744, I wanted to share a link of `testing-playground` but the query was not suggested as I expected.  

**How**: If the element is inaccessible and the query is `getByRole` I have added to `queryArgs` an `hidden:true`

**Checklist**:

- [X] Tests
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/suggestions.js
+++ b/src/__tests__/suggestions.js
@@ -549,3 +549,18 @@ test('should get the first label with aria-labelledby contains multiple ids', ()
     variant: 'get',
   })
 })
+
+test('should suggest hidden option if element is not in the accessibilty tree', () => {
+  const {container} = renderIntoDocument(`
+    <input type="text" aria-hidden=true />
+  `)
+
+  expect(
+    getSuggestedQuery(container.querySelector('input'), 'get', 'role'),
+  ).toMatchObject({
+    queryName: 'Role',
+    queryMethod: 'getByRole',
+    queryArgs: ['textbox', {hidden: true}],
+    variant: 'get',
+  })
+})
--- a/src/suggestions.js
+++ b/src/suggestions.js
@@ -2,7 +2,7 @@ import {computeAccessibleName} from 'dom-accessibility-api'
 import {getDefaultNormalizer} from './matches'
 import {getNodeText} from './get-node-text'
 import {DEFAULT_IGNORE_TAGS, getConfig} from './config'
-import {getImplicitAriaRoles} from './role-helpers'
+import {getImplicitAriaRoles, isInaccessible} from './role-helpers'
 
 const normalize = getDefaultNormalizer()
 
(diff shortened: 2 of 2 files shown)


eps1lon (changes requested):
It shouldn't just suggest it but rather say that the/a element is not part of the a11y tree. 

The element being removed from the a11y tree has strong implications for users relying on screen readers. It's equivalent to the element being invisible for a sighted user.

I fear people will just randomly add aria-hidden and query for it if we explicitly recommend it.

marcosvega91:
Yes it was something to discuss. I think that we should add this to testing playground too. Maybe we should do a console.warn? While on testing playground show a text ?

marcosvega91:
what do you think of this ?
I have added a `warning` prop with a message, maybe this will be useful for other things. On testing-playground side we can show the warning message

p.s. I have added some types to `suggestions` 

```diff
diff --git a/src/suggestions.js b/src/suggestions.js
index f584e7b..0d23434 100644
--- a/src/suggestions.js
+++ b/src/suggestions.js
@@ -39,6 +39,8 @@ function getRegExpMatcher(string) {
 }
 
 function makeSuggestion(queryName, element, content, {variant, name}) {
+  let warning = ''
+
   const queryArgs = [
     queryName === 'Role' || queryName === 'TestId'
       ? content
@@ -51,6 +53,10 @@ function makeSuggestion(queryName, element, content, {variant, name}) {
 
   if (queryName === 'Role' && isInaccessible(element)) {
     queryArgs.push({hidden: true})
+    warning = `Element is inaccessible. This means that the element and all his children are invisible to screen readers.
+    If you are using aria-hidden prop make sure this is the right choice for your case.
+    `
+    console.warn(warning)
   }
 
   const queryMethod = `${variant}By${queryName}`
@@ -60,6 +66,7 @@ function makeSuggestion(queryName, element, content, {variant, name}) {
     queryMethod,
     queryArgs,
     variant,
+    warning,
     toString() {
       let [text, options] = queryArgs
 
diff --git a/types/suggestions.d.ts b/types/suggestions.d.ts
index 17e48d6..8c1e68c 100644
--- a/types/suggestions.d.ts
+++ b/types/suggestions.d.ts
@@ -1,5 +1,13 @@
+export type QueryArg = {
+  [key: string]: RegExp | boolean
+}
+
 export interface Suggestion {
   queryName: string
+  queryMethod: string
+  queryArgs: QueryArg[]
+  variant: string
+  warning: string
   toString(): string
 }
```

nullromo:
```
-    warning = `Element is inaccessible. This means that the element and all his children are invisible to screen readers.
-    If you are using aria-hidden prop make sure this is the right choice for your case.

+    warning = `Element is inaccessible. This means that the element and all its children are invisible to screen readers.
+    If you are using the aria-hidden prop, make sure this is the right choice for your case.
```

eps1lon:
> what do you think of this ? [...]

-- https://github.com/testing-library/dom-testing-library/pull/745#issuecomment-676369200

:+1:  That is way clearer. There's always some risk involved but I hope this won't cause too much damage.

marcosvega91:
I have added the message as we discussed above 👍

eps1lon (approved):

PR #748, opened 2020-08-20
fix: update dom-accessibility-api

### LINKED ISSUE #747: getByRole can't find button by name if title is ""
- `@testing-library/dom` version: 7.22.2
- Testing Framework and version: Jest (whatever version comes with CodeSandbox)
  
- DOM Environment: whatever version comes with CodeSandbox
  

### Relevant code or config:

```js
import React from "react";

export default function App() {
  return <button title="">OK</button>;
}
```

### What you did:

```js
screen.getByRole("button", { name: /ok/i });
```

### What happened:

```text
Unable to find an accessible element with the role "button" and name `/ok/i`

  button:

  Name "":
  <button
    title=""
  />
```

### Reproduction:

https://codesandbox.io/s/happy-chandrasekhar-311qt?file=/src/App.js

### Problem description:

Chrome reports that the button has a name:

image

But `getByRole` says it does not.

JSFiddle: https://jsfiddle.net/91retmv2/

marcosvega91:
fix #747 

**What**: This PR will update dom-accessibility-api to the latest version

**Why**: Because in the new version there is a fix for the issue.

The update will solve a problem where title should be ignored when is empty while using name option

**How**: 

**Checklist**:

- [X] Tests
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -543,3 +543,10 @@ describe('configuration', () => {
     }
   })
 })
+
+test('should get button using his text content when title is empty', () => {
+  const {getByRole} = renderIntoDocument(
+    '<button title="">click on me</button>',
+  )
+  expect(getByRole('button', {name: 'click on me'})).toBeInTheDocument()
+})
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -543,10 +543,3 @@ describe('configuration', () => {
     }
   })
 })
-
-test('should get button using his text content when title is empty', () => {
-  const {getByRole} = renderIntoDocument(
-    '<button title="">click on me</button>',
-  )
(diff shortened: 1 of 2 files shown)


src/__tests__/role.js
```diff
@@ -543,3 +543,10 @@ describe('configuration', () => {
     }
   })
 })
+
+test('should get button using his text content when title is empty', () => {
```
eps1lon, line 547: I wouldn't add these tests. There's certainly an argument to be made that the dependency is an implementation detail but the point of dependencies is also to cover certain functionality. Otherwise why not copy the whole test suite from dom-accessibility-api?

marcosvega91: Yes you're right, is a duplicate 👍

eps1lon (approved):

PR #753, opened 2020-08-24
fix: getByTitle returning element matching textContent instead of title

### LINKED ISSUE #741: getByTitle finds any element text?
- `@testing-library/dom` version: 7.22.2
- Testing Framework and version: Jest 26.0.10
- DOM Environment: jsdom 11.12.0

### Relevant code or config:

```js
// component
export default () => <div title="WrongTitle">HelloWorld</div>;

// test
const wantTitle = "HelloWorld";
const { getByTitle } = render(<Hello />);

// I would expect this to fail here
expect(getByTitle(wantTitle)).toBeInTheDocument();

// ...but it fails here
expect(getByTitle(wantTitle).title).toBe(wantTitle);
```

### What you did:

I have a React component with some text, and a `title` attribute that matches that text. I was modifying the component (and tests) to have a different `title` (but the same text) when I disabled it.

### What happened:

I started getting unexpected behavior from my tests because they were expecting the title to have changed, and `getByTitle` was finding the element text.

### Reproduction:

[CodeSandbox](https://codesandbox.io/s/tl-getbytitle-finding-element-text-srtmy)

### Problem description:

Maybe I'm missing something, but it seems very unintuitive that this line:
`expect(getByTitle(wantTitle).title).toBe(wantTitle)`
can fail on the `toBe`. If it's going to fail, shouldn't it fail on the `getByTitle`?

### Suggested solution:

Looking at `queries/title.js`, I see this part:
```js
return Array.from(container.querySelectorAll('[title], svg > title')).filter(
    node =>
      matcher(node.getAttribute('title'), node, text, matchNormalizer) ||
      matcher(getNodeText(node), node, text, matchNormalizer),
  )
```
I assume the `matcher(getNodeText(node)` part is intended for `svg` `title` elements, which makes sense, but it looks to me like this will find the text for any element. Forcing that second condition to only be true for `title` elements would solve it, I think.

wyze:
Fix #741 

**What**: The `getByTitle` query was meant to match the `title` attribute or `svg > title` element. The logic was implemented that it would also match the text content of any element containing a `title` attribute.

**Why**: We want to correct the behavior of the query. It should only match text content of the `svg > title` element and not any element with a `title` attribute.

**How**: Add a condition of the text content matcher to ensure the node we are checking is `svg > title` element.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -558,7 +558,7 @@ test('query/get element by its title', () => {
   expect(getByTitle('Delete').id).toEqual('2')
   expect(queryByTitle('Delete').id).toEqual('2')
   expect(queryByTitle('Del', {exact: false}).id).toEqual('2')
-  expect(queryByTitle("HelloWorld")?.id).not.toEqual('4')
+  expect(queryByTitle("HelloWorld")).toBeNull()
 })
 
 test('query/get title element of SVG', () => {
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -551,12 +551,14 @@ test('query/get element by its title', () => {
         <span title="Ignore this" id="1"/>
         <span title="Delete" id="2"/>
         <span title="Ignore this as well" id="3"/>
+        <div title="WrongTitle" id="4">HelloWorld</div>
     </div>
   `)
 
   expect(getByTitle('Delete').id).toEqual('2')
   expect(queryByTitle('Delete').id).toEqual('2')
   expect(queryByTitle('Del', {exact: false}).id).toEqual('2')
+  expect(queryByTitle("HelloWorld")?.id).not.toEqual('4')
 })
 
 test('query/get title element of SVG', () => {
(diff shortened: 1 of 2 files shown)


src/__tests__/element-queries.js
```diff
@@ -551,12 +551,14 @@ test('query/get element by its title', () => {
...
         <span title="Delete" id="2"/>
         <span title="Ignore this as well" id="3"/>
+        <div title="WrongTitle" id="4">HelloWorld</div>
     </div>
   `)
 
   expect(getByTitle('Delete').id).toEqual('2')
   expect(queryByTitle('Delete').id).toEqual('2')
   expect(queryByTitle('Del', {exact: false}).id).toEqual('2')
+  expect(queryByTitle("HelloWorld")?.id).not.toEqual('4')
```
kentcdodds, line 561: Let's avoid conditionals in tests. We're assuming that this will return `null` right? So should this be:

```suggestion
  expect(queryByTitle("HelloWorld")).toBeNull()
```

wyze: Yup, that woorks too!

kentcdodds:
@all-contributors please add @awesomeunleashed for bugs

kentcdodds:
@all-contributors please add @wyze for tests

PR #764, opened 2020-09-10
fix(screen): broken link when there's no document

MatanBobi:
**What**: Fix a broken link when there's no document defined 

**Why**: The link is broken

**How**: I've put a link to the screen api definition. I couldn't find the page that was there before in the docs repo neither in the commits history.

**Checklist**:

- [X] Tests
- [ ] Typescript definitions updated - N/A
- [X] Ready to be merged

### DIFF
--- a/src/__node_tests__/screen.js
+++ b/src/__node_tests__/screen.js
@@ -4,6 +4,6 @@ test('the screen export throws a helpful error message when no global document i
   expect(() =>
     screen.getByText(/hello world/i),
   ).toThrowErrorMatchingInlineSnapshot(
-    `"For queries bound to document.body a global document has to be available... Learn more: https://testing-library.com/s/screen-global-error"`,
+    `"For queries bound to document.body a global document has to be available... Learn more: https://testing-library.com/docs/dom-testing-library/api-queries#screen"`,
   )
 })
--- a/src/screen.js
+++ b/src/screen.js
@@ -14,7 +14,7 @@ export const screen =
         (helpers, key) => {
           helpers[key] = () => {
             throw new TypeError(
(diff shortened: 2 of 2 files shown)


eps1lon (changes requested):
I think we just want to update the redirect in testing-library-docs. Links in error messages should always work since some users might not be able to upgrade the library.

MatanBobi:
> I think we just want to update the redirect in testing-library-docs. Links in error messages should always work since some users might not be able to upgrade the library.

That's a good point. I'll close this one and add it to the docs. Thanks!

eps1lon:
We don't need this anymore, right? The link is now valid.

MatanBobi:
> We don't need this anymore, right? The link is now valid.

The link is valid but as I wrote above, I think that this fix should be in both places.. IMO we shouldn't rely on a redirect to always exist and should fix this here also.. WDYT?

eps1lon:
We're actually less backwards compatible. Note that we can't change these links once published. If we decide to move the page then any version published will have 404 links. We could then decide to make redirect but we might as well use a permalink to begin with. Otherwise we have to maintain redirects for every published version of the library.

I believe this is the same rationale that React is using for links in dev warnings. All of these links redirect to the actual page.

MatanBobi:
> We're actually less backwards compatible. Note that we can't change these links once published. If we decide to move the page then any version published will have 404 links. We could then decide to make redirect but we might as well use a permalink to begin with. Otherwise we have to maintain redirects for every published version of the library.
> 
> I believe this is the same rationale that React is using for links in dev warnings. All of these links redirect to the actual page.

That's a good point, I didn't think about it that way..
So I'll close this one :)
Thanks!

PR #777, opened 2020-10-06
fix: element error delegation

Tismas:
**What**: Using `getElementError` from configuration in `waitFor` and `singleQuries` so that user can customize error reporting consistently.

**Why**: Currently when configuring getElementError it works differently in for example `getByText` vs `waitFor(() => getByText)`.  

**How**: Instead of calling prettyDOM directly I changed it so that it uses `getElementError` from configuration.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/wait-for.js
+++ b/src/__tests__/wait-for.js
@@ -1,4 +1,5 @@
 import {waitFor} from '../'
+import {configure, getConfig} from '../config'
 import {renderIntoDocument} from './helpers/test-utils'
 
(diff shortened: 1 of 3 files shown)


alexkrolick (approved):

kentcdodds (approved):
Perfect. Thank you 👏

kentcdodds:
@all-contributors please add @Tismas for code and tests

PR #782, opened 2020-10-12
feat(waitFor): allow returning a promise

kentcdodds:
You can now return a promise in the waitFor callback and your callback
will not be called again until the promise rejects. If the promise
rejects, then the callback will be called again until it gets a promise
which resolves or it times out.

**What**: feat(waitFor): allow returning a promise

**Why**: So we don't have to do weird things like this: https://github.com/kentcdodds/bookshelf/pull/83/files#diff-f3caa0e4f827bc632c53d30ddb2aeb60R26-R29

Specifically, we need to `waitFor` something, but the callback has to do something asynchronous to determine whether we can continue our test. Today, `waitFor` does not support this... Until now...

**How**: Store a promiseStatus and if the waitFor callback returns a "thennable" then we attach our own handler onto the promise and update the promiseStatus to be "pending". So long as the promiseStatus is "pending" we will not call the callback again.

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs/pull/642)
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/wait-for.js
+++ b/src/__tests__/wait-for.js
@@ -2,6 +2,15 @@ import {waitFor} from '../'
 import {configure, getConfig} from '../config'
 import {renderIntoDocument} from './helpers/test-utils'
 
+function deferred() {
+  let resolve, reject
+  const promise = new Promise((res, rej) => {
+    resolve = res
+    reject = rej
+  })
+  return {promise, resolve, reject}
+}
+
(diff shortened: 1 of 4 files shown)


marcosvega91 (approved):
For me is good :)

kentcdodds:
@all-contributors please add @marcosvega91 for review

PR #784, opened 2020-10-13
fix: error when selecting element without 'matches' fn

smeijer:
**What**:
Add an extra check to `getSuggestedQuery` to make sure that the element has a `matches` handler, before calling it.

**Why**:
Not all DOM elements have a `matches` handler. The one that I've encountered is `document`. I feel the question rising: "_But who tests `document`_"? Well, this can happen when entering iframes. Or when hovering the document in testing-playground (chrome extension).

**How**:
```diff
  if (
+    typeof element.matches !== 'function' ||
    element.matches(DEFAULT_IGNORE_TAGS)
  ) {
```

**Checklist**:

- [n/a Documentation added to the
- n/a Tests N/A
- n/a Typescript definitions updated
- [x] Ready to be merged

### DIFF
--- a/src/suggestions.js
+++ b/src/suggestions.js
@@ -96,7 +96,10 @@ function canSuggest(currentMethod, requestedMethod, data) {
 
 export function getSuggestedQuery(element, variant = 'get', method) {
   // don't create suggestions for script and style elements
-  if (element.matches(DEFAULT_IGNORE_TAGS)) {
+  if (
+    typeof element.matches !== 'function' ||
+    element.matches(DEFAULT_IGNORE_TAGS)
+  ) {
     return undefined
   }
 


eps1lon (changes requested):
This would be a new feature not a fix since we explicitly only accept HTML elements (see https://github.com/testing-library/dom-testing-library/blob/8eebd9e5b264ed8435fd3f531a7d0eb112d3faa9/types/suggestions.d.ts#L42-L46).

Additionally, we should add some test. Especially some integration tests that show how this branch is hit. Why would we not want to suggest a query to get the document?

smeijer:
> Why would we not want to suggest a query to get the document?

I don't know. Is there any value of selecting the document node? testing-library motivates to select "humanly visible elements", document doesn't fit in that spirit.

I only came across this while hovering elements with the chrome extension. And because of that, I was looking for a quick fix. I only now see that the types declare that only `HTMLElement` types should be selected.

What's wisdom here? If we upgrade this to a feature, it means that we need to change the type, and add proper support to the selectors / suggestions if that doesn't already work.

If we treat it as a bug, then the question is if it really is a bug, because that types should cover it at compile time. 

I can handle the error in `testing-playground`, by adding a check before I'm feeding the hovered element to `getSuggestedQuery`. So I'm okay with closing this PR.

eps1lon:
We first need to identify the issue. Right now we're fixing symptoms. Might be that the testing-playground needs to guard against it and not testing-library.

kentcdodds:
> I can handle the error in testing-playground, by adding a check before I'm feeding the hovered element to getSuggestedQuery. So I'm okay with closing this PR.

Yeah, I think that's the best path forward 👍  Thanks!

PR #789, opened 2020-10-15
fix: getByLabelText for output

### LINKED ISSUE #787: ByLabelText does not work properly with output element
- `@testing-library/dom` version: 7.24.5
- Testing Framework and version: jest@26.5.2
  
- DOM Environment: jsdom@16.4.0
  

### Relevant code or config:

```js
describe('labeled output', () => {
  it('passes', () => {
    const root = render('<label>foo <output></output></label>');

    expect(getByLabelText(root, 'foo')).toBeInstanceOf(HTMLOutputElement);
  });

  it('fails', () => {
    const root = render('<label>foo <output>bar</output></label>');

    expect(getByLabelText(root, 'foo')).toBeInstanceOf(HTMLOutputElement);
  });

  it('passes too', () => {
    const root = render('<label>foo <output>bar</output></label>');

    expect(getByLabelText(root, 'foo bar')).toBeInstanceOf(HTMLOutputElement);
  });
});

function render(html: string) {
  const root = document.createElement('div');

  root.innerHTML = html;

  return root;
}
```

### What you did:

I want to find `output` element by its label

### What happened:

When `output` has text content it is not ignored during the scan.

### Reproduction:

- 

### Problem description:

It's impossible to find output by its label. 

### Suggested solution:

I think similar solution to [this](https://github.com/testing-library/dom-testing-library/blob/53829b8651bcea86410de73eeb81751b0f2ccae6/src/queries/label-text.js#L29) would be ok.

marcosvega91:
**What**: I have fixed #787 and the comment in https://github.com/testing-library/dom-testing-library/issues/343#issuecomment-555385756

**Why**: Because following specs it should be possible to get `output` by it's label. 

**How**: Instead of replacing string that could cause the issue in the comment https://github.com/testing-library/dom-testing-library/issues/343#issuecomment-555385756 I have looped al the childNodes taking text only from the right nodes

**Checklist**:

- [X] Tests
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -293,7 +293,7 @@ test('returns the labelable element control inside a label', () => {
       <output />
       <progress />
       <select />
-      <textarea />
+      <textarea ></textarea>
     </label>
   `)
 
@@ -558,7 +558,7 @@ test('query/get element by its title', () => {
   expect(getByTitle('Delete').id).toEqual('2')
   expect(queryByTitle('Delete').id).toEqual('2')
   expect(queryByTitle('Del', {exact: false}).id).toEqual('2')
-  expect(queryByTitle("HelloWorld")).toBeNull()
+  expect(queryByTitle('HelloWorld')).toBeNull()
 })
 
 test('query/get title element of SVG', () => {
@@ -1199,4 +1199,24 @@ it('gets form controls by label text on IE and other legacy browsers', () => {
     </label>
   `)
   expect(getByLabelText('Label text').id).toBe('input-id')
+})
+
+// https://github.com/testing-library/dom-testing-library/issues/787
+it(`get the output element by it's label`, () => {
+  const {getByLabelText} = renderIntoDocument(`
+    <label>foo
+      <output>bar</output>
+    </label>
+  `)
+  expect(getByLabelText('foo')).toBeInTheDocument()
+})
+
+// https://github.com/testing-library/dom-testing-library/issues/343#issuecomment-555385756
+it(`should get element by it's label when there are elements with same text`, () => {
+  const {getByLabelText} = renderIntoDocument(`
+    <label>test 1
+      <textarea defaultValue="test"></textarea>
+    </label>
+  `)
+  expect(getByLabelText('test 1')).toBeInTheDocument()
 })
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -1224,7 +1224,7 @@ it(`get the output element by it's label`, () => {
 it(`should get element by it's label when there are elements with same text`, () => {
   const {getByLabelText} = renderIntoDocument(`
     <label>test 1
-      <textarea defaultValue="test"></textarea>
+      <textarea>test</textarea>
     </label>
   `)
   expect(getByLabelText('test 1')).toBeInTheDocument()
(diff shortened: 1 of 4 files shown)


src/queries/label-text.js
```diff
@@ -51,15 +36,38 @@ function queryAllLabelsByText(
...
+    'output',
+    'progress',
+    'select',
+    'textarea',
+    'input',
+  ]
+  if (
+    labelledNodeNames.some(nodeName => node.nodeName.toLowerCase() === nodeName)
+  ) {
+    return ''
```
delca85, line 53: Just a silly comment, could maybe be more clear to write the if condition like this:
`labelledNodeNames.includes(node.nodeName.toLowerCase())`?

marcosvega91: yes, thanks :)

src/queries/label-text.js
```diff
@@ -51,15 +36,38 @@ function queryAllLabelsByText(
...
+  return Array.from(node.childNodes)
+    .map(childNode => getTextContent(childNode))
+    .join('')
+}
+
+function getLabelContent(node) {
+  let textContent
+  if (node.tagName.toLowerCase() === 'label') {
+    textContent = getTextContent(node)
+  } else {
```
delca85, line 65: A question jumps in my mind (I beg you pardon if it is a stupid one): is the desired component that only the `label` elements go through `getTextContent` to obtain their value? Will never happen that a labelling element (not a `label` one) needs to be processed by `getTextContent`?

marcosvega91: I don't think because the else case should be covered only using the  `aria-labelledby`. in that case we should take everything

delca85: Got it! Thank you!

src/queries/label-text.js
```diff
@@ -38,15 +48,7 @@ function queryAllLabelsByText(
 
 function getTextContent(node) {
   const TEXT_NODE = 3
```
delca85, line 50: I know you moved `labelledNodeNames` because it is used by `queryAllByLabelText` too but maybe you could move outside `getTextContent` the declaration of `TEXT_NODE` too.

Anyway this PR looks legit to me! 👍 😄

marcosvega91: hahaha yes

src/__tests__/element-queries.js
```diff
@@ -1200,3 +1200,32 @@ it('gets form controls by label text on IE and other legacy browsers', () => {
...
+  `)
+
+  expect(getByLabelText('foo')).toBeInTheDocument()
+})
+
+// https://github.com/testing-library/dom-testing-library/issues/343#issuecomment-555385756
+it(`should get element by it's label when there are elements with same text`, () => {
+  const {getByLabelText} = renderIntoDocument(`
+    <label>test 1
+      <textarea defaultValue="test"></textarea>
```
alexkrolick, line 1227: I don't think "defaultValue" is a valid DOM attribute for `textarea` - it's a React concept
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

marcosvega91: Yes you're right. I only added the end tag because it was not right. But I think that it should be fixed

kentcdodds (approved):
LGTM. Thanks!

PR #790, opened 2020-10-16
refactor: export label helpers to his own file and use both in suggestions and query

marcosvega91:
**What**: Moved label helpers to his own file

**Why**: suggestions was using his logic to get label content that was not working with `aria-labelledby`

**How**: Using the logic implemented by @delca85 both for suggestions and label query

**Checklist**:

- [ ] Tests
- [X] Ready to be merged


src/suggestions.js
```diff
@@ -113,7 +90,9 @@ export function getSuggestedQuery(element, variant = 'get', method) {
     })
   }
 
-  const labelText = getLabelTextFor(element)
+  const labelText = getLabels(document, element)
+    .map(label => label.content)
+    .join(' ')
```
kentcdodds, line 95: This looks like a behavior change. Is this a bug fix or a new feature?

kentcdodds (commented):
Looks pretty good. Just one thing.

src/suggestions.js
marcosvega91: ooopss The space was a mistake

marcosvega91: ooopss The space was a mistake

kentcdodds: We should probably have a test for this. But still this looks like a change from the original behavior which is concerning as well. Maybe I'm missing something though.

marcosvega91: I have added the method `getLabels` that could return ore or more labels. Every label should be separated by space because they are regarding to different label object. I have added a test for it

kentcdodds (dismissed):
Ok, I think it looks good to me. I'd love another reviewer here though. @delca85?

delca85:
> Ok, I think it looks good to me. I'd love another reviewer here though. @delca85?

I will be happy to review it but I am going to look at it in a couple of hours, ok?

marcosvega91:
Is not a problem ☺️ thanks for the help @delca85

src/suggestions.js
delca85: Should this fix the problem related to the suggestion in case of concatenated labels?

marcosvega91: Yes. I have added a test to prove this

src/label-helpers.js
```diff
@@ -0,0 +1,55 @@
...
+        const textToMatch = getLabelContent(label)
+        const formControlSelector =
+          'button, input, meter, output, progress, select, textarea'
+        const labelledFormControl = Array.from(
+          label.querySelectorAll(formControlSelector),
+        ).filter(formControlElement => formControlElement.matches(selector))[0]
+
+        return {content: textToMatch, formControl: labelledFormControl}
+      })
+  labelsList = labelsList.filter(Boolean)
```
delca85, line 50: I am not sure this line is meaningful written in this way. No element of `labelsList` could be `undefined` or `null`.
In the code you are replacing, this filter was performed to remove the empty element of `labelsList`, but I think you are doing this check at line 79 in `src/queries/label-text`.
So, maybe you could remove this line and keep the filter in `src/queries/label-text` or replace this line with `labelsList = labelsList.filter(label => Boolean(label.content))` and remove the filter in `src/queries/label-text`.

marcosvega91: Yes right I can remove

delca85 (dismissed):

delca85:
I approved this one. I think using the same logic in both suggestion and get labels logic is really a good idea! As well as fixing the suggestion defect related to label concatenation, for sure.

kentcdodds:
Looks like @marcosvega91's other PR made a conflict with this one. Could you get that fixed @marcosvega91? Thanks!

marcosvega91:
I have rebased the branch 👍  :)

kentcdodds (approved):
Excellent. Thank you!

Issue #800, opened 2020-10-30
Test really slow because of MutationObserver not disconnecting directly


romain-trotard:
- `@testing-library/dom` version: 7.26.3
- Testing Framework and version:
`@testing-library/react` version: 11.1.0
`jest` version: 26.6.1
- DOM Environment:
`jsdom` version: 16.4.0

Hello guys :)

### What happened:

My tests are really slow.

### Reproduction:

I made a little codesandbox: https://codesandbox.io/s/mutationobserver-reproduction-pe9gf?file=/src/__tests__/index.js where you can see my problem.

The issue is that the callback of waitFor is called even when the element has been found:
![image](https://user-images.githubusercontent.com/17161484/97722999-8eb76900-1acb-11eb-9c62-8fff9c334cb3.png)

The console log is printed thrice instead of one:
![image](https://user-images.githubusercontent.com/17161484/97723071-a858b080-1acb-11eb-834a-a484fd0f25da.png)

### Problem description:

The MutationObserver is disconnected at the end of the test instead of directly when the [onDone callback](https://github.com/testing-library/dom-testing-library/blob/master/src/wait-for.js#L91) is called. I guess this is because the thread is not available until the end of the test.
These unnecessary callback's calls (in the MutationObserver) make my tests really slow as you can see below:

![image](https://user-images.githubusercontent.com/17161484/97723439-11d8bf00-1acc-11eb-9885-045233664b01.png)

### Suggested solution:

I have tested to directly disconnect the observer instead of inside the setImmediate method (https://github.com/testing-library/dom-testing-library/blob/master/src/wait-for.js#L97). I get really faster tests:
```
if (!usingFakeTimers) {
    clearInterval(intervalId)
    observer.disconnect()
}
```

![image](https://user-images.githubusercontent.com/17161484/97724098-cb379480-1acc-11eb-82f8-c2b6768d6c5a.png)

 
Is there a reason to disconnect the observer inside setImmediate instead of synchronously ?
Am I doing something wrong ?
Or maybe I could make a PR to make the disconnection configurable globally (choice would be between synchronously and asynchronously) ?

kentcdodds:
Hi @romain-trotard,

Thanks for this! I'd love to speed this up. Here's the PR that added the `setImmediate` behavior: https://github.com/testing-library/dom-testing-library/pull/102

If you can figure out how to avoid the problem that PR solved *and* make it faster, I'd welcome a PR for that.

[...]

PR #801, opened 2020-11-01
Disconnect MutationObserver synchronously in wait-for

romain-trotard:
Here is a little PR after opening this issue https://github.com/testing-library/dom-testing-library/issues/800
(my investigation is on it)

**What**:

My test are slow because of MutationObersion disconnections made at the end of my tests.

**Why**:

Currently `observer.disconnect()`is called in setImmediate, I guess the main thread is not available until the end of the test, so the check callback is called too much times.
This was done in this PR #102 

**How**:

All my explanations are visible here: https://github.com/testing-library/dom-testing-library/issues/800#issuecomment-720140316

After making many test, I didn't see any problem to call both: 
```
setImmediate(() => observer.disconnect());
observer.disconnect();
```

**Checklist**:

- [ ] Tests N/A
- [] Ready to be merged: I made a change on the husky config to be able to commit. Do not merge that like that!!

### DIFF
--- a/.huskyrc.js
+++ b/.huskyrc.js
@@ -1,1 +1,0 @@
-module.exports = require('kcd-scripts/husky')
--- a/.huskyrc.js
+++ b/.huskyrc.js
@@ -1,0 +1,1 @@
+module.exports = require('kcd-scripts/husky')
--- a/src/wait-for.js
+++ b/src/wait-for.js
@@ -95,10 +95,6 @@ function waitFor(
       if (!usingFakeTimers) {
         clearInterval(intervalId)
         observer.disconnect()
-
-        // Call also the disconnect method on setImmediate for projects not using the MutationObersver implementation of jsdom.
-        // But using for example `@sheerun/mutationobserver-shim` which  provokes infinite loop when being called immediately on `onDone` function
-        setImmediate(() => observer.disconnect())
       }
 
       if (error) {
--- a/src/wait-for.js
+++ b/src/wait-for.js
@@ -94,6 +94,10 @@ function waitFor(
 
       if (!usingFakeTimers) {
         clearInterval(intervalId)
+        observer.disconnect()
+
+        // Call also the disconnect method on setImmediate for projects not using the MutationObersver implementation of jsdom.
+        // But using for example `@sheerun/mutationobserver-shim` which  provokes infinite loop when being called immediately on `onDone` function
         setImmediate(() => observer.disconnect())
       }
 


.huskyrc.js
```diff
@@ -1 +0,0 @@
-module.exports = require('kcd-scripts/husky')
```
romain-trotard, line 1: @kentcdodds I removed this to be able to commit. I had an infinite loader that you can see in this little video. (I am using ubuntu 18)
infiniteLoading
Did you ever heard of any problem of this kind ?

eps1lon: `git commit --no-verify` avoids any pre-commit hooks. Any change to the pre-commit hook should be dealt with  in a separate issue/pr.

romain-trotard: Oh thanks for the command, I totally forgot it exists ^^
I push directly the revert :)
(by the way I didn't expect to merge this change on master, it was only to be able to commit hoping someone will help me like you did ;) )
Otherwise do you have any idea why could the test run infinitely ? (in the hook)

eps1lon: No idea. I have husky disabled globally since it doesn't work well with my workflow (commit often, commit failing tests etc).

romain-trotard:
The CI failed because of the new node v15, it is waiting for a response after the `npx codecov@3` command: 
image

src/wait-for.js
```diff
@@ -94,6 +94,10 @@ function waitFor(
 
       if (!usingFakeTimers) {
         clearInterval(intervalId)
+        observer.disconnect()
+
+        // Call also the disconnect method on setImmediate for projects not using the MutationObersver implementation of jsdom.
+        // But using for example `@sheerun/mutationobserver-shim` which  provokes infinite loop when being called immediately on `onDone` function
```
kentcdodds, line 100: This looks fine to me. I'm only concerned about the implications of calling disconnect twice. I guess it's fine, but I'm not sure it's worthwhile having code that is only around to support a shim that nobody should be using (because they should be using a modern version of jsdom or a real browser environment). Anyone want to argue against just removing the next line?

marcosvega91: I have commented and on the issue and proposing two solutions:
- remove the line as you suggest
- check if we are using MutationObserver of `jsdom` or another custom impl
https://github.com/testing-library/dom-testing-library/issues/800#issuecomment-720379391
https://github.com/testing-library/dom-testing-library/issues/800#issuecomment-720437358

romain-trotard: If nobody should use shim, it's good for me to remove it.
Thank you Kent and Marco, I'm gonna remove it asap 

Does someone know what I should do for the breaking CI ? (if there is something I can do)

.huskyrc.js
romain-trotard: Ahaha no problem :)

kentcdodds:
I'm fixing the build failure now: https://github.com/testing-library/dom-testing-library/pull/802

kentcdodds:
I'm not sure why, but travis seems to be taking forever to run our builds. I'll look at this again tomorrow.

kentcdodds:
@All-contributors please add @romain-trotard for code

PR #804, opened 2020-11-03
Fix null labels on hidden inputs

thomasmarshall:
**What**:

This PR ensures `*LabelText` helpers do not throw an error when there are hidden inputs.

**Why**:

The `labels` property on `input` elements of type `hidden` is [`null` rather than `NodeList`](https://html.spec.whatwg.org/multipage/forms.html#dom-lfe-labels). This meant the `getRealLabels` function would return `null` causing `queryAllByLabelText` to throw an error when it tried to [call `length` on `null`](https://github.com/testing-library/dom-testing-library/blob/62f4e5e09a4b81ef66679560b540523edccdef98/src/queries/label-text.js#L52).

I noticed this issue when using `cy.findByLabelText` (from [@testing-library/cypress](https://github.com/testing-library/cypress-testing-library)) – which fails with the following exception on pages with hidden inputs:

```
Timed out retrying: Cannot read property 'length' of null
```

**How**:

This fixes the issue by ensuring the element is labelable before calling `labels` on it. There was already an `isLabelable` guard clause in place, but it was only checked _after_ returning `element.labels` (which is `null` for hidden inputs). Alternative solutions might be to add a different guard clause to return an empty array if `element.labels` is `null`, or to check the return value from `getRealLabels` before calling `length` on it.

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs) **N/A**
- [x] Tests
- [ ] Typescript definitions updated **N/A**
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/label-helpers.js
+++ b/src/__tests__/label-helpers.js
@@ -1,0 +1,7 @@
+import {getRealLabels} from '../label-helpers'
+
+test('hidden inputs are not labelable', () => {
+  const element = document.createElement('input')
+  element.type = 'hidden'
+  expect(getRealLabels(element)).toEqual([])
+})
--- a/src/label-helpers.js
+++ b/src/label-helpers.js
@@ -34,9 +34,9 @@ function getLabelContent(node) {
 
 // Based on https://github.com/eps1lon/dom-accessibility-api/pull/352
 function getRealLabels(element) {
-  if (element.labels !== undefined) return element.labels
-
   if (!isLabelable(element)) return []
+
+  if (element.labels !== undefined) return element.labels
 
   const labels = element.ownerDocument.querySelectorAll('label')
   return Array.from(labels).filter(label => label.control === element)
--- a/src/label-helpers.js
+++ b/src/label-helpers.js
@@ -34,9 +34,9 @@ function getLabelContent(node) {
 
 // Based on https://github.com/eps1lon/dom-accessibility-api/pull/352
 function getRealLabels(element) {
-  if (!isLabelable(element)) return []
+  if (element.labels !== undefined) return element.labels ?? []
 
-  if (element.labels !== undefined) return element.labels
+  if (!isLabelable(element)) return []
 
   const labels = element.ownerDocument.querySelectorAll('label')
   return Array.from(labels).filter(label => label.control === element)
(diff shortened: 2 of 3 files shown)


src/__tests__/label-helpers.js
```diff
@@ -0,0 +1,7 @@
+import {getRealLabels} from '../label-helpers'
+
+test('hidden inputs are not labelable', () => {
+  const element = document.createElement('input')
+  element.type = 'hidden'
+  expect(getRealLabels(element)).toEqual([])
```
kentcdodds, line 6: What was the result before your changes? 🤔

kentcdodds (commented):
Thanks for the contribution! One question please.

src/__tests__/label-helpers.js
thomasmarshall: Before the change it returns `null`:

Screenshot 2020-11-03 at 17 44 01

marcosvega91: I didn't see the PR and I have opened a PR on the same behaviour because I saw an issue about it.

src/label-helpers.js
```diff
@@ -34,10 +34,10 @@ function getLabelContent(node) {
 
 // Based on https://github.com/eps1lon/dom-accessibility-api/pull/352
 function getRealLabels(element) {
-  if (element.labels !== undefined) return element.labels
-
   if (!isLabelable(element)) return []
 
+  if (element.labels !== undefined) return element.labels
+
```
marcosvega91, line 40: Is better to do in this way instead of swap the two functions?

```suggestion
  if (element.labels !== undefined) return element.labels ?? []

```

marcosvega91 (changes requested):
Thanks! :) I have added a small request

src/label-helpers.js
thomasmarshall: I've added a `squash` commit with your suggestion – thanks @marcosvega91.

kentcdodds (approved):
Thank you very much! I apologize for not reading your original post more carefully. Had I done that, I wouldn't have needed you to take the time to make the screenshot 🤦‍♂️

I blame covid.

Thanks!

kentcdodds:
@all-contributors please add @thomasmarshall for code and tests

thomasmarshall:
No worries, thanks!

PR #809, opened 2020-11-04
chore: switch to github actions

kentcdodds:
Because Travis gave OSS the shaft... https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

This works: https://github.com/kentcdodds/generator-kcd-oss/releases/tag/v4.9.0

We've got the NPM_TOKEN set at the org-level so we're solid 👍 

We won't actually *know* if the release works until we have a release, but 🤷‍♂️


.github/workflows/validate.yml
```diff
@@ -0,0 +1,68 @@
+name: validate
+on: [push, pull_request]
+jobs:
+  main:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
```
eps1lon, line 7: Do we need windows here? May be worth trying but might drain free minutes since it doubles the number of runs.

kentcdodds: Unless I'm reading this incorrectly, there is no limit to how much time the jobs run in total, only in how long an individual job runs: https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration

There's also a limit on concurrency, but that's probably not a concern either I think.

.github/workflows/validate.yml
```diff
@@ -0,0 +1,68 @@
...
+        run: npm run validate
+
+      - name: ⬆️ Upload coverage report
+        uses: codecov/codecov-action@v1
+
+  release:
+    needs: main
+    runs-on: ubuntu-latest
+    if:
+      ${{ contains('master,beta,next,alpha', github.base_ref) &&
```
kentcdodds, line 34: This doesn't appear to work. This job should not run on this PR but it is... I'm not sure how else to skip this job for non-release branches 🤔

marcosvega91: It's not enough to do ?

```md
on:
    push:
        branches:
            - master
```

wyze: Not 100%, but the docs say `github.base_ref` is only present on `pull_request` events. So it might be treating it like an empty string and it is matching the `contains` function. Just a possible guess to try something like `${{ github.base_ref != '' && contains(...` or `${{ contains(..., github.base_ref || 'not-found')`.

kentcdodds: Oh yeah! I think I need to use `ref` 👍 Thanks!

kentcdodds:
Alright! I think we're ready to make this happen!

Now I just need to apply this same change to LITERALLY EVERY REPO I MAINTAIN 😱

Sure wish it were easier to share these action configs 🙄

MatanBobi:
> Alright! I think we're ready to make this happen!
> 
> Now I just need to apply this same change to LITERALLY EVERY REPO I MAINTAIN 😱
> 
> Sure wish it were easier to share these action configs 🙄

That's odd that unlike gitlab-ci, there's no extend/include flow in github actions (at least I didn't find this in the documentation).

Need some help with applying the change in other repos?

afontcu:
> Alright! I think we're ready to make this happen!
> 
> Now I just need to apply this same change to LITERALLY EVERY REPO I MAINTAIN 😱
> 
> Sure wish it were easier to share these action configs 🙄

Just found this https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/sharing-workflows-with-your-organization, not sure if it helps?

.github/workflows/validate.yml
eps1lon: > There's also a limit on concurrency, but that's probably not a concern either I think.

There are only 20 per account. dom-testing-library alone now uses 8 instead of 4. I'm missing some context why we need to introduce the windows run.

kentcdodds: I guess I figured it was easy and "free," but I guess we can drop windows 👍

kentcdodds:
@MatanBobi, thanks for the offer! Feel free to make PRs. I'm thinking about making a custom action that does everything so it's easier to keep up-to-date.

@afontcu, thanks. Unfortunately that's just a way to make a template which is only marginally helpful. I want to be able to keep things up-to-date without having to update every repo.

MatanBobi:
@kentcdodds would you like me to wait for the custom action or can I start adding this workflow in our other repos? :)

kentcdodds:
I don't think custom actions will work very well, so you can go ahead. One thing I haven't figured out yet is how to properly ignore all-acontributers branches for pull requests. Other than that, generator-kcd-oss's validate job is pretty good 👍

PR #813, opened 2020-11-05
chore: Update MAINTAINING.md

MichaelDeBoey:


kentcdodds (approved):
👍

nickmccurdy:
How is doctoc being used here? If it's part of kcd-scripts I feel like the upstream MAINTAINING.md should be updated. If not, I'm curious if this should be added to either kcd-scripts or other testing-library repositories manually.

kentcdodds:
Yes, this should be applied to any repo that's using actions.

nickmccurdy:
I'm working on updating `generator-kcd-oss` from `dom-testing-library`, which will include this and other fixes

PR #818, opened 2020-11-10
fix: export pretty format

marcosvega91:
**What**: referring to this issue https://github.com/testing-library/react-testing-library/issues/694 I have exported `pretty-format` from this library

**Why**: Because we can then import in `@testing-library/react`

**How**:

**Checklist**:

- [ ] Tests
- [X] Typescript definitions updated
- [X] Ready to be merged

### DIFF
--- a/src/pretty-dom.js
+++ b/src/pretty-dom.js
@@ -71,4 +71,4 @@ const logDOM = (...args) => {
   }
 }
 
-export {prettyDOM, logDOM}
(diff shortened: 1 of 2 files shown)


nickmccurdy (approved):

kentcdodds (approved):
I think this is a reasonable solution to the problem (to make sure everyone's using the correct version of prettyFormat and it's typings). Thank you!

PR #821, opened 2020-11-13
Feature: adds aria-expanded filter to *ByRole

goblindegook:
**What**: Adds an `expanded` filter to `*ByRole` queries.

**Why**: Allow supporting roles to be filtered by the `aria-expanded` widget attribute.

**How**: Implemented in a similar pattern as other attribute filters.

**Checklist**:

- [x] Tests
- [x] Ready to be merged


nickmccurdy (approved):
The code looks good, though I wonder if this has an impact on `*ByRole`'s performance

goblindegook:
I imagine the impact on performance comes before --- the change affects filtering logic that comes after that.

I am wondering though, given that in this instance we're filtering by an attribute, wouldn't it be possible to replace that `container.querySelectorAll('*')` at the start with a `container.querySelectorAll('[aria-foo="bar"]')`?

eps1lon (approved):

eps1lon:
> I am wondering though, given that in this instance we're filtering by an attribute, wouldn't it be possible to replace that container.querySelectorAll('*') at the start with a container.querySelectorAll('[aria-foo="bar"]')

Create a benchmark, measure and share the results. I believe I checked this in the past and only observed negligible gains.

eps1lon:
"while dany kind of forgot"-meme format captioned with "While Sebastian kind of forgout about this PR"

PR #822, opened 2020-11-15
chore: Update validate.yml

MichaelDeBoey:


nickmccurdy:
@MichaelDeBoey Thanks for doing this, but I want to warn you I have another pull request planned for the generator soon, so you may want to wait until that's merged before updating anything else to save time

PR #823, opened 2020-11-16
fix: use node property instead of selector for type=text input

### LINKED ISSUE #814: Input with invalid type does not have textbox role
### Problem description:

According to spec:

input type=text or with a missing or invalid type, with no list attribute should have role=textbox
https://www.w3.org/TR/html-aria/

see repro https://codesandbox.io/s/react-testing-library-demo-forked-6mh5f

marcosvega91:
**What**: close #814 

**Why**: It was a bug

**How**: Adding an exception for type=text.
Instead of using the selector I have used the node property.
 

**Checklist**:

- [X] Tests
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -543,3 +543,8 @@ describe('configuration', () => {
     }
   })
 })
+
+test('should find the input using type property instead of attribute', () => {
+  const {getByRole} = render('<input type="124">')
+  expect(getByRole('textbox')).not.toBeNull()
+})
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
Looks good to me 👍 Anyone else want to weigh in?

eps1lon (approved):
Works for now. I'll revisit this in aria-query though or rather with a more holistic approach. Either aria-query always means "check the property not the attribute" or aria-query needs to be explicit when you check attributes vs. when you check properties.

kentcdodds:
@eps1lon should we just wait for aria-query instead? Or do you think that's further down the road?

eps1lon:
> @eps1lon should we just wait for aria-query instead? Or do you think that's further down the road?

Future improvement. This PR is good to go.

PR #827, opened 2020-11-17
fix(suggestions): only warn about inaccessible elements when actually showing the suggestion

benmonro:
**What**:
Console warnings are being emitted, even when they shouldn't be.  For example, if you're using getByRole, you shouldn't see a warning about it, only when the suggestion is actually used.

**Why**:

causes noise in test runs.

**How**:
by moving the console.warn into the toString() which is only used when the suggestion is thrown.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/suggestions.js
+++ b/src/__tests__/suggestions.js
@@ -551,16 +551,44 @@ test('should get the first label with aria-labelledby contains multiple ids', ()
   })
 })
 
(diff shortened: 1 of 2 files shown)


marcosvega91 (approved):
Hi @benmonro, looks good to me :)

PR #835, opened 2020-11-20
fix(suggestions): don't suggest byLabelText when using byLabelText

benmonro:
**What**: a bug fix on suggestions for findByLabelText

**Why**: when using findByLabelText, you would get a suggestion even if you're already using findByLabelText.

**How**:  use the right function when passing the name of the method.

**Checklist**:

- [x] Documentation added to the N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/suggestions.js
+++ b/src/__tests__/suggestions.js
@@ -232,6 +232,16 @@ test('should suggest getByLabelText when no role available', () => {
   )
 })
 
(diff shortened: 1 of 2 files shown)


benmonro:
@testing-library/dom 👋

nickmccurdy (approved):

Issue #582, opened 2020-05-19
Leverage levenshtein distance to help users debug easily when there are close matches


NeekSandhu:
### Describe the feature you'd like:

Sometimes, due to minor typo or an extra space here and there, a match is not successful and queries throw the error. Due to the typo being non-obvious (like double-space or a dash instead of em-dash etc), it isn't always an instantaneous pin-point to where the problem is.

I think it might be very useful for the users if they see something like this:

```bash
Unable to find an element with exact match: 15-25

Found close matches. Did you mean one of these:
  15–25
  19–22

<div>
	<ul> ... </ul>
</div>
```

### Suggested implementation:

When a candidate node is tested for a match, also calculate levenshtein distance from its `textContent` to the target string, and push the score to an array (along with the node itself). After all nodes are tested, and no match is found, sort the array on distance score and suggest top 3 close candidates to the user.

kentcdodds:
I've considered this and think it would be really cool. However, I'm worried that this will be very challenging to do well. I don't have the bandwidth to work on it myself, but I would be willing to look at a pull request for it.

kentcdodds:
Note that this doesn't mean it would get merged. So keep that in mind. It would need to not complicate the codebase more than is reasonable.

NeekSandhu:
>  I'm worried that this will be very challenging to do well

what challenges do you see off top of your head?

kentcdodds:
Just feels like there are dragons lurking. I could be wrong though 😅

dougbacelar:
I created a [proof of concept](https://github.com/testing-library/dom-testing-library/pull/836) to gather some feedback. If there is interest i can finish it up 👍

PR #836, opened 2020-11-21
feat: suggest close matches using Levenshtein distance [POC]

dougbacelar:
**What**:
Inspired by https://github.com/testing-library/dom-testing-library/issues/582

This provides a way to suggest close matches to users when the query cannot find any elements.
```js
render(<div data-testid="cat" />)
screen.getByTestId('kat');

// output
`Unable to find an element by: [data-testid="kat"]. Did you mean one of the following?
cat`
```

This is a POC, looking to gather feedback to see if its worth it pursuing.

**Why**:
Might make it easier to debug, specially when there are typos in the queries or when a certain element name has changed slightly.

**How**:
_query by attribute:_
1. iterate through all elements and calculate close matches
2. keep only matches that are the closest to the search string, keep all that are the same distance

_calculate close matches:_
1. initialise a dynamic programming table of size MxN where `M = element text length` and `N = search string length`
2. use the dp table above to calculate the Levenshtein distance between the element text and the search string

Note: this was implemented only on the byTestId query for now and behind a `computeClosetMatches` flag

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/close-matches.js
+++ b/src/close-matches.js
@@ -1,41 +1,5 @@
 import {makeNormalizer} from './matches'
-
-const initializeDpTable = (rows, columns) => {
-  const dp = Array(rows + 1)
-    .fill()
-    .map(() => Array(columns + 1).fill())
-
-  // fill rows
-  for (let i = 0; i <= rows; i++) {
-    dp[i][0] = i
-  }
-
-  // fill columns
-  for (let i = 0; i <= columns; i++) {
-    dp[0][i] = i
-  }
-  return dp
-}
-
-export const calculateLevenshteinDistance = (text1, text2) => {
-  const dp = initializeDpTable(text1.length, text2.length)
-
-  for (let row = 1; row < dp.length; row++) {
-    for (let column = 1; column < dp[row].length; column++) {
-      if (text1[row - 1] === text2[column - 1]) {
-        dp[row][column] = dp[row - 1][column - 1]
-      } else {
-        dp[row][column] =
-          Math.min(
-            dp[row - 1][column - 1],
-            dp[row][column - 1],
-            dp[row - 1][column],
-          ) + 1
-      }
-    }
-  }
-  return dp[text1.length][text2.length]
-}
+import calculateLevenshteinDistance from 'leven'
 
 const MAX_LEVENSHTEIN_DISTANCE = 4
 
--- a/src/close-matches.js
+++ b/src/close-matches.js
@@ -1,0 +1,96 @@
+import {makeNormalizer} from './matches'
+
+const initializeDpTable = (rows, columns) => {
+  const dp = Array(rows + 1)
+    .fill()
+
+    .map(() => Array(columns + 1).fill())
+
+  // fill rows
+
+  for (let i = 0; i <= rows; i++) {
+    dp[i][0] = i
+  }
+
+  // fill columns
+
+  for (let i = 0; i <= columns; i++) {
+    dp[0][i] = i
+  }
+
+  return dp
+}
+
+export const calculateLevenshteinDistance = (text1, text2) => {
+  const dp = initializeDpTable(text1.length, text2.length)
+
+  for (let row = 1; row < dp.length; row++) {
+    for (let column = 1; column < dp[row].length; column++) {
+      if (text1[row - 1] === text2[column - 1]) {
+        dp[row][column] = dp[row - 1][column - 1]
+      } else {
+        dp[row][column] =
+          Math.min(
+            dp[row - 1][column - 1],
+
+            dp[row][column - 1],
+
+            dp[row - 1][column],
+          ) + 1
+      }
+    }
+  }
+  return dp[text1.length][text2.length]
+}
+
+const MAX_LEVENSHTEIN_DISTANCE = 4
+
+export const getCloseMatchesByAttribute = (
+  attribute,
+
+  container,
+
+  searchText,
+
+  {collapseWhitespace, trim, normalizer} = {},
+) => {
+  const matchNormalizer = makeNormalizer({collapseWhitespace, trim, normalizer})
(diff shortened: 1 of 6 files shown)


src/queries/test-id.js
```diff
@@ -11,8 +12,28 @@ function queryAllByTestId(...args) {
 
 const getMultipleError = (c, id) =>
   `Found multiple elements by: [${getTestIdAttribute()}="${id}"]`
-const getMissingError = (c, id) =>
-  `Unable to find an element by: [${getTestIdAttribute()}="${id}"]`
+const getMissingError = (
+  c,
+  id,
+  {computeCloseMatches = false, ...options} = {},
```
dougbacelar, line 18: could be a config instead?

src/close-matches.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  // fill columns
+  for (let i = 0; i <= columns; i++) {
+    dp[0][i] = i
+  }
+  return dp
+}
+
+export const calculateLevenshteinDistance = (text1, text2) => {
```
kentcdodds, line 20: Rather than maintain/test this ourselves, I'd much rather use a third party module. I've used https://github.com/sindresorhus/leven before and it works well.

src/queries/test-id.js
kentcdodds: Could be both. We have several options that default to global config which you should be able to find around the codebase. Check `waitFor`.

src/queries/test-id.js
```diff
@@ -11,8 +12,24 @@ function queryAllByTestId(...args) {
...
+) => {
+  const defaultMessage = `Unable to find an element by: [${getTestIdAttribute()}="${id}"]`
+
+  const closeMatches =
+    !computeCloseMatches || typeof id !== 'string'
+      ? []
+      : getCloseMatchesByAttribute(getTestIdAttribute(), c, id, options)
```
kentcdodds, line 25: I'm concerned about this increasing the performance issues we already have with `find*` queries which are expected to fail at least once. Any chance we could lazily calculate this value so it's only run when the error is actually displayed? I don't know whether this is possible.

But perhaps my concern is unwarranted? Maybe this is faster than I think?

kentcdodds (commented):
Thanks for this! I'm thinking this would be pretty useful. Good progress so far.

src/queries/test-id.js
dougbacelar-houzz: > I'm concerned about this increasing the performance issues we already have with find* queries which are expected to fail at least once.

Good point. I tested a few times on my local running `getByTestId('search', {computeCloseMatches: true})` vs `getByTestId('search', {computeCloseMatches: false})`. 

When false, it finished consistently at around 20ms. When true finished at around 35ms. It increases with the number of elements found, but not by much. Might become slightly quicker with the [recommended lib](https://github.com/sindresorhus/leven)

> Any chance we could lazily calculate this value so it's only run when the error is actually displayed? 

An alternative would be to throw functions for `find*` queries. And then check if `typeof lastError === 'function'`  when [waitFor](https://github.com/testing-library/dom-testing-library/blob/master/src/wait-for.js#L156) times out. 

That might mean decoupling `find*` and `get*` queries a bit or perhaps make `get` queries depend on `find` queries instead of the other way around. 

i.e: a `get*` query could be a `find*` query with `{timeout: 0, interval: 0}` ? (not sure if that would work)

Since that seems a bit involved, we could start with a config `computeCloseMatches` that defaults to `false` and give that some testing?

dougbacelar: > I'm concerned about this increasing the performance issues we already have with find* queries which are expected to fail at least once.

Good point. I tested a few times on my local running `getByTestId('search', {computeCloseMatches: true})` vs `getByTestId('search', {computeCloseMatches: false})`. 

Not a reliable benchmark, but when false it finished consistently at around 20ms. When true finished at around 35ms. It increases with the number of elements found, but not by much. Might become slightly quicker with the [recommended lib](https://github.com/sindresorhus/leven). 

> Any chance we could lazily calculate this value so it's only run when the error is actually displayed?

An alternative would be to throw functions for `find*` queries. And then check if `typeof lastError === 'function'`  when [waitFor](https://github.com/testing-library/dom-testing-library/blob/master/src/wait-for.js#L156) times out.

That might mean decoupling `find*` and `get*` queries a bit or perhaps make `get` queries depend on `find` queries instead of the other way around.

i.e: a `get*` query could be a `find*` query with `{timeout: 0, interval: 0}` ? (not sure if that would work)

Since that seems a bit involved, we could start with a config `computeCloseMatches` that defaults to `false` and give that some testing?

kentcdodds: I'm ok with giving it a trial run and seeing what real-world experience with it will be like, so defaulting to disabled makes sense to me. Let's try it out using `leven`.

dougbacelar: @kentcdodds do you think we should start with the testId query or add this feature to all 10 queries? Wondering if i can break down the work somehow...

Re. the performance concern. Maybe if we realise there is a huge performance impact we can skip the computation of close matches on `find*` queries(similar to what was done for the `role` query here: https://github.com/testing-library/dom-testing-library/pull/590

kentcdodds: I'm not sure. What does everyone else think?

MatanBobi: IMO if the default is false, we can give it a try in more queries, though, this seems to me like a configuration that shouldn't be set to true by default at all, especially since it has a performance impact. I see this as something that will not be helpful in CI for example and will only take more time so if the developer wants to opt in they will have an option to do that.
Putting aside what I wrote above, I really like this PR and do think it can have a valuable impact so thanks for this :)

nickmccurdy: Can you please try a benchmark again using the new `leven` implementation?

PR #834, opened 2020-11-20
fix(types): allow all elements

nickmccurdy:
Port of testing-library/react-testing-library#833

**Checklist**:

- [ ] Documentation added to the
      [docs site](https://github.com/testing-library/testing-library-docs)
- [ ] Tests
- [ ] Typescript definitions updated
- [ ] Ready to be merged

PR #848, opened 2020-12-10
feat: migrate some files to typescript 

marcosvega91:
follow-up #614 

I have started migrating @testing-library/dom to typescript.  The idea is also to complete step by step PR #834 

Types are not generated until everything is completed

Any suggestions and advice are always welcome 🙏🏽 

I want to add more files to this PR.

**What**: File migrated

- [x] config.js
- [x] matches.js
- [x] label-helpers.js

**Why**: Is cool and fun 

**How**:

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged


src/__tests__/config.js
```diff
@@ -1,4 +1,4 @@
-import {configure, getConfig} from '../config'
+import {configure, getConfig} from '../config.ts'
```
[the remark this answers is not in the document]
marcosvega91: I don't like too, but I didn't find a way to not specify it. `eslint` gave me some trouble.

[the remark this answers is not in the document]
kentcdodds: We can deal with that later. Not a ship stopper.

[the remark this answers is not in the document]
marcosvega91: I'll leave the point open, maybe someone could help

kentcdodds (approved):
We need to remember that the `types` directory should be completely removed when this migration is finished. There's no reason to keep it around once we can generate the types from the source.

But for now, this is perfect 👍

kentcdodds:
Personally, I think let's get this merged, then we can continue to iterate and avoid large branches.

marcosvega91:
I have only one more file

marcosvega91:
ready for now

src/label-helpers.ts
```diff
@@ -22,37 +25,46 @@ function getTextContent(node) {
     .join('')
 }
 
-function getLabelContent(node) {
+function getLabelContent(node: Node | Element | HTMLInputElement) {
   let textContent
-  if (node.tagName.toLowerCase() === 'label') {
+  if ('tagName' in node && node.tagName.toLowerCase() === 'label') {
```
kentcdodds, line 30:
```suggestion
  if (node.tagName?.toLowerCase() === 'label') {
```

Does this work?

src/label-helpers.ts
```diff
@@ -22,37 +25,46 @@ function getTextContent(node) {
...
     textContent = getTextContent(node)
+  } else if ('value' in node) {
+    textContent = node.value
```
kentcdodds, line 33:
```suggestion
```

src/label-helpers.ts
```diff
@@ -22,37 +25,46 @@ function getTextContent(node) {
...
   } else {
-    textContent = node.value || node.textContent
+    textContent = node.textContent
```
kentcdodds, line 35: Could this work?

```suggestion
    textContent = node?.value ?? node.textContent
```

Or even

```suggestion
    textContent = node.value ?? node.textContent
```

kentcdodds (dismissed):
Looking good to me. I don't plan to merge this myself, but I did leave a few thoughts and ideas that aren't a big deal either way.

src/config.ts
```diff
@@ -1,9 +1,15 @@
...
 
+interface InternalConfig {
+  _disableExpensiveErrorDiagnostics: boolean
+}
+
 // It would be cleaner for this to live inside './queries', but
 // other parts of the code assume that all exports from
 // './queries' are query functions.
-let config = {
+let config: Config & InternalConfig = {
```
eps1lon, line 12: Can you extend `Config` instead with `interface InternalConfig extends Config`? 

(Intersection) types are slightly slower than interfaces: https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections

src/label-helpers.ts
eps1lon: Shouldn't we make sure we don't pass a `Node` here?

src/label-helpers.ts
```diff
@@ -22,37 +25,46 @@ function getTextContent(node) {
...
 }
 
 // Based on https://github.com/eps1lon/dom-accessibility-api/pull/352
-function getRealLabels(element) {
-  if (element.labels !== undefined) return element.labels ?? []
+function getRealLabels(element: Element | HTMLInputElement) {
+  // for old browsers
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if ('labels' in element && element.labels !== undefined)
+    return element.labels ?? []
```
eps1lon, line 45: I would prefer we don't change any runtime behavior but rather override the typechecker with type-casting or `@ts-expect-error`. Right now I can't know from the review whether this changes runtime behavior, or was unsound, or just TS being overly strict.

src/label-helpers.ts
```diff
@@ -22,37 +25,46 @@ function getTextContent(node) {
...
-  const labelsId = element.getAttribute('aria-labelledby')
-    ? element.getAttribute('aria-labelledby').split(' ')
-    : []
+function getLabels(
+  container: Element,
+  element: Element,
+  {selector = '*'} = {},
+) {
+  const ariaLabelledBy = element.getAttribute('aria-labelledby')
+  const labelsId = isNotNull(ariaLabelledBy) ? ariaLabelledBy.split(' ') : []
```
eps1lon, line 67: That's a change in behavior which we probably don't cover. For `<div aria-labelledby="">` we now return `[{ content: '', formControl: null }]` because `"".split(' ') => [""]`

src/utils.ts
```diff
@@ -0,0 +1,3 @@
+export function isNotNull<T>(arg: T): arg is NonNullable<T> {
+  return arg !== null
+}
```
eps1lon, line 3: TypeScript should no longer need this. If it does then override the type-checker instead of creating a new file for a one-liner.

tests/jest.config.dom.js
```diff
@@ -3,6 +3,8 @@ const baseConfig = require('kcd-scripts/jest')
 
 module.exports = {
   ...baseConfig,
+  moduleFileExtensions: [...baseConfig.moduleFileExtensions, 'd.ts'],
```
eps1lon, line 6: Why does jest need `.d.ts` files? They shouldn't have any runtime behavior.

types/matches.d.ts
```diff
@@ -1,14 +1,22 @@
...
 
 // Get autocomplete for ARIARole union types, while still supporting another string
 // Ref: https://github.com/microsoft/TypeScript/issues/29729#issuecomment-505826972
 export type ByRoleMatcher = ARIARole | MatcherFunction | {}
 
 export type NormalizerFn = (text: string) => string
 
+export type NormalizerOptions = DefaultNormalizerOptions & {
+  normalizer?: NormalizerFn
+}
```
eps1lon, line 18: Same point as above regarding intersection types and interfaces.

types/utils.d.ts
```diff
@@ -0,0 +1,3 @@
+export type Nullish<T> = T | null | undefined
+
+export type Callback<T> = () => T
```
eps1lon, line 3: Can we inline them first and then see if we really need another utils files?

eps1lon (changes requested):

tests/jest.config.dom.js
marcosvega91: I don't why but without using .d.ts jest not recognizes those files says that the module can"t be found

src/label-helpers.ts
marcosvega91: The problem here is the labels can be null and not undefined. Btw I think that on old browser like ie labels is undefined and not null

eps1lon: Looks to me covered all these cases and TypeScript is not able to narrow. Let's keep the runtime and annotate the compile time properly.

tests/jest.config.dom.js
eps1lon: Sounds like it's resolving a .d.ts file which it shouldn't do. Could you be a bit more specific where this exactly failing?

src/label-helpers.ts
marcosvega91: The problem is that the linter is giving the problem, so `@ts-expect-error` will not work

tests/jest.config.dom.js
marcosvega91: I have removed both lines, Yesterday jest gave some problems  🤷🏼

src/label-helpers.ts
eps1lon: Ignore the linter.

tests/jest.config.dom.js
eps1lon: It's Friday so `jest` is like :shrug:

eps1lon (dismissed):
Proposed follow-up in https://github.com/marcosvega91/dom-testing-library/pull/2 

Should have the minimal set of runtime changes. Kept the switch from `match` to `test` which could be merged regardless of types.

eps1lon:
Forgot to update the PR title of the recently merged https://github.com/testing-library/dom-testing-library/pull/821. Changing this PR to a feature (can always do that) so that we can kick of a release.

marcosvega91:
thanks for the help @eps1lon :)

package.json
```diff
@@ -53,11 +53,14 @@
...
-    "kcd-scripts": "^7.5.1",
+    "kcd-scripts": "^7.5.3",
     "typescript": "^4.1.2"
   },
   "eslintConfig": {
-    "extends": "./node_modules/kcd-scripts/eslint.js",
+    "extends": [
+      "./node_modules/kcd-scripts/eslint.js",
+      "plugin:import/typescript"
+    ],
```
eps1lon, line 63: As recommended: https://github.com/benmosher/eslint-plugin-import#typescript

marcosvega91: Yes I see, the strange things is that is should already be in kcd

kentcdodds: It should be. That's an oversight

PR #850, opened 2020-12-17
feat: migrate suggestions/pretty-dom/helpers to ts

marcosvega91:
**What**:

Follow-up #848 and #614

I have migrated the following files:

- [X] suggestions.js
- [X] pretty-dom.js
- [X] helpers.js 

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged


marcosvega91:
The CI is in error ( I don't understand why it wasn't in the previous PR ). The `Matcher` type can be of type Regex, string or a function. In the test types a number is used, should we handle this case ?

src/__tests__/suggestions.ts
```diff
@@ -187,39 +188,41 @@ test('escapes regular expressions in suggestion', () => {
   )
 
   expect(
-    getSuggestedQuery(
-      container.querySelector('img'),
+    (getSuggestedQuery(
+      container.querySelector('img') as HTMLImageElement,
```
eps1lon, line 192: question: Doesn't dom.d.ts infer the element type? Or does this only work for document.querySelector?

src/__tests__/suggestions.ts
```diff
@@ -400,11 +414,11 @@ test('getSuggestedQuery returns rich data for tooling', () => {
     variant: 'get',
   })
 
-  expect(getSuggestedQuery(button).toString()).toEqual(
+  expect((getSuggestedQuery(button) as Suggestion).toString()).toEqual(
```
eps1lon, line 417: question: Is a non-null assertion sufficient?
```suggestion
  expect(getSuggestedQuery(button)!.toString()).toEqual(
```
To make sure we're not casting to something completely different.

src/helpers.ts
```diff
@@ -113,9 +118,9 @@ function checkContainerType(container) {
     )
   }
 
-  function getTypeName(object) {
+  function getTypeName<T>(object: T) {
```
eps1lon, line 121: nit:
```suggestion
  function getTypeName(object: unknown) {
```
Is equivalent. We don't use the generic parameter anyway.

src/label-helpers.ts
```diff
@@ -55,10 +60,10 @@ function isLabelable(element: Element) {
 }
 
 function getLabels(
-  container: Element,
+  container: Element | Document,
```
eps1lon, line 63: Why would we pass a document here? That doesn't make much sense IMO.

src/__tests__/suggestions.ts
```diff
@@ -238,7 +241,7 @@ it('should not suggest by label when using by label', async () => {
   )
 
   // if a suggestion is made, this call will throw, thus failing the test.
-  const password = await screen.findByLabelText(/bar/i)
+  const password = (await screen.findByLabelText(/bar/i)) as Element
```
eps1lon, line 244: question: Why do we need this type casting?

src/helpers.ts
```diff
@@ -4,7 +4,7 @@ const globalObj = typeof window === 'undefined' ? global : window
 const TEXT_NODE = 3
 
 // Currently this fn only supports jest timers, but it could support other test runners in the future.
-function runWithRealTimers(callback) {
+function runWithRealTimers<T>(callback: () => T) {
```
eps1lon, line 7:
```suggestion
function runWithRealTimers(callback: () => unknown) {
```

Same as above: If we don't use the generic parameter, we can remove it.

src/suggestions.ts
```diff
@@ -50,38 +77,48 @@ function makeSuggestion(queryName, element, content, {variant, name}) {
...
   if (element.matches(DEFAULT_IGNORE_TAGS)) {
     return undefined
   }
 
   //We prefer to suggest something else if the role is generic
-  const role =
-    element.getAttribute('role') ?? getImplicitAriaRoles(element)?.[0]
+  const role: string =
+    element.getAttribute('role') ??
+    (getImplicitAriaRoles(element) as string[])[0]
```
eps1lon, line 121: From what are you casting here?

src/suggestions.ts
```diff
@@ -101,35 +138,40 @@ export function getSuggestedQuery(element, variant = 'get', method) {
 
   const placeholderText = element.getAttribute('placeholder')
   if (canSuggest('PlaceholderText', method, placeholderText)) {
-    return makeSuggestion('PlaceholderText', element, placeholderText, {
-      variant,
-    })
+    return makeSuggestion(
+      'PlaceholderText',
+      element,
+      placeholderText as string,
```
eps1lon, line 144: I would rather let `makeSuggestion` handle `null`. Looks like we pass `null` a lot.  Otherwise this is unsound.

src/__tests__/suggestions.ts
marcosvega91: the problem is that querySelector will return null or the element

src/__tests__/suggestions.ts
marcosvega91: Yes it'll be ok :)

src/__tests__/suggestions.ts
marcosvega91: we don't have already converted findByLabelText so it will return an any type for now

src/helpers.ts
marcosvega91: Maybe I miss something here. With this approach, if I call `runWithRealTimers` then I should cast the return value of the callback 

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAJ3AVTAazHA7mAdRigAsAlAUwEMAbAFRgFsLkBnACglpoCMqJMALkTsAlIgC8APkThseMOIDeAKERrEEBKyibufAZSghkYAGq0QFSXpq9+mMSo0bkFY6dv3D7k+csUKgC+zipaYDoovqYWNFZo8vg2qGAYiYTE5NT0TCwcYpIyAOQkFHZwReJUrIg6yDBgAOahoJCwCCjgAOIUYCwwEESklLQMzGwAPHRSnPoOwgXSiHTKahrhkVx2BphGfrFWNlvejqLOiOoXbh5Ixzt7MQHBoRu61-sByd29-YOZIzlxvlxEsSmUaBUzkA

src/suggestions.ts
marcosvega91: as for screen `getImplicitAriaRoles` has not been converted so it will return an array of any. Maybe we can add a TODO for those two cases?

src/suggestions.ts
marcosvega91: The problem is we already check in the `canSuggest` function that the third parameter is not null but typescript not recognize it

src/label-helpers.ts
marcosvega91: I have it now because in `getSuggestedQuery` function document is passed as container

src/suggestions.ts
marcosvega91: I have migrated role-helpers so we will not have this cast anymore

src/__tests__/suggestions.ts
eps1lon: Then this is a job for the non-null assertion operator `!`.

src/__tests__/suggestions.ts
eps1lon: Could you do a 
```suggestion
  const password: Element = await screen.findByLabelText(/bar/i)
```
so that we're still safe once we type findByLabelText?

src/helpers.ts
eps1lon:
```suggestion
function runWithRealTimers<T>(callback: () => T): T {
```
to make it obvious we're using the parameter.

src/suggestions.ts
eps1lon: Would `placeholderText is string` help as the return value of `canSuggest`?

src/label-helpers.ts
eps1lon: Ah, I didn't notice that is just the container not the target element :+1:

src/__tests__/suggestions.ts
marcosvega91: it will still remain a warning, maybe we can leave as is

src/role-helpers.ts
```diff
@@ -19,7 +41,7 @@ function isSubtreeInaccessible(element) {
   }
 
   const window = element.ownerDocument.defaultView
-  if (window.getComputedStyle(element).display === 'none') {
+  if (window && window.getComputedStyle(element).display === 'none') {
```
mathiassoeholm, line 44:
```suggestion
  if (window?.getComputedStyle(element).display === 'none') {
```

src/role-helpers.ts
```diff
@@ -40,17 +62,17 @@ function isSubtreeInaccessible(element) {
...
  */
-function isInaccessible(element, options = {}) {
+function isInaccessible(element: Element, options: InaccessibleOptions = {}) {
   const {
     isSubtreeInaccessible: isSubtreeInaccessibleImpl = isSubtreeInaccessible,
   } = options
   const window = element.ownerDocument.defaultView
   // since visibility is inherited we can exit early
-  if (window.getComputedStyle(element).visibility === 'hidden') {
+  if (window && window.getComputedStyle(element).visibility === 'hidden') {
```
mathiassoeholm, line 71:
```suggestion
  if (window?.getComputedStyle(element).visibility === 'hidden') {
```

src/__tests__/suggestions.ts
marcosvega91: we have a rule `@typescript-eslint/no-non-null-assertion`, we can disable it for the project, what do you think ?

src/role-helpers.ts
marcosvega91: right, thanks :)

nickmccurdy:
@marcosvega91 If it's not too much work would you mind extracting the CI fix into a separate PR so we can merge that first?

marcosvega91:
Yes, sure

src/__tests__/suggestions.ts
eps1lon: > we have a rule @typescript-eslint/no-non-null-assertion, we can disable it for the project, what do you think ?

I'd disable it. We're already violating the rationale by type-casting anyway.

Issue #837, opened 2020-11-25
Adding playground link when a test fails


nofar9792:
### Describe the feature you'd like:

The playground link is really useful.
Each time `find`/`get` function fails, it prints a really long HTML.
Maybe you can add the playground link by default, each time 
1. `find`/`get` function fails
2. One of the `expect` fails. like `expect(element).toBeChecked();`

### Suggested implementation:

1. Adding the link to `getElementError`.
2. I would like to hear what do you think

kentcdodds:
I think that's a pretty good idea personally. Maybe it could be disabled globally?

AlonMiz:
That sounds like a great idea.

delca85:
It sounds good to me too, it could be very useful but I agree with @kentcdodds that should be disabled globally.

kentcdodds:
To be clear, I think it should be enabled by default. Just something you can disable globally.

kentcdodds:
Also, to be clear further, anyone can claim this. Feel free to volunteer to implement this :)

AmirTugi:
Ooo sounds neat.
I'd definitely wan't this in my pipeline.

Aren't the `expect`s standalone, and don't relate to any testing lib?
How can we put the the link in that case without binding `jest-dom` to `testing-library`?

lidoravitan:
@kentcdodds i'd like to help with it

kentcdodds:
Go for it 💯

PR #852, opened 2020-12-19
Pr/playground link in error message

mathiassoeholm:
**What**:

Adds a link to the Testing-Library playground, when `getBy*` queries fail, as requested in https://github.com/testing-library/dom-testing-library/issues/837.

**Why**:

It adds awareness of the playground, which might help in making a better query.

**How**:

Changes the default `getElementError` function.

**Checklist**:

- [x] Tests
- [ ] Ready to be merged
      

Still need to add documentation about the `printPlaygroundLink` documentation, that's why this is still a draft.


mathiassoeholm:
It appears that Prettier might have formatted the inline snapshots differently, I'm not sure why 🤔

mathiassoeholm:
Before I will call this _Ready to be merged_, I have a few questions:

1. Is it okay that I remove part of a test, that was causing the type check to fail? https://github.com/testing-library/dom-testing-library/pull/852/commits/43e7464c158a2f66bc12453d7ec201b1757b6831
2. Should `@types/lz-string` be a dependency or a dev-dependency? I added it as dev dependency, because it doesn't provide any types to the public API.
3. Why do I get a `@typescript-eslint/no-unsafe-call` lint warning, when I've added the types? 
Screenshot 2020-12-23 at 12 24 01
4. Does anyone have a better suggestion for a text than "Open this DOM in the Testing-Library Playground"

src/__tests__/get-by-errors.js
```diff
@@ -97,6 +97,33 @@ cases(
   },
 )
 
+test.each([['getByText'], ['getByLabelText']])(
+  '%s query will print the playground link when enabled in the config',
+  query => {
+    configure({printPlaygroundLink: true})
+    document.body.innerHTML = '<div>Hello</div>'
+    expect(() => screen[query]('TEST QUERY')).toThrow(/playground/i)
```
eps1lon, line 105: Could you include a more sensible partial match here? It's not clear that `playground` refers to the link. This would still pass if we mention the word `playground` but not include the link, no?

src/__tests__/element-queries.js
```diff
@@ -2,6 +2,7 @@ import {configure} from '../config'
 import {render, renderIntoDocument} from './helpers/test-utils'
 
 beforeEach(() => {
+  configure({printPlaygroundLink: false})
```
eps1lon, line 5: I would just disable it in a global setup due to the noise it generates. And then make sure that when we do test it, we disable it again after the tests.

src/playground-helper.ts
```diff
@@ -0,0 +1,31 @@
...
+  // testing-playground will reformat on load any ways.
+  return value.replace(/[ \t]*[\n][ \t]*/g, '\n')
+}
+
+function encode(value: string) {
+  return compressToEncodedURIComponent(unindent(value))
+}
+
+function getPlaygroundUrl(element: Element | null, logErrors = true) {
+  if (!element || !('innerHTML' in element)) {
```
eps1lon, line 14:
```suggestion
  if (element === null) {
```
Should be sufficient looking at the type definitions.

src/playground-helper.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (logErrors) {
+      console.log(`The element you're providing isn't a valid DOM element.`)
```
eps1lon, line 16: I'm a bit uncomfortable with logging i.e. side-effects in a `get*` methods. Especially when it's about logging errors. Don't we usually want to throw errors?

eps1lon (changes requested):
Sorry for taking so long.

mathiassoeholm:
> Sorry for taking so long.

No worries! This is hardly an urgent issue, so thanks for taking the time to review :-)

src/playground-helper.ts
mathiassoeholm: The original if-statement in `logTestingPlaygroundURL` in `screen.js` look like this:

```javascript
if (!element || !('innerHTML' in element))
```

So changing it will cause a behaviour change for that function. Are we okay with that?

src/__tests__/get-by-errors.js
mathiassoeholm: Good point! It seems like a great use-case for `toThrowErrorMatchingSnapshot`, so I've changed it to that instead :-)

src/playground-helper.ts
mathiassoeholm: Makes sense :-) I've changed it to throw errors. The error message is logged (same as before) in `logTestingPlaygroundURL`, but ignored in `getElementError`.

src/__tests__/element-queries.js
mathiassoeholm: Good point! I tried changing it, but it causes a very interesting issue. I move the call to `configure({printPlaygroundLink: false})` inside a `beforeEach` in `setup-env.js`. But some tests are now failing:

https://github.com/testing-library/dom-testing-library/pull/852/checks?check_run_id=1697913416#step:6:548

It's failing to use the mocked `getUserCodeFrame` in the `pretty-dom` test, so `getUserCodeFrame.mockImplementationOnce` has no effect.

My theory is that this happens because `config.ts` uses `pretty-dom` and because we're importing `configure` from `config.ts` inside `setup-env`, `pretty-dom` in turn get's imported befored any tests are run.

This means that the `pretty-dom` module is "instantiated" (not sure if that's the right word) and imports `getUserCodeFrame` before it gets mocked, hence it ends up using the actual implementation of that function.

I can confirm my suspicion, because if I remove the use and import of `pretty-dom` inside `config.ts`, the `pretty-dom` tests will pass.

mathiassoeholm: I actually managed to solve it by removing the import in `setup-env` and changing the `beforeEach` to:

```javascript
beforeEach(() => {
  require('../src/config').configure({printPlaygroundLink: false})
})
```

What do you think, should I proceed with that solution?

eps1lon: > What do you think, should I proceed with that solution?

Sounds good. Make sure to add a comment explaining it. A link to the above explanation is probably sufficient. That made (at least to me) sense.

src/playground-helper.ts
eps1lon: Considering we now have the type-safety I'm ok with it. Unless the types are inaccurate.

mathiassoeholm: Hmm.. 🤔
In `screen.d.ts` the type for the element parameter is `Element | HTMLDocument`. However `HTMLDocument` does not have an `innerHTML` property.

It's not really clear to me why `logTestingPlaygroundURL` accepts a `HTMLDocument` if it requires `innerHTML` at the same time.

Should I keep `Element | null` or is `Element | HTMLDocument | null` the correct type?
Perhaps the type definition in `screen.d.ts` needs to be updated.

mathiassoeholm: Another question: do we even want the `| null`? From the type definition it sounds like `null` is a valid value, when in fact it's an erroneous value to give as parameter.
If we do not have `| null` though, we will have to remove the first check or shut up ESLint with:
```
// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
if (!element) {
```

mathiassoeholm: ### Update:

- I have decided to remove the `| null`, so we're not "advertising" that `null` can be used as a valid value.
- I don't think  `HTMLDocument` is a valid type to use here, so to me it looks like that was an error in the old type definition.
- I have removed a test that tried using `{}` as an argument and expecting `The element you're providing isn't a valid DOM element`. That error will no longer be logged if you're passing `{}`, because I removed the `!('innerHTML' in element)` check, and rely on static type checking instead.

src/__tests__/get-by-errors.js
```diff
@@ -97,6 +97,33 @@ cases(
...
+    configure({printPlaygroundLink: true})
+    document.body.innerHTML = '<div>Hello</div>'
+    expect(() => screen[query]('TEST QUERY')).toThrowErrorMatchingSnapshot()
+  },
+)
+
+test.each([['getByText'], ['getByLabelText']])(
+  '%s query will NOT print the playground link when disabled in the config',
+  query => {
+    configure({printPlaygroundLink: false})
```
eps1lon, line 112: Since these are side-effects, we need to make sure they're cleaned up after the test. So something like
1. save current config in `beforeEach`
2. set saved config in `afterEach`

mathiassoeholm: Something like that is already in place in this file :-)

https://github.com/testing-library/dom-testing-library/blob/5bc93643f312d9ca4210b97681686c9aa8a902d7/src/__tests__/get-by-errors.js#L6-L9

mathiassoeholm: I guess changing it to an `afterEach` will make sure it's also cleaned up after the last test runs inside this file.

mathiassoeholm: I have changed it to `afterEach` :-)
https://github.com/testing-library/dom-testing-library/pull/852/commits/1c4f22b754e1b3d52539d762f7109ba55290bd7b

src/config.ts
```diff
@@ -31,14 +32,37 @@ let config: InternalConfig = {
 
   // called when getBy* queries fail. (message, container) => Error
   getElementError(message, container) {
+    const playgroundUrl = (() => {
+      try {
+        return (
+          message &&
```
fildon, line 38: Why does the value of 'message' matter here?

src/config.ts
```diff
@@ -31,14 +32,37 @@ let config: InternalConfig = {
...
+      }
+    })()
+
     const error = new Error(
-      [message, prettyDOM(container)].filter(Boolean).join('\n\n'),
+      [
+        message,
+        prettyDOM(container),
+        playgroundUrl &&
+          `Open this DOM in the Testing-Library Playground:\n${playgroundUrl}`,
```
eps1lon, line 52:
```suggestion
          `Open this markup in the Testing-Library Playground:\n${playgroundUrl}`,
```
The DOM is an interface i.e. there's only one DOM

eps1lon (changes requested):

src/config.ts
mathiassoeholm: It's because of the usage of `getElementError` in `makeSingleQuery`:

```javascript
      const elementStrings = els
        .map(element => getElementError(null, element).message)
        .join('\n\n')
```

You can see that it's running through multiple element strings and joining the output. If we don't have the `message &&` part it would print the playground URL multiple times, for example:

```
Open this markup in the Testing-Library Playground:
https://testing-playground.com/#markup=JIOwxgTgpgtlIBcAEA3AhgGwK5SA

<button>
   Decrement value
</button>

Open this markup in the Testing-Library Playground:
https://testing-playground.com/#markup=CIUwxgTiC2IHYBcAEA3AhgGwK4iA

<button>
   Reset value
</button>

Open this markup in the Testing-Library Playground:
https://testing-playground.com/#markup=EoUwziAuAEBuCGAbAriIA
```

and this seams like a bit too much :-)

src/playground-helper.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+function encode(value: string) {
+  return compressToEncodedURIComponent(unindent(value))
+}
+
+function getPlaygroundUrl(element: Element) {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!element) {
+    throw new Error(`The element you're providing isn't a valid DOM element.`)
```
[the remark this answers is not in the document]
mathiassoeholm: There's an existing test in `screen.js`, testing that you can call `logTestingPlaygroundURL` with `null`: https://github.com/testing-library/dom-testing-library/blob/c273ed518cf783591692be8c0c5e5a34feafd4bb/src/__tests__/screen.js#L44

That test should also be removed if we remove this check. Otherwise we should expand the type for both `getPlaygroundUrl` and `logTestingPlaygroundURL` to use `Element | null`.

I'm leaning towards removing the check and the test. Not sure what you prefer?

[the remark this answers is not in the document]
eps1lon: > I'm leaning towards removing the check and the test.

That would be a breaking change. Let's not change unrelated code in this PR and match the types.

src/__tests__/screen.js
```diff
@@ -46,12 +46,6 @@ test('logs messsage when element is not a valid HTML', () => {
   expect(console.log.mock.calls[0][0]).toMatchInlineSnapshot(
     `"The element you're providing isn't a valid DOM element."`,
   )
-  console.log.mockClear()
-  screen.logTestingPlaygroundURL({})
-  expect(console.log).toHaveBeenCalledTimes(1)
-  expect(console.log.mock.calls[0][0]).toMatchInlineSnapshot(
-    `"The element you're providing isn't a valid DOM element."`,
-  )
```
eps1lon, line 54: Why was this test removed?

types/screen.d.ts
```diff
@@ -16,7 +16,7 @@ export type Screen<Q extends Queries = typeof queries> = BoundFunctions<Q> & {
    * Convenience function for `Testing Playground` which logs URL that
    * can be opened in a browser
    */
-  logTestingPlaygroundURL: (element?: Element | HTMLDocument) => void
+  logTestingPlaygroundURL: (element?: Element) => void
```
eps1lon, line 19: This shouldn't change considering the scope of this PR.

src/playground-helper.ts
```diff
@@ -0,0 +1,26 @@
...
+  return value.replace(/[ \t]*[\n][ \t]*/g, '\n')
+}
+
+function encode(value: string) {
+  return compressToEncodedURIComponent(unindent(value))
+}
+
+function getPlaygroundUrl(element: Element) {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!element) {
```
eps1lon, line 15:
```suggestion
function getPlaygroundUrl(element: Element | null) {
  if (element === null) {
```

Let's match runtime and static types

eps1lon (changes requested):

src/__tests__/screen.js
mathiassoeholm: I explained that some time ago :-) https://github.com/testing-library/dom-testing-library/pull/852#discussion_r563312385

types/screen.d.ts
mathiassoeholm: It's been a while now, but this change was actually triggered by your suggestion of removing `if (!element || !('innerHTML' in element))` and relying on static type checking: https://github.com/testing-library/dom-testing-library/pull/852#discussion_r555794226

For the static types to be correct after that change, we would have to remove `HTMLDocument` as a valid type, I explained this here: https://github.com/testing-library/dom-testing-library/pull/852#discussion_r557629382.
If I change `getPlaygroundUrl` to also accept `HTMLDocument` it will trigger a compile error:

```
Property 'innerHTML' does not exist on type 'HTMLDocument'.
```

`HTMLDocument` can be kept as part of `logTestingPlaygroundURL`, but it would be invalid and might cause a runtime error when it forwards the argument to `getPlaygroundUrl`.

src/playground-helper.ts
mathiassoeholm: I touched upon this a little bit: https://github.com/testing-library/dom-testing-library/pull/852#discussion_r557631676
Do you still think we should match the types, given that `null` is an erroneous type/value to pass in?

src/playground-helper.ts
eps1lon: > That error will no longer be logged if you're passing {}, because I removed the !('innerHTML' in element) check, and rely on static type checking instead.

We generally don't relly on static type checking for users of our library. We may change that in the future but for now additional runtime errors should not be removed.

eps1lon: > That error will no longer be logged if you're passing {}, because I removed the !('innerHTML' in element) check, and rely on static type checking instead.

We generally don't relly on static type checking for users of our library. We may change that in the future but for now additional runtime errors should not be removed.

src/playground-helper.ts
eps1lon: See https://github.com/testing-library/dom-testing-library/pull/852#discussion_r646339965

internal methods: static type checking
public methods: static type checking + extraneous runtime type checking for users not using static type checking

src/playground-helper.ts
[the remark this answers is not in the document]
mathiassoeholm: The test has been re-added 👍

src/playground-helper.ts
mathiassoeholm: Thanks for the explanation, I think I get it now! I have added the runtime check again in `logTestingPlaygroundURL`.

src/playground-helper.ts
mathiassoeholm: Since the runtime checks has been moved to `logTestingPlaygroundURL` I was able to remove the `if (element === null)` from here :-)

mathiassoeholm:
@eps1lon I think I finally understood the difference between error handling in internal and public facing functions, so the PR should be ready again now :-)

eps1lon (approved):
> @eps1lon I think I finally understood the difference between error handling in internal and public facing functions, so the PR should be ready again now :-)

Mostly because I was side-lining this PR and didn't put much time into explaining what I think this PR should look like. Thank your for sticking with it :+1: 

I'll do some git hygene with the snapshot diffs, ensure compatibility with the alpha and then try it out in Material-UI to check if the playground link is not too noisy.

eps1lon (changes requested):
On second thought: Are we sure this doesn't affect `findBy*` perf? `getMissingError` for `Byrole` has a special branch when called in `findBy*`. We need to make sure performance of `findBy` isn't affected by this change.

golergka:
What's the status of this issue? I want something like this implemented and would love to help.

RLRabinowitz:
I'm interested in this as well

Issue #854, opened 2020-12-21
Normalizer is missing when querying by aria-label


johnjesse:
- `@testing-library/dom` version: latest

### Relevant code or config:

https://github.com/testing-library/dom-testing-library/blob/accb6cc60628cb6b5bd4d9be2ead41724995e5ae/src/queries/label-text.js#L94

The normalizer is missing as an option passed into the `queryByAllAttribute` function - and so it's ignored if you're looking for an element by the aria-label

I'm going to add a PR to fix this, but I wanted to raise the issue first for some visibility

PR #855, opened 2020-12-21
fix: add normalizer for matching aria-label

johnjesse:
**What**:
fixed the label matcher - so that it passes on the normalizer when matching the aria-label, added a test too
see #854 

**Why**:
because the normalizer wasn't being correctly passed on

**How**:
see - src/queries/label-text.js

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

-  I did opt in to run git hooks - but had some issues with the `validate` script - it just seemed to error and not output anything useful so I ended up bypassing it.
- Also - all the tests in src/__tests__/text-matchers.js were named after a method. Because `...byLabelText` can get the label in multiple ways - and I wanted a specific test for the aria-label case - I've added a case that isn't named after a specific method

### DIFF
--- a/src/__tests__/text-matchers.js
+++ b/src/__tests__/text-matchers.js
@@ -231,6 +231,10 @@ cases(
         <input id="username" />`,
       queryFn: 'queryAllByLabelText',
     },
+    queryAllByAriaLabelText: {
+      dom: `<div aria-label="User ${LRM}name"/>`,
+      queryFn: 'queryAllByLabelText',
+    },
     queryAllByPlaceholderText: {
       dom: `<input placeholder="User ${LRM}name" />`,
       queryFn: 'queryAllByPlaceholderText',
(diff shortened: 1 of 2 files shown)


johnjesse:
The failing checks appear to be either 
```error TS2345: Argument of type 'number' is not assignable to parameter of type 'Matcher'.``` - which I don't think is down to my changes

OR

```@testing-library/dom@0.0.0-semantically-released validate: `kcd-scripts validate``` - again I don't know what I may have done to cause that

marcosvega91:
Hi @johnjesse, I have fixed this issue in #850. The problem is that the `typescript` configuration is wrong because is using types from `node_modules` instead of types from `types` folder.

johnjesse:
Happy new year :smile:, is there anyone available to review this pr?

eps1lon (approved):
Thanks :+1:

eps1lon:
@all-contributors add @johnjesse for bug and code.

PR #856, opened 2020-12-22
fix: add number support for matches

marcosvega91:
**What**:add number support for matches

**Why**: because of an issue with typescript configuration I have dropped the support in #848 

**How**:

**Checklist**:

- [ ] Tests
- [X] Ready to be merged

### DIFF
--- a/types/tsconfig.json
+++ b/types/tsconfig.json
@@ -1,3 +1,7 @@
 {
-  "extends": "../tsconfig.json"
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {"@testing-library/dom": ["."]}
+  }
 }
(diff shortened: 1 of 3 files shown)


types/tsconfig.json
```diff
@@ -1,3 +1,7 @@
 {
-  "extends": "../tsconfig.json"
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {"@testing-library/dom": ["."]}
+  }
```
nickmccurdy, line 6: These config changes seem unrelated, unless you're planning on pushing more files to this PR

nickmccurdy (changes requested):

types/tsconfig.json
marcosvega91: this config is useful because types tests are now using the types from `node_modules` instead of using types from their folder. I you made some changes to types the tests will fail without it because is used the alias `'@testing-library/dom'` to import types instead of relative path

nickmccurdy (approved):

PR #858, opened 2020-12-22
fix: use a data transfewr object supplied by the event caller

johnjesse:
**What**:

Pass the data transfer value through to the event if it's already an instance of `window.DataTransfer`

**Why**:

If you try and use a real `DataTransfer` object for a drag/drop event it is thrown away and it's properties aren't copied over to the one created within create event.

**How**:

I test the dataTransferValue to see if it is an `instanceof window.DataTransfer`

**Checklist**:

- [x] Tests
- [x] Ready to be merged


marcosvega91:
Hi @johnjesse could you please rebase this branch with master?

nickmccurdy:
@eps1lon Would we still want to support this considering the discussion #865?

eps1lon:
> @eps1lon Would we still want to support this considering the discussion #865?

I generally have a big problem with doing these polyfills in a library. So I'm the wrong person to ask since I was already against the original PR that started to polyfill these. This PR supports my position on not doing any polyfill in this library since it might cause bugs in proper environments.

nickmccurdy (approved):

nickmccurdy:
I think I agree with you on that, JSDOM workarounds can be pretty complex and sometimes cause more subtle bugs.

We can merge this if you're okay with it.

johnjesse:
Hey - just commenting to see if you want any changes on this PR or not, as it's been sat here for a while

nickmccurdy:
@eps1lon Do you think this is ready to merge?

johnjesse:
Is there any movement on this?

PR #868, opened 2021-01-03
fix: Pass container from findBy through to waitFor

nathanforce:
**What**:

Pass `container` from `findBy` into the `waitFor` call to support using `findBy` in non-global DOM environments.

**Why**:

The library should be agnostic of the runtime environment/globals when passing in a `container`.

**How**:

It looks like there are sensible defaults for `container` throughout the call stack, so I've just ferried the argument through.

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs) n/a
- [x] Tests
- [ ] Typescript definitions updated n/a
- [x] Ready to be merged


src/wait-for.js
```diff
@@ -89,6 +90,12 @@ function waitFor(
         await new Promise(r => setImmediate(r))
       }
     } else {
+      try {
```
nathanforce, line 93: This change was needed to keep these warnings intact. Without it, `getWindowFromNode(container)` throws and the error messages do not match the expected error. I thought it best to keep the warnings/behavior the same rather than update the test snapshots to match the new error message.

 https://github.com/testing-library/dom-testing-library/blob/master/src/__tests__/base-queries-warn-on-invalid-container.js#L95-L127

nathanforce:
Worth noting that I had to commit with `--no-verify` for this as it looks like #848 may have introduced a type error? 

```
[typecheck] types/__tests__/type-tests.ts(30,22): error TS2345: Argument of type 'number' is not assignable to parameter of type 'Matcher'.
```

marcosvega91:
Hi @nathanforce could you please rebase this branch with master?

eps1lon:
@all-contributors add @nathanforce for code

eps1lon:
@nathanforce Thanks!

PR #871, opened 2021-01-12
fix: prevent crashing on elements without parentElement

### LINKED ISSUE #870: waitForElementToBeRemoved: TypeError: Cannot read property 'parentElement' of null
- Fresh project generated by create-react-app@4:
- `@testing-library/dom` version: 5.11.4
  
  - jest@26.6.0
- DOM Environment:
  
  - jest-environment-jsdom@26.6.2
  - jsdom@16.4.0

### Relevant code or config:

Same code is available in codesandbox below. The real use case is a bit different but I was able to reproduce the issue with this example: 

```js
import React, { useReducer } from "react";
import Modal from "react-modal";
import userEvent from "@testing-library/user-event";
import {
    render,
    waitForElementToBeRemoved,
    screen,
} from "@testing-library/react";

beforeAll(() => {
    const root = document.createElement("div");
    root.id = "app-root";
    document.body.appendChild(root);
    Modal.setAppElement("#app-root");
});

function Component() {
    const [isOpen, toggle] = useReducer((s) => !s, false);

    return (
        <>
            <button onClick={toggle}>Open</button>

            <Modal isOpen={isOpen}>
                <button onClick={toggle}>Close</button>
                {isOpen && <div>Content</div>}
            </Modal>
        </>
    );
}

test("poc", async () => {
    render(<Component />);

    userEvent.click(screen.getByText("Open"));
    const content = await screen.findByText("Content");

    userEvent.click(screen.getByText("Close"));
    await waitForElementToBeRemoved(content);
});
```

### What you did:

Queried element rendered inside `react-modal`. Closed the modal and waited for the content to disappear. The modal seems to do some direct DOM manipulation, https://github.com/reactjs/react-modal/blob/master/src/components/Modal.js. 

### What happened:

```
    TypeError: Cannot read property 'parentElement' of null

      37 | 
      38 |     userEvent.click(screen.getByText("Close"));
    > 39 |     await waitForElementToBeRemoved(content);
         |           ^
      40 | });
      41 | 

      at node_modules/@testing-library/dom/dist/wait-for-element-to-be-removed.js:30:21
          at Array.map (<anonymous>)
      at waitForElementToBeRemoved (node_modules/@testing-library/dom/dist/wait-for-element-to-be-removed.js:27:43)
      at Object.<anonymous> (src/App.test.js:39:11)
```

https://github.com/testing-library/dom-testing-library/blob/cb7090238ff587d8778c1efae545643d4d076a02/src/wait-for-element-to-be-removed.js#L21-L25

### Reproduction:

https://codesandbox.io/s/amazing-brook-7xssv?file=/src/App.test.js

### Problem description:

`react-modal` is removing the parentElement of the queried element. The queried element is given to `waitForElementToBeRemoved` which crashes while accessing the parentElement. 

I'm sure there are work-arounds which can be used to avoid the crash caused by this minimal example. This is still a valid bug in the `waitForElementToBeRemoved` utility. 

### Suggested solution:

`waitForElementToBeRemoved` should not crash if given element has no `.parentElement` available. It should resolve successfully once missing parentElement is detected. 

Debug information:

```js
    const getRemainingElements = elements.map(element => {
      let parent = element.parentElement;

      debugger
      // element.outerHTML -> '<div>Content</div>'
      // element.parentElement -> null

      while (parent.parentElement) parent = parent.parentElement;
```

AriPerkkio:
**What**:
Fixes #870. 

**Why**:
Utility method `waitForElementToBeRemoved` should not crash when given element is already removed from its parent element. It should handle these cases and clearly report element is already removed. 

**How**:
Check whether element has `.parentElement` before accessing it. Prevents unexpected null pointer. 

**Checklist**:

- [ ] Documentation added to the N/A
- [x] Tests
- [x] Ready to be merged
      

I was unable to use the pre-commit hook. The `kcd-scripts test --findRelatedTests` keeps running forever. After 10-15mins I cancelled it and commited with `--no-verify`. Same thing has happened to me before with `@testing-library/eslint-plugin-jest-dom` repository.

Here's output of the test case **without the fix applied** just to indicate the error is same as in #870.
<details>
    <summary>Test case error</summary>

```
 FAIL   dom  src/__tests__/wait-for-element-to-be-removed.js
  ✕ requires element's parent to exist first (8 ms)

  ● requires element's parent to exist first

    expect(received).rejects.toThrowErrorMatchingInlineSnapshot(snapshot)

    Snapshot name: `requires element's parent to exist first 1`

    Snapshot: "The element(s) given to waitForElementToBeRemoved are already removed. waitForElementToBeRemoved requires that the element(s) exist(s) before waiting for removal."
    Received: "Cannot read property 'parentElement' of null"

      54 |   return expect(
      55 |     waitForElementToBeRemoved(div),
    > 56 |   ).rejects.toThrowErrorMatchingInlineSnapshot(
         |             ^
      57 |     `"The element(s) given to waitForElementToBeRemoved are already removed. waitForElementToBeRemoved requires that the element(s) exist(s) before waiting for removal."`,
      58 |   )
      59 | })

      at Object.toThrowErrorMatchingInlineSnapshot (node_modules/expect/build/index.js:241:20)
      at Object.<anonymous> (src/__tests__/wait-for-element-to-be-removed.js:56:13)
```
</details>

### DIFF
--- a/src/wait-for-element-to-be-removed.js
+++ b/src/wait-for-element-to-be-removed.js
@@ -20,7 +20,7 @@ async function waitForElementToBeRemoved(callback, options) {
     const elements = Array.isArray(callback) ? callback : [callback]
     const getRemainingElements = elements.map(element => {
       let parent = element.parentElement
-      if(!parent) return () => null
+      if(parent === null) return () => null
       while (parent.parentElement) parent = parent.parentElement
       return () => (parent.contains(element) ? element : null)
     })
--- a/src/wait-for-element-to-be-removed.js
+++ b/src/wait-for-element-to-be-removed.js
@@ -20,6 +20,7 @@ async function waitForElementToBeRemoved(callback, options) {
     const elements = Array.isArray(callback) ? callback : [callback]
     const getRemainingElements = elements.map(element => {
       let parent = element.parentElement
+      if(!parent) return () => null
       while (parent.parentElement) parent = parent.parentElement
       return () => (parent.contains(element) ? element : null)
     })
--- a/src/__tests__/wait-for-element-to-be-removed.js
+++ b/src/__tests__/wait-for-element-to-be-removed.js
@@ -42,6 +42,20 @@ test('requires an element to exist first', () => {
   )
 })
 
+test("requires element's parent to exist first", () => {
+  const {getByTestId} = renderIntoDocument(`
+  <div data-testid="div">asd</div>
+`)
+  const div = getByTestId('div')
+  div.parentElement.removeChild(div)
+
+  return expect(
+    waitForElementToBeRemoved(div),
+  ).rejects.toThrowErrorMatchingInlineSnapshot(
+    `"The element(s) given to waitForElementToBeRemoved are already removed. waitForElementToBeRemoved requires that the element(s) exist(s) before waiting for removal."`,
+  )
+})
+
 test('requires an unempty array of elements to exist first', () => {
   return expect(
     waitForElementToBeRemoved([]),


src/wait-for-element-to-be-removed.js
```diff
@@ -20,6 +20,7 @@ async function waitForElementToBeRemoved(callback, options) {
     const elements = Array.isArray(callback) ? callback : [callback]
     const getRemainingElements = elements.map(element => {
       let parent = element.parentElement
+      if(!parent) return () => null
```
eps1lon, line 23:
```suggestion
      if(parent === null) return () => null
```

eps1lon (approved):
Thanks!

Don't tell the TypeScript police we had this bug

eps1lon:
@all-contributors add @AriPerkkio for bug, code

PR #885, opened 2021-01-31
feat: add isInstanceOfElement helper

ph-fritsche:
Add a helper to verify the element type across libraries using `testing-library/dom` under the hood.

**What**:

Export a helper in `@testing-library/dist/helpers` to determine if an element is an instance of a specific `HTMLElement` type.

**Why**:

`@testing-library/user-event` [requires a better way to determine the element type](https://github.com/testing-library/user-event/pull/550).
As the requirement is not specific to `user-event`, I think it would be the cleaner solution to add it here.

**How**:

The helper tries to find the element constructor on the window.

As the window is neither required to be associated with the observed document nor is it required to provide any element constructors, the helper provides a fallback for other environments than Jest+JSDOM.

**Checklist**:

- N/A Documentation added to the
- [x] Tests
- N/A? Typescript definitions updated
There are no typescript definitions for internal methods at the moment.
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/helpers.js
+++ b/src/__tests__/helpers.js
@@ -3,7 +3,9 @@ import {
   getWindowFromNode,
   checkContainerType,
   runWithRealTimers,
+  isInstanceOfElement,
 } from '../helpers'
+import {render} from './helpers/test-utils'
 
 const globalObj = typeof window === 'undefined' ? global : window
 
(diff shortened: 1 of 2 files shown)


eps1lon:
> As the requirement is not specific to user-event, I think it would be the cleaner solution to add it here.

But only user-event needs this? Could you only implement this in user-event and then we'll see if we need it in other places as well and whether `@testing-library/dom` is the right place for this abstraction.

ph-fritsche:
See https://github.com/testing-library/user-event/pull/552

I've opened pull requests for both so we could implement it there and wait if this PR gets merged here.

As far as I know only `user-event` depends on identifying element types at the moment, but as the helper itself is strictly DOM related and has nothing to do with simulating events, collecting such abstractions here might prevent creating similar bugs in different places.

eps1lon:
Thanks for the suggestions. We're going to let this incubate in user-event and see if and in what form we need it in /dom. 

I suspect that most use-cases can live with a much simpler nodeType+localName check. That's what we're doing in dom-accessibility-api and we haven't heard any problems so far.

Issue #886, opened 2021-02-03
bug: runWithRealTimers activates fake Timers


ph-fritsche:
- `@testing-library/dom` version: master = v7.29.4

### Relevant code or config:

https://github.com/ph-fritsche/dom-testing-library/blob/24ebf282d10782b3d0a22b939a1b2950d4b5bd1b/src/__tests__/helpers.js#L86-L118

### What you did:

Tried to fix #884

### What happened:

The test leaks `setTimeout` only if `jest.useFakeTimers('modern')` has been used in some other test before.

### Reproduction:

See relevant code above and parent commits.

### Problem description:

Stopping the leak is not sufficient as the side-effect is in the tested unit.

The assumption that `jest.getRealSystemTime()` could be used to determine if a modern jest fake timer is active is wrong.
https://github.com/testing-library/dom-testing-library/blob/5bc93643f312d9ca4210b97681686c9aa8a902d7/src/helpers.js#L43-L45

### Suggested solution:

Determine if a jest fake timer is active by calling `jest.useRealTimers()` and observing the change of global timer functions.

Issue #884, opened 2021-01-31
test: Leaking test


ph-fritsche:
- `@testing-library/dom` version: master = v7.29.4

### Relevant code or config:

This test leaks `setTimeout` so that tests added below are not run by jest.
https://github.com/testing-library/dom-testing-library/blob/5bc93643f312d9ca4210b97681686c9aa8a902d7/src/__tests__/helpers.js#L78

### What you did:

Added tests for a pull request.

### What happened:

Jest fails to `setTimeout` for subsequent tests.
Adding a `describe` block with `test`, the tests are ignored without error messages.
Adding a `describe` block with `test.only` reveals the problem.

### Suggested solution:

Enclose the tests in a `describe` block and define `beforeAll`/`afterEach` to properly clean up after the test.

eps1lon:
Good catch. A PR would be appreciated!

PR #887, opened 2021-02-03
fix: remove side-effect from runWithRealTimers

ph-fritsche:
**What**:

Rewritten `runWithRealTimers` and the related tests.

**Why**:

`runWithRealTimers` activated fake timers when a `jest.useFakeTimers('modern')` has been used in any test before.
The tests leaked `setTimeout`.

See #884 and #886

**How**:

Call `jest.useRealTimers` and check if any timer functions changed to determine if fake timers have been active.

**Checklist**:
- N/A Documentation added to the
- [x] Tests
- N/A Typescript definitions updated
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/helpers.js
+++ b/src/__tests__/helpers.js
@@ -53,7 +53,9 @@ describe('run with real timers', () => {
   const realSetTimeout = global.setTimeout
 
   afterEach(() => {
+    // restore timers replaced by jest.useFakeTimers()
     jest.useRealTimers()
+    // restore setTimeout replaced by assignment
     global.setTimeout = realSetTimeout
   })
 
@@ -61,7 +63,7 @@ describe('run with real timers', () => {
     // legacy timers use mocks and do not rely on a clock instance
     jest.useFakeTimers('legacy')
     runWithRealTimers(() => {
-      expect(global.setTimeout).toEqual(realSetTimeout)
+      expect(global.setTimeout).toBe(realSetTimeout)
     })
     expect(global.setTimeout._isMockFunction).toBe(true)
     expect(global.setTimeout.clock).toBeUndefined()
@@ -71,7 +73,7 @@ describe('run with real timers', () => {
     // modern timers use a clock instance instead of a mock
     jest.useFakeTimers('modern')
     runWithRealTimers(() => {
-      expect(global.setTimeout).toEqual(realSetTimeout)
+      expect(global.setTimeout).toBe(realSetTimeout)
     })
     expect(global.setTimeout._isMockFunction).toBeUndefined()
     expect(global.setTimeout.clock).toBeDefined()
@@ -86,7 +88,8 @@ describe('run with real timers', () => {
     global.setTimeout = fakedSetTimeout
 
     runWithRealTimers(() => {
-      expect(global.setTimeout).toEqual(fakedSetTimeout)
+      expect(global.setTimeout).toBe(fakedSetTimeout)
     })
+    expect(global.setTimeout).toBe(fakedSetTimeout)
   })
 })
--- a/src/__tests__/helpers.js
+++ b/src/__tests__/helpers.js
@@ -60,27 +60,6 @@ describe('analyze test leak', () => {
     global.setTimeout = realSetTimeout
   })
 
-  const testReal = () => {
-    const originalSetTimeout = globalObj.setTimeout
-
-    // legacy timers use mocks and do not rely on a clock instance
-    jest.useFakeTimers('legacy')
-    runWithRealTimers(() => {
-      expect(originalSetTimeout).toEqual(globalObj.setTimeout)
-    })
-    expect(globalObj.setTimeout._isMockFunction).toBe(true)
-    expect(globalObj.setTimeout.clock).toBeUndefined()
-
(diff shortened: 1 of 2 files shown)


src/helpers.js
```diff
@@ -5,52 +5,42 @@ const TEXT_NODE = 3
...
+    clearTimeout,
+    setImmediate,
+    setInterval,
+    setTimeout,
+  }
+
+  // istanbul ignore else
+  if (jest) {
     jest.useRealTimers()
   }
```
nickmccurdy, line 24: Could we use this or does istanbul not support it?
```suggestion
  jest?.useRealTimers()
```

nickmccurdy (changes requested):

src/helpers.js
nickmccurdy: Ah yes, it doesn't seem to check that syntax

src/__tests__/helpers.js
```diff
@@ -53,42 +49,44 @@ describe('query container validation throws when validation fails', () => {
...
+describe('run with real timers', () => {
+  const realSetTimeout = global.setTimeout
 
-  // legacy timers use mocks and do not rely on a clock instance
-  jest.useFakeTimers('legacy')
-  runWithRealTimers(() => {
-    expect(originalSetTimeout).toEqual(globalObj.setTimeout)
+  afterEach(() => {
+    jest.useRealTimers()
+    global.setTimeout = realSetTimeout
```
eps1lon, line 57: Could you explain with a code comment why we need to restore this? Naively I would expect `jest.useRealTimers` to handle it.

eps1lon (changes requested):

src/helpers.js
eps1lon: Do we know that this function isn't entered in an environment where `jest` is not defined i.e. are we sure this doesn't throw a ReferenceError?

nickmccurdy: That's why I used `jest?.` If Jest is undefined, nothing will happen.

src/__tests__/helpers.js
ph-fritsche: I've added comments.

`jest.useRealTimers` restores the timers to the values they had before `jest.useFakeTimers` is called.
If `jest.useFakeTimers` has not been called, it does nothing.

See comments here for a more in depth explanation what happened (before the changes): https://github.com/testing-library/dom-testing-library/blob/24ebf282d10782b3d0a22b939a1b2950d4b5bd1b/src/__tests__/helpers.js#L63-L119

src/helpers.js
nickmccurdy: You're right, both these examples would throw a `ReferenceError`. Instead we should check that:
```js
  if (typeof jest !== undefined) {
    jest.useRealTimers()
  }
```

ph-fritsche: Good catch.
I've pushed this as a replacement:
```js
  try {
    jest.useRealTimers()
  } catch (e) {
    // not running in jest environment
  }
```

src/__tests__/helpers.js
eps1lon: > jest.useRealTimers restores the timers to the values they had before jest.useFakeTimers is called.

I would expect jest to restore this i.e. `jest.useFakeTimers().useRealTimers();` should have no effect. It sounds like you're saying that it does which may be a bug with jest.

src/helpers.js
```diff
@@ -5,52 +5,43 @@ const TEXT_NODE = 3
...
 
   const callbackReturnValue = callback()
 
-  if (fakeTimersType) {
-    jest.useFakeTimers(fakeTimersType)
-  }
+  const usedJestFakeTimers = Object.keys(timerAPI).filter(
+    k => timerAPI[k] !== globalObj[k],
+  ).length
```
nickmccurdy, line 31: Code style nitpick:
```suggestion
  const usedJestFakeTimers = Object.entries(timerAPI).some(
    ([name, func]) => func !== globalObj[name],
  )
```

src/__tests__/helpers.js
ph-fritsche: That has no effect. It restores the state before the fake was activated.
That's why we need to restore manually. One of our tests replaces `setTimeout` per assignment and not with `jest.useFakeTimers`.

If you replace a timer function manually between `useFakeTimers` and `useRealTimers`, it breaks `useRealTimers` as shown in that commit.
This might be considered a bug with jest, but could also be considered to be deliberately messing with the constraints jest imposes. ;)
Either way it is solved with the changes here.

src/helpers.js
nickmccurdy: I would still prefer the `typeof` check so we aren't swallowing errors from `useRealTimers`.

nickmccurdy (approved):

eps1lon:
Going to test this build on Material-UI first so that we have a integration test with non-jest setups.

eps1lon (approved):
LGTM: https://github.com/testing-library/dom-testing-library/pull/887#issuecomment-782146390

eps1lon:
@all-contributors add @ph-fritsche for code

@ph-fritsche Thanks!

PR #888, opened 2021-02-04
[feat] includes methods to allow finding elements by its id

ralves20:
**What**:
Inclusion of methods to allow finding elements by its id

**Why**:
Since, most of web applications commonly use ID instead of data-testid, would be good if there is a way to find elements by id when using testing-library, otherwise when the developer develop a component he would have to put the ID (which is commonly used) and the data-testid. So, a way to find an element by id would help a lot.

**How**:
Following the same structure that the project have

**Checklist**:

- [x] Tests
- [x] Ready to be merged


afontcu:
Hi,

Thank you for taking the time. However, this has been discussed extensively for a very long time ([see here](https://github.com/testing-library/react-testing-library/issues/183) as an example) and I doubt very much that this feature will ever made its way to the library.

nickmccurdy:
Thanks for contributing, but it's intentional that we support querying by test ID and not ID. Testing Library is designed to help test from the user's perspective, and test ID is a fallback for when there isn't a better way to query an element based on information a user would have. Plain IDs on the other hand are implementation details.

You're welcome to use this in your own projects if you need it, but I don't think it's in line with the philosophy of Testing Library, so I don't think we'll be merging it.

PR #900, opened 2021-02-19
fix: Don't assume mocked timers imply jest fake timers

### LINKED ISSUE #898: Recently published 7.29.5 has ReferenceError: jest is not defined when mocha is used
- `@testing-library/dom` version: 7.29.5
- Testing Framework and version: mocha 8.3.0
  
- DOM Environment: jsdom
  

### Relevant code or config:

mocha unit test that was working before 7.29.5

### What you did:

updated @testing-library/dom to 7.29.5

### What happened:

unit test failed with `jest is not defined` error

### Problem description:

```
ReferenceError: jest is not defined
    at _runWithRealTimers (node_modules/.f/_/@testing-library/dom/7.29.5/@testing-library/dom/dist/helpers.js:42:5)
```

### Suggested solution:

There's another use of `jest` a few lines above and it checks for it being `undefined` first, maybe need to do the same thing at the problem line.

eps1lon:
**What**:

Closes #898

**Why**:

`callback` might mock timers or timers might be mocked in another way. 

**How**:

Check if we're in a `jest`-like environment via `typeof jest !== 'undefined'`.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -25,9 +25,9 @@ function _runWithRealTimers(callback) {
 
   const callbackReturnValue = callback()
 
-  const usedJestFakeTimers = Object.entries(timerAPI).some(
-    ([name, func]) => func !== globalObj[name],
-  )
+  const usedJestFakeTimers =
+    typeof jest !== 'undefined' &&
+    Object.entries(timerAPI).some(([name, func]) => func !== globalObj[name])
 
   if (usedJestFakeTimers) {
     jest.useFakeTimers(timerAPI.setTimeout?.clock ? 'modern' : 'legacy')
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -41,8 +41,13 @@ function runWithJestRealTimers(callback) {
   }
 }
 
-const jestFakeTimersAreEnabled = () =>
-  typeof jest !== 'undefined' && runWithJestRealTimers(() => {}).usedFakeTimers
+function jestFakeTimersAreEnabled() {
+  // istanbul ignore else
+  if (typeof jest !== 'undefined') {
+    return runWithJestRealTimers(() => {}).usedFakeTimers
+  }
+  return false
+}
 
 // we only run our tests in node, and setImmediate is supported in node.
 // istanbul ignore next
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -5,10 +5,15 @@ const TEXT_NODE = 3
 
 // Currently this fn only supports jest timers, but it could support other test runners in the future.
 function runWithRealTimers(callback) {
-  return _runWithRealTimers(callback).callbackReturnValue
+  // istanbul ignore else
+  if (typeof jest !== 'undefined') {
+    return runWithJestRealTimers(callback).callbackReturnValue
+  }
+
+  return callback()
 }
 
-function _runWithRealTimers(callback) {
+function runWithJestRealTimers(callback) {
   const timerAPI = {
     clearImmediate,
     clearInterval,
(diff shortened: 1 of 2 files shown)


src/helpers.js
```diff
@@ -25,9 +25,9 @@ function _runWithRealTimers(callback) {
 
   const callbackReturnValue = callback()
 
-  const usedJestFakeTimers = Object.entries(timerAPI).some(
-    ([name, func]) => func !== globalObj[name],
-  )
+  const usedJestFakeTimers =
+    typeof jest !== 'undefined' &&
```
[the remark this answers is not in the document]
eps1lon: At the cost of having multiple return points though. Not sure this would actually improve things.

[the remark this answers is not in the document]
eps1lon: Might have a better idea.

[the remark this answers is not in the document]
eps1lon: I like the early return but not in this function. I moved the check higher up the tree so that we now have dedicated functions for jest timers and framework agnostic timer functions. This also gets rid of the weird `runWithRealTimers` vs `_runWithRealTimers` where `_runWithRealTimers` is now `runWithJestRealTimers`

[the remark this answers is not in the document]
eps1lon: @timdeschryver What do you think about [`a29bd91` (#900)](https://github.com/testing-library/dom-testing-library/pull/900/commits/a29bd919b3815912f76a1cb95200d43c8292f154)? The branching makes more sense to me now.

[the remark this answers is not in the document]
timdeschryver: I like it! 👍

timdeschryver (approved):

src/helpers.js
```diff
@@ -5,10 +5,16 @@ const TEXT_NODE = 3
 
 // Currently this fn only supports jest timers, but it could support other test runners in the future.
 function runWithRealTimers(callback) {
-  return _runWithRealTimers(callback).callbackReturnValue
+  // istanbul ignore else
+  if (typeof jest !== 'undefined') {
```
SimenB, line 9: note that https://jestjs.io/docs/en/configuration#injectglobals-boolean is a thing, so if people enable that this check will give a false positive.

Not sure what the best solution is...

SimenB: Possibly we could stick some non+enumerable symbol on the global object?

eps1lon: Thanks for the pointer!

I'm honestly at the point where we should just create some wrapper library that requires passing some adapter to fake/real timers. That logic is just way too hairy for my liking and we have to consider quite a number of environments by now that we don't actually test with.

Using the `exports` field might be the best approach to stick with zero-config but I don't know if we could/should add an extra `jest` or `mocha` or `browser` entry.

PS: [`react` has the same check](https://github.com/eps1lon/react/blob/09916479219a61ae86d2ec8ce159a161337b9007/packages/react-reconciler/src/ReactFiberHooks.old.js#L1419-L1422) so ideally we fix both instances so that we don't create even more pseudo-contexts (testing-library/react vs react in jest without globals).

SimenB: I came over this since I had to revert the `@testing-library/dom` update as it broke our tests using fake timers 😅 

Not sure if Jest should put somewhere "I'm here, running code and stuff" or if libraries should take config. I like the former as a consumer but it requires you to keep detection (albeit a "blessed" one), and creating adaptors for everything doesn't sound scalable. And makes it potentially harder for other libraries to integrate cleanly... Dunno what's best

eps1lon: > I came over this since I had to revert the @testing-library/dom update as it broke our tests using fake timers sweat_smile

Which version did you have to rollback?

kentcdodds: Is there any way jest exposes for us to detect whether we're running in a jest environment? I think that this feature is very useful and considering the vast majority of users are using jest it seems appropriate to include even as a "blessed" solution. I think it would make sense to support other popular runners as well assuming it doesn't complicate the codebase too much.

Is there an environment variable or something we can use to reliably detect whether we're running in a jest environment?

SimenB: > > I came over this since I had to revert the @testing-library/dom update as it broke our tests using fake timers sweat_smile
> 
> Which version did you have to rollback?

We were (and are back to) using `7.29.4`. Got `7.29.6` when updating, did not try `.5`

> Is there an environment variable or something we can use to reliably detect whether we're running in a jest environment?

Not as of now, that's what I was suggesting with the `Symbol`. I guess we could put something on `process.env`, tho, that sounds like a good idea. That might make sense regardless even though it requires polyfilling `process.env` for browser usage to use it. Jest won't be the _runner_ in the browser though (at least not any time soon), so I guess that's not really a concern

SimenB: We have `JEST_WORKER_ID`, maybe you can use that?

https://jestjs.io/docs/environment-variables#jest_worker_id

pastak:
https://github.com/testing-library/dom-testing-library/pull/900#discussion_r586480544

I have faced (maybe) the same issue since updating `@testing-library/dom` . Error is below:

```
ReferenceError: You are trying to access a property or method of the Jest environment after it has been torn down.

      at runWithJestRealTimers (node_modules/@testing-library/dom/dist/helpers.js:46:8)
      at jestFakeTimersAreEnabled (node_modules/@testing-library/dom/dist/helpers.js:65:12)
      at node_modules/@testing-library/dom/dist/wait-for.js:55:52

/app/node_modules/jest-runtime/build/index.js:1751
      this._fakeTimersImplementation.useFakeTimers();
                                     ^

TypeError: Cannot read property 'useFakeTimers' of null
    at Object.useFakeTimers (/app/node_modules/jest-runtime/build/index.js:1751:38)
    at runWithJestRealTimers (/app/node_modules/@testing-library/dom/dist/helpers.js:53:10)
    at jestFakeTimersAreEnabled (/app/node_modules/@testing-library/dom/dist/helpers.js:65:12)
    at /app/node_modules/@testing-library/dom/dist/wait-for.js:55:52
    at runNextTicks (internal/process/task_queues.js:62:5)
    at processImmediate (internal/timers.js:434:9)
```

and I solve it by reverting to v7.29.5

Issue #901, opened 2021-02-22
Tests with setTimeout fail after update to v7.29.6


ma7moudat:
- `@testing-library/dom` version: 7.29.6
- Testing Framework and version: jest 26.6.3
- DOM Environment: 

### Relevant code or config:

```js
describe(() => {
    it('should show loading state while search is running', async () => {
        MockInstance(Service, () => ({
            search: (request) => NEVER,
        }))
        const fixture = MockRender(SearchComponent)
        const el = fixture.point.nativeElement
        fixture.point.componentInstance.filters$.next(requestFilterMock)
        await sleep(100) // <-- uses setTimeout
        fixture.detectChanges()

        // initial search hint not shown
        expect(queryByText(el, matchers.initialSearch)).toBeFalsy()

        // loading text
        expect(getByText(el, matchers.loading))

        // show 15 skeleton clusters while loading
        expect(ngMocks.findAll(WagonHireClusterComponent)).toHaveLength(15)
        expect(getAllByLabelText(el, /loading/).length).toBeGreaterThan(0)
    })
})
```

### What you did:

Nothing, test was running fine before updating from 7.29.4 to 7.29.6. Now it fails with the error message below.

### What happened:

```shell
: Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.
Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:
    at new Spec (xyz\node_modules\jest-jasmine2\build\jasmine\Spec.js:116:22)
    at new Spec (xyz\node_modules\jest-jasmine2\build\setup_jest_globals.js:78:9)
    at new JBPatchedSpec (C:\Program Files\JetBrains\WebStorm\plugins\JavaScriptLanguage\helpers\jest-intellij\lib\jest-intellij-jasmine-reporter.js:94:7)
    at specFactory (xyz\node_modules\jest-jasmine2\build\jasmine\Env.js:523:24)
    at Env.it (xyz\node_modules\jest-jasmine2\build\jasmine\Env.js:592:24)
    at Env.it (xyz\node_modules\jest-jasmine2\build\jasmineAsyncInstall.js:134:23)
    at it (xyz\node_modules\jest-jasmine2\build\jasmine\jasmineLight.js:100:21)
    at context.<computed> (xyz\node_modules\zone.js\bundles\zone-testing-bundle.umd.js:4288:39)
[...]

PR #906, opened 2021-02-28
feat: Improve error message when passing an Array while a single Element is expected

juanca:
**What**: New error message when passing an Array.

**Why**: This can be a quick typo when using `screen.getAll*` or `screen.queryAll*` queries. 

**How**: Type checking for `Array`.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -94,7 +94,7 @@ function getWindowFromNode(node) {
     )
   } else if (node instanceof Array) {
     throw new Error(
-      `It looks like you passed an Array object instead of a DOM node. Did you do something like \`fireEvent.click(screen.getAllBy...\` when you meant to use a \`getBy\` query \`fireEvent.click(screen.getBy...\`?`,
+      `It looks like you passed an Array instead of a DOM node. Did you do something like \`fireEvent.click(screen.getAllBy...\` when you meant to use a \`getBy\` query \`fireEvent.click(screen.getBy...\`?`,
     )
   } else {
     // The user passed something unusual to a calling function
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -92,6 +92,10 @@ function getWindowFromNode(node) {
     throw new Error(
       `It looks like you passed a Promise object instead of a DOM node. Did you do something like \`fireEvent.click(screen.findBy...\` when you meant to use a \`getBy\` query \`fireEvent.click(screen.getBy...\`, or await the findBy query \`fireEvent.click(await screen.findBy...\`?`,
     )
+  } else if (node instanceof Array) {
+    throw new Error(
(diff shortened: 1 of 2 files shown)


src/helpers.js
```diff
@@ -92,6 +92,10 @@ function getWindowFromNode(node) {
     throw new Error(
       `It looks like you passed a Promise object instead of a DOM node. Did you do something like \`fireEvent.click(screen.findBy...\` when you meant to use a \`getBy\` query \`fireEvent.click(screen.getBy...\`, or await the findBy query \`fireEvent.click(await screen.findBy...\`?`,
     )
+  } else if (node instanceof Array) {
```
eps1lon, line 95:
```suggestion
  } else if (Array.isArray(node)) {
```

See http://web.mit.edu/jwalden/www/isArray.html

src/helpers.js
```diff
@@ -92,6 +92,10 @@ function getWindowFromNode(node) {
...
+    throw new Error(
+      `It looks like you passed an Array object instead of a DOM node. Did you do something like \`fireEvent.click(screen.getAllBy...\` when you meant to use a \`getBy\` query \`fireEvent.click(screen.getBy...\`?`,
```
eps1lon, line 97:
```suggestion
      `It looks like you passed an Array instead of a DOM node. Did you do something like \`fireEvent.click(screen.getAllBy...\` when you meant to use a \`getBy\` query \`fireEvent.click(screen.getBy...\`?`,
```

While technically correct, I don't think the additional qualifier helps identifying the issue.

eps1lon (commented):
Looks good. Some minor suggestions.

juanca:
I'm not really sure what went wrong with the validate script. It passes locally. And I do not see a re-run button.

Reading a bit into it, it might be an async failure. I think I've seen a PR about some timer shenanigans, it might be related. 

Lemme know how I should proceed.

MatanBobi:
> I'm not really sure what went wrong with the validate script. It passes locally. And I do not see a re-run button.
> 
> Reading a bit into it, it might be an async failure. I think I've seen a PR about some timer shenanigans, it might be related.
> 
> Lemme know how I should proceed.

I've just re-ran the jobs.
Did you try to reproduce it locally with node 10.14.2? I'm trying to figure out if this only fails in CI for that version or not.

eps1lon (approved):
Nice work, thanks!

MatanBobi:
> Nice work, thanks!

@eps1lon did this pass in CI? It was breaking in Node 10..

eps1lon:
> @eps1lon I see that it passed in CI, did you just re-run it?

Yep, it's a flaky test. Seems like others have the same problem: https://github.com/testing-library/dom-testing-library/issues/901#issuecomment-792537845.

MatanBobi:
> > @eps1lon I see that it passed in CI, did you just re-run it?
> 
> Yep, it's a flaky test. Seems like others have the same problem: [#901 (comment)](https://github.com/testing-library/dom-testing-library/issues/901#issuecomment-792537845).

Cool, thanks for the update :)

Issue #902, opened 2021-02-24
Allow narrowing down the part of the DOM that is printed when queries fail


Floriferous:
### Describe the feature you'd like:

When a query fails, the default behavior appears to be to log the entire `<body />` in the error message. In our case, even during tests, the body contains a lot of `<script>` tags, which makes the output so long that we don't get to see the actual interesting part, where the error is.

It would be great if we could specify which part of the DOM we want to be logged, as I control the wrapping element, and know how to query it. So a kind of config that says `onErrorQuery: '.test-wrapper'` could query for that element and print it?

<details>
<summary>Here's an example</summary>

```sh

Unable to find an element with the text: All files. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.

<body>
  
  
  <script
    type="text/javascript"
  >
[...]

PR #907, opened 2021-03-01
feat: Add `filterNode` option to `prettyDOM`

eps1lon:
BREAKING CHANGE: `<script />`, `<style />` and comment nodes are now ignored by default in `prettyDOM` .If you whish to return to the old behavior, use a custom `filterNode` function. In this case `prettyDOM(element, { filterNode: () => true })`.

I would find this change useful. But I believe this is possible to do in userland so I'm preparing a PR that just implements this in userland. We can then decide whether we want to defer to userland, release it as a breaking change or make it backwards compatible.

**What**:

Ignore `<script />`, `<style />` and comment nodes in `prettyDOM`

**Why**:

test frameworks like `karma` and `mocha` add `<script />` and comment nodes that [add quite a bit of noise making debugging harder](https://app.circleci.com/pipelines/github/mui-org/material-ui/38685/workflows/83279e68-0b89-4683-8c07-748ecf9cfc64/jobs/228457/parallel-runs/0/steps/0-112).

Would also improve https://github.com/testing-library/dom-testing-library/issues/902

**How**:

Fork `DOMElement` plugin from `pretty-dom` to allow filtering single `Node` instances.

**Checklist**:

      Will do when we release the stable v8
- [x] Tests
- [x] Ready to be merged


MatanBobi:
I can see the value of this one but it is a breaking change and should be considered thoroughly. 
What about adding this one behind a configuration?

eps1lon:
> I can see the value of this one but it is a breaking change and should be considered thoroughly.
> What about adding this one behind a configuration?

I think I referred to all of this in the PR descript: "We can then decide whether we want to defer to userland, release it as a breaking change or make it backwards compatible."

nickmccurdy:
Perhaps we could hide this behind a config boolean, and set the default to true on a major release.

eps1lon:
I appreciate the review but let's focus on the high level problem: Do we want this as a feature or breaking change? What is the use case for the current behavior?

Then we can talk about the implementation. The implementation is problematic for a feature since it adds configuration but trivial to do.

MatanBobi:
> I appreciate the review but let's focus on the high level problem: Do we want this as a feature or breaking change? What is the use case for the current behavior?

I'm actually not so sure I'm in favor of adding this one. I think that isn't a part of our responsibility and may just cause bugs/issues in some cases. As you already wrote, this can be patched on the user-land so I'm not sure a solution on our end is really necessary.
That's just my opinion of course.

eps1lon:
> I think that isn't a part of our responsibility and may just cause bugs/issues in some cases.

I think neither of these statements are relevant. What is our responsibility is better error diagnostics. So the question is if this helps or hinders error diagnostic which we should talk about.

Causing bugs is inherent to any change. The question is how likely they are.

> As you already wrote, this can be patched on the user-land so I'm not sure a solution on our end is really necessary.

I **believe**. I haven't actually done it so let's not jump to conclusions and assume that the userland implementation is trivial.

kentcdodds:
We already have `DEFAULT_IGNORE_TAGS`: https://github.com/testing-library/dom-testing-library/blob/ffc8f26d48543de95c7fff4e5d961cc4beaa3872/src/config.ts#L44

That's used for the `ignore` option on queries. I wonder whether it'd be possible to ignore everything that the failing query would ignore by default. That would be more correct anyway because right now we're displaying stuff that actually couldn't be queried.

kentcdodds:
To be clear, I'm in favor of this change in light of that :)

eps1lon:
> We already have DEFAULT_IGNORE_TAGS:

That makes a lot of sense. I was a bit suprised that I couldn't find existing implementation. Will try to see if I can leverage `DEFAULT_IGNORE_TAGS` usage instead.

kentcdodds:
To be clear, I think we should try to use whatever the `ignore` configuration is for the query that's failed. In most cases this will be the default, but it's an important distinction.

kentcdodds:
In fact, it may be useful to also display the `ignore` configuration as well so folks know that not everything is being displayed.

src/shared.ts
```diff
@@ -0,0 +1 @@
+export const DEFAULT_IGNORE_TAGS = 'script, style'
```
eps1lon, line 1: Breaks up a circular dep between `config` and `prettyDOM`

src/DOMElementFilter.ts
```diff
@@ -0,0 +1,259 @@
+/**
+ * Source: https://github.com/facebook/jest/blob/master/packages/pretty-format/src/plugins/DOMElement.ts
```
eps1lon, line 2:
```suggestion
 * Source: https://github.com/facebook/jest/blob/e7bb6a1e26ffab90611b2593912df15b69315611/packages/pretty-format/src/plugins/DOMElement.ts
```
Permalink

src/DOMElementFilter.ts
```diff
@@ -0,0 +1,259 @@
+/**
+ * Source: https://github.com/facebook/jest/blob/e7bb6a1e26ffab90611b2593912df15b69315611/packages/pretty-format/src/plugins/DOMElement.ts
+ */
+import type {Config, NewPlugin, Printer, Refs} from 'pretty-format'
```
eps1lon, line 4: Requires https://github.com/facebook/jest/pull/11342

Might need to use module augmentation depending on the targetted release of https://github.com/facebook/jest/pull/11342

src/config.ts
```diff
@@ -31,8 +31,18 @@ let config: InternalConfig = {
 
   // called when getBy* queries fail. (message, container) => Error
   getElementError(message, container) {
+    const prettifiedDOM = prettyDOM(container)
     const error = new Error(
-      [message, prettyDOM(container)].filter(Boolean).join('\n\n'),
+      [
+        message,
+        prettifiedDOM.length > 0
+          ? `Ignored nodes: comments, <script />, <style />\n${prettyDOM(
```
kentcdodds, line 39: I'm not sure I'm a fan of this note.

kentcdodds (commented):
Thanks for this work!

Not excited about vendoring the DOMElementFilter file. I'm guessing they weren't receptive to a PR for this feature?

How much work would it be to take the `ignore` option into account for the errors that are thrown? For example, let's say I'm trying to find text within a style tag. I would set the `selector` and `ignore` values like so:

```tsx
screen.getByText('.css-selector', {selector: 'style', ignore: 'script'})
```

Then my error should only show the things that could have matched. And it should show the selector and ignore configuration options as well.

So would it be possible to automatically filter the error message print out in a way that matches the `selector` and `ignore` values? I just think that a mismatch in those two could lead to confusing results.

I think it would amount to passing the `selector` and `ignore` options to `prettyDOM` and when those are provided, a `filterNode` function is created which only let's nodes that match the selector and don't match the `ignore` through. That way the error message displayed shows *exactly* what the user could have queried for given the options.

eps1lon:
> Not excited about vendoring the DOMElementFilter file. 

Me neither :smile: 

>  I'm guessing they weren't receptive to a PR for this feature?

It was pretty cumbersome to integrate while not breaking all the other tests. They generally weren't receptive to filtering stuff out at all which makes sense considering `pretty-format` is used for generating snapshot matchers. And if you want to filter stuff out from snapshots you're better off implementing custom snapshot matchers. But that's not how we're using `pretty-format`. 

I think I'm generally fine with the "fork now, upstream later" approach here.

> I just think that a mismatch in those two could lead to confusing results.

This is a real concern I agree. But we might want to revisit if we want to consider `screen.getByText('.css-selector', {selector: 'style', ignore: 'script'})` valid in the first place. I can't imagine that this is a useful query. It looks like querying implementation details when you're better off using `querySelect(.css-selector)` and assert the computed styles of that element.
In other words: Is the effort worth the edge cases it covers or does this edge case virtually never happen i.e. we're aware of the defect and are willing to push a bug fix if somebody needs it.

kentcdodds:
> I think I'm generally fine with the "fork now, upstream later" approach here.

Agreed.

> valid in the first place.

Yeah, I was worried my contrived example would be distracting 😅

I think in general the biggest problem is the mismatch between what we see in the error and what we ignore/select. If you don't use `ignore` or `select` then it's fine because we can just go with the defaults, but once you use those options then it can become pretty unhelpful, and that's what I would like to avoid.

> Is the effort worth the edge cases it covers or does this edge case virtually never happen i.e. we're aware of the defect and are willing to push a bug fix if somebody needs it.

That's fair. I currently don't have a use case for `ignore` or `selector` personally so I would be satisfied with a TODO code comment just so folks know that we'd like to support this in the future (rather than trying to figure out why we didn't do it in the first place).

eps1lon:
> I currently don't have a use case for ignore or selector personally so I would be satisfied with a TODO code comment just so folks know that we'd like to support this in the future (rather than trying to figure out why we didn't do it in the first place).

Sounds good. I just go ahead and add a test so that it's really "just" about fixing the problem. Makes it easier to get started contributing.

types/pretty-dom.d.ts
```diff
@@ -1,13 +1,21 @@
 import * as prettyFormat from 'pretty-format'
 
+export interface PrettyFormatOptions extends prettyFormat.OptionsReceived {
+  /**
+   * Given a `Node` return `false` if you whish to ignore that node in the output.
```
timdeschryver, line 5:
```suggestion
   * Given a `Node` return `false` if you wish to ignore that node in the output.
```

timdeschryver (commented):
This would be a great addition to ATL, where I believe that Karma is the most popular. Just a small typo...

types/pretty-dom.d.ts
eps1lon: English language: which, wish :man_facepalming:

kentcdodds (approved):
Solid 👍 Thanks!

Issue #904, opened 2021-02-26
getByRole fails if filtering by accessible name with &nbsp; in it


jegtnes:
- `@testing-library/dom` version: 7.28.1 (via `react-testing-library` 11.2.5)
- Testing Framework and version: Jest 26.6.3
- DOM Environment: JSDom 16.2.1

### Relevant code or config:
DOM:
```html
<label for="testy">Fascinating&nbsp;input</label>
<input id="testy" name="ice-cream-flavour" />
```

Test code:
```js
expect(
  screen.getByRole("textbox", { name: "Fascinating input" })
).toBeInTheDocument();
```

### What you did:

Attempted to find an element by role with a non-breaking space in the accessible name, by searching for the accessible name without having to encode a non-breaking space in JavaScript where the space goes (`\xa0`). 

#### Fails on getByRole
If searching for the accessible name with a regular space in `getByRole`, the test will fail. 

```js
getByRole("textbox", { name: "Fascinating input" })
```

#### Succeds on getByLabelText
If searching for the name with a regular space in `getByLabelText`, the test will succeed.

```js
getByLabelText("Fascinating input"))
```

#### Succeeds on getByRole with `\xa0` instead of a regular space
If searching for the accessible name with a JS representation of `&nbsp`, `\xa0` in `getByRole`,  the test will succeed.
```js
getByRole("textbox", { name: "Fascinating\xa0input" })
```

#### Fails on getByLabelText with `\xa0` instead of a regular space
If searching for the name with a JS representation of `&nbsp`, `\xa0`, in `getByLabelText`  the test will fail.
```js
getByLabelText("Fascinating\xa0input")
```

### Reproduction:

[Codesandbox reproduction](https://codesandbox.io/s/react-testing-library-demo-forked-hedd4?file=/src/__tests__/App.js)

### Problem description:

This seems incongrugous with testing-library's philosophy of "testing what the user sees", and is additionally inconsistent with other methods. If there's a reason we must specifically encode a breaking space to test for an element with a breaking space in the accessible name, it might be a good idea to be clearer about this in the error messaging. I was converting an existing code base from Enzyme to react-testing-library, and seeing an error like this understandably confused me a bit:

[...]

PR #908, opened 2021-03-04
Fix getByRole using matcher and normalizer on accessible names

julienw:
**What**: While studying the source code I realized that we weren't using the normal matcher and normalizer when matching the accessible name in `getByRole`. I believe this is a mistake.

**Why**: This removes some functionality, and is surprising tot he user.

**How**: I only reused existing objects.
Note: I've added 2 commits, solving separately the 2 issues I've seen, with tests.

**Checklist**:

- [X] Tests
- [X] Ready to be merged
      

While looking at the code, I've seen that the role matching also uses the full matcher and the normalizer, for example:
https://github.com/testing-library/dom-testing-library/blob/f7b5c33c44632fba1579cb44f9f175be1ec46087/src/queries/role.js#L97-L100
I find it somewhat weird that we can normalize roles, or look for inexact roles. But I haven't changed that because this might cause breakage in your users, and didn't want to risk delaying this patch either.

Thanks!

### DIFF
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -1,5 +1,6 @@
 import {configure, getConfig} from '../config'
 import {getQueriesForElement} from '../get-queries-for-element'
+import {getDefaultNormalizer} from '../'
 import {render, renderIntoDocument} from './helpers/test-utils'
 
 test('by default logs accessible roles when it fails', () => {
@@ -279,6 +280,19 @@ test('accessible name obeys exact: false', () => {
   // subset using exact: false
   expect(getByRole('heading', {name: 'gn u', exact: false})).not.toBeNull()
 });
+
+test('accessible name obeys text normalizer', () => {
+  const {getByRole} = render(
+    `<h1>Sign \u2068up\u2069</h1>`,
+  )
+
+  // match using a normalizer
+  expect(getByRole('heading', {
+    name: /gn u/,
+    // Remove all isolation characters
+    normalizer: str => getDefaultNormalizer()(str).replace(/[\u2066-\u2069]+/g, ''),
+  })).not.toBeNull()
+})
 
 test('TextMatch serialization in error message', () => {
   const {getByRole} = render(`<h1>Sign <em>up</em></h1>`)
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -271,6 +271,15 @@ test('accessible name filter implements TextMatch', () => {
   ).not.toBeNull()
 })
 
+test('accessible name obeys exact: false', () => {
+  const {getByRole} = render(
+    `<h1>Sign <em>up</em></h1><h2>Details</h2><h2>Your Signature</h2>`,
+  )
+
+  // subset using exact: false
+  expect(getByRole('heading', {name: 'gn u', exact: false})).not.toBeNull()
+});
+
 test('TextMatch serialization in error message', () => {
   const {getByRole} = render(`<h1>Sign <em>up</em></h1>`)
 
--- a/src/queries/role.js
+++ b/src/queries/role.js
@@ -153,7 +153,7 @@ function queryAllByRole(
         }),
         element,
         name,
-        text => text,
+        matchNormalizer,
       )
     })
 }
--- a/src/queries/role.js
+++ b/src/queries/role.js
@@ -146,7 +146,7 @@ function queryAllByRole(
         return true
       }
 
-      return matches(
+      return matcher(
         computeAccessibleName(element, {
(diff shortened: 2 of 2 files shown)


eps1lon:
I'm not so sure about this one. It's unclear whether we're not removing characters that are considered by screen-readers. 

Could you share some real world examples that this new behavior affects?

julienw:
Hey @eps1lon !
I'm not sure what you refer to with "this one". I don't think that this patch changes anything with screen readers by default.

To be clear, there are 2 changes here. Both of them are related to the fact that as a user of the library I expect to have all functionality of TextMatch that we have elsewhere and that are described [in the documentation](https://testing-library.com/docs/queries/about/#textmatch). The 2 commits are:
1. The first commit uses "matcher" instead of "matches". This obeys the `exact` property.
2. The second commit uses the passed normalizer function or obeys `trim` or `collapseWhitespace`.

Nothing should change for users that don't pass a normalizer function, or the properties `exact` `trim` or `collapseWhitespace`.
Also:
* neither `trim` or `collapseWhitespace` are described in the documentation, but they're accepted by the existing code and described in types.
* currently all these properties were used for the role matching. So we might have issues with users using them to match roles but who don't want to use them to match the actual value. But I find this very strange TBH, I don't see why a user would want to normalize a role before matching it, or trim/collapse whitespace on those. I would think that we should always match roles exactly, or possibly with a regexp. So my expectation is that users didn't use this before and we should remove this behavior.

Some real-world uses:
* for normalizer: we use a localization library that inserts these isolation characters between variables. We end up with a string like I used in the test: `"some text \u2068some variable\u2069"`, and this isn't matched by a regexp like `/some text some variable/`.
  This is confusing because these characters are zero-width and we don't see them in the error report. Also the user doesn't see them directly: the user would see the _effect_ of them when `some variable` is a LTR text but `some text` is in a RTL context, but in tests we may want ("may") to ignore them.
* for the use of matcher: I mostly find it surprising that we don't obey `exact` for the accessible name. I also don't see the use of "exact" for the roles...

Hope these explanations help!

eps1lon:
> This is confusing because these characters are zero-width and we don't see them in the error report. Also the user doesn't see them directly: the user would see the effect of them when some variable is a LTR text but some text is in a RTL context, but in tests we may want ("may") to ignore them.

I'm not concerned with what users "see" but what a screen reader would announce. I'm not familiar with what exactly we "normalize" or why this is desired. But `name` specifically refers to the accessible name that screen readers process so it's not clear whether the "normalization" actually reflects what happens with assistive technology.

> Both of them are related to the fact that as a user of the library I expect to have all functionality of TextMatch that we have elsewhere and that are described in the documentation.

`TextMatch` refers to the type you can pass to `name`. Not any other options. 

The existing matchers don't have a strong relation to assistive technology. `ByRole` has, so we can't just allow everything that the other queries would allow.

> I don't think that this patch changes anything with screen readers by default.

If it is purely additive then we can release it as a feature. But the PR title makes it sound like we change existing behavior. Could you clarify whether this is purely additive, or what existing behavior changed (emphasis on what was the behavior before and what it is now).

julienw:
If the user calls the *ByRole queries without additional options (only the role and the name), then the behavior is indeed slightly changed with this patch: the default normalizer is used. The default normalizer trims (removes spaces at the start and end of the text) and collapse whitespaces (replaces sequences of spaces with one space). _But_ the accessible name is already trimmed and with whitespaces collapsed, see https://github.com/eps1lon/dom-accessibility-api/blob/2cc14bc1def2714f8ea9b7695eeaef1a9af9f870/sources/accessible-name-and-description.ts#L49-L51.
So I don't think the behavior would change here.

If the user calls the *ByRole queries with additional options, such as `exact: false`, `trim: false`, `collapseWhitespace: false`, or `normalizer: someFunction`, because for some reason they wanted them to match the `role`, then the behavior would be changed on the accessible name match because they would apply to it as well. Note that `trim` and `collapseWhitespace` are useless to match roles, because the code splits on space to get the role values anyway.

For example:
```js
getByRole("text", { name: "foo", exact: false }); // The user wants to match the "textbox" role
getByRole("link", { name: "foo", trim: false }); // The user may reuse some default option object for some reason?
getByRole("link", { name: "foo", normalizer: () => {}}); // Seriously, who would use a normalizer to normalize the role names??
```

I'd like to close with a last note. I believe we want to encourage our users to use getByRole more, because tests would be accessibility-friendly. To achieve this we need that the ease of using getByRole is similar than for other queries. I believe this patch goes to this direction.

Hope htis helps.

eps1lon:
See https://github.com/testing-library/dom-testing-library/issues/904#issuecomment-905472296

Issue #914, opened 2021-03-20
When running in Jest, code assumes `setImmediate` and `clearImmediate`


SimenB:
- `@testing-library/dom` version: 7.29.6
- Testing Framework and version: Jest@https://github.com/facebook/jest/pull/11222
  
- DOM Environment: JSDOM@16.4.0
  

### Relevant code or config:

```js
require('@testing-libray/react')
```

### What you did:

Ran the test without global `setImmediate` and `clearImmediate` defined

### What happened:

![image](https://user-images.githubusercontent.com/1404810/111865461-64295800-8967-11eb-9d7b-d2dff7b3e94e.png)

### Reproduction:

See https://github.com/facebook/jest/pull/11222. Essentially, delete `setImmediate` and `clearImmediate` from `global` in the JSDOM env Jest sets up.

### Problem description:

It throws runtime errors at `require`/`import`-time

### Suggested solution:

Do a `typeof setImmediate !== 'undefined'` before using it.

kentcdodds:
If I'm not mistaken, the only place we rely on a global `setImmediate` is here: https://github.com/testing-library/dom-testing-library/blob/300bfe26f253b8f045da24bf3c75a83d1cc0f49c/src/helpers.js#L18-L25

That could probably be changed to use the `setImmediatePolyfill` and we could probably create a `clearImmediate` pollyfill as well.

craigkovatch:
What is it used for? `setImmediate` is *only* supported by IE10/11, and not on any kind of standards plan. To me it seems an odd choice to polyfill something that won’t ever exist natively in a modern browser.

gaearon:
So it seems like the place that uses is just this helper which is meant to `runWithJestRealTimers`. But if Jest removes `setImmediate` in jsdom (which *is* the plan), then there's no need for `setImmediate` to be in that list.

Therefore, the fix is just to change this code to be resilient to `setImmediate` not existing. I don't think any polyfilling is needed there.

renatoalencar:
> What is it used for? `setImmediate` is _only_ supported by IE10/11, and not on any kind of standards plan. To me it seems an odd choice to polyfill something that won’t ever exist natively in a modern browser.

`setImmediate` is supported by Node.js and it looks like is used to wait for changes in the DOM. There's already a polyfill in the code but used in other part of the code. It looks like this issue comes from this function `runWithJestRealTimers`, which really doesn't make sense to patch with a polyfill.

gaearon:
[...]

PR #916, opened 2021-03-25
Safe check for setImmediate and clearImmediate

renatoalencar:
**What**:

Avoid crashing when `setImmediate` and `clearImmediate` is not defined, which is now the case for JSDOM.

**Why**:

See #914 

**How**:

Safe check suggested by @gaearon

**Checklist**:

- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/helpers.js
+++ b/src/__tests__/helpers.js
@@ -98,15 +98,22 @@ describe('run with real timers', () => {
     expect(global.setTimeout).toBe(fakedSetTimeout)
   })
 
-  test('safe check for setImmediate and clearImmediate', () => {
-    const setImeediate = global.setImmediate
+  describe('run with setImmediate and clearImmediate deleted', () => {
+    const setImmediate = global.setImmediate
     const clearImmediate = global.clearImmediate
-    delete global.setImmediate
-    delete global.clearImmediate
 
-    expect(() => runWithRealTimers(() => {})).not.toThrow()
+    beforeEach(() => {
+      delete global.setImmediate
+      delete global.clearImmediate
+    })
 
-    global.setImmediate = setImeediate
-    global.clearImmediate = clearImmediate
+    afterEach(() => {
+      global.setImmediate = setImmediate
+      global.clearImmediate = clearImmediate
+    })
+
+    test('safe check for setImmediate and clearImmediate', () => {
+      expect(() => runWithRealTimers(() => {})).not.toThrow()
+    })
   })
 })
--- a/src/__tests__/helpers.js
+++ b/src/__tests__/helpers.js
@@ -97,4 +97,16 @@ describe('run with real timers', () => {
     })
     expect(global.setTimeout).toBe(fakedSetTimeout)
   })
+
+  test('safe check for setImmediate and clearImmediate', () => {
+    const setImeediate = global.setImmediate
+    const clearImmediate = global.clearImmediate
+    delete global.setImmediate
+    delete global.clearImmediate
+
+    expect(() => runWithRealTimers(() => {})).not.toThrow()
+
+    global.setImmediate = setImeediate
+    global.clearImmediate = clearImmediate
+  })
 })
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -16,12 +16,17 @@ function runWithRealTimers(callback) {
 
 function runWithJestRealTimers(callback) {
   const timerAPI = {
-    clearImmediate,
     clearInterval,
     clearTimeout,
-    setImmediate,
     setInterval,
     setTimeout,
+  }
+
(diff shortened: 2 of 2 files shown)


gaearon:
It would be good to check if this is enough to fix https://github.com/facebook/jest/pull/11222.

renatoalencar:
> It would be good to check if this is enough to fix [facebook/jest#11222](https://github.com/facebook/jest/pull/11222).

That's great, I'm trying to find a reasonable way to add tests for this.

gaearon:
Looks like https://github.com/facebook/jest/pull/11222#issuecomment-807030072 works!

src/__tests__/helpers.js
```diff
@@ -97,4 +97,16 @@ describe('run with real timers', () => {
...
+    const setImeediate = global.setImmediate
+    const clearImmediate = global.clearImmediate
+    delete global.setImmediate
+    delete global.clearImmediate
+
+    expect(() => runWithRealTimers(() => {})).not.toThrow()
+
+    global.setImmediate = setImeediate
+    global.clearImmediate = clearImmediate
+  })
```
eps1lon, line 111: It would add a bit more ceremony but make sure that we restore deleted globals in case this test fails e.g by wrapping in `describe` and adding `beforeEach` and `afterEach`.

Otherwise a failure of this test might break other tests even if they're passing. This could result in debugging the wrong test.

src/__tests__/helpers.js
```diff
@@ -97,4 +97,16 @@ describe('run with real timers', () => {
     })
     expect(global.setTimeout).toBe(fakedSetTimeout)
   })
+
+  test('safe check for setImmediate and clearImmediate', () => {
+    const setImeediate = global.setImmediate
```
eps1lon, line 102: nit: 
```suggestion
    const setImediate = global.setImmediate
```

src/__tests__/helpers.js
```diff
@@ -97,4 +97,16 @@ describe('run with real timers', () => {
...
+    const clearImmediate = global.clearImmediate
+    delete global.setImmediate
+    delete global.clearImmediate
+
+    expect(() => runWithRealTimers(() => {})).not.toThrow()
+
+    global.setImmediate = setImeediate
```
eps1lon, line 109: nit:
```suggestion
    global.setImmediate = setImediate
```

renatoalencar:
> Looks like [facebook/jest#11222 (comment)](https://github.com/facebook/jest/pull/11222#issuecomment-807030072) works!

That's awesome! It just finished testing it.

eps1lon (commented):
Haven't looked into the issue but this seems ok. Just some minor issues with the testing approach to consider.

src/__tests__/helpers.js
renatoalencar: I really thought that it had something wrong here, thanks for the suggestion.

src/__tests__/helpers.js
renatoalencar: haha sorry

src/__tests__/helpers.js
renatoalencar: sorry again

renatoalencar:
@eps1lon, could I wrap inside the `run with real timers` block like I just did or should I bring it outside this block?

eps1lon:
This PR would close #914?

renatoalencar:
> This PR would close #914?

Yeah. I also tested temporarily deleting `setImmediate` from `tests/setup-env.js` to see if everything would still works and it looks like that's fine.

kentcdodds (dismissed):
LGTM 👍 And with the 👍👍 from several folks, I'd say we can merge this.

src/helpers.js
```diff
@@ -16,14 +16,19 @@ function runWithRealTimers(callback) {
...
   const timerAPI = {
-    clearImmediate,
     clearInterval,
     clearTimeout,
-    setImmediate,
     setInterval,
     setTimeout,
   }
 
+  if (typeof setImmediate === 'function') {
```
kentcdodds, line 25:
```suggestion
  // For more on why we have the check here,
  // checkout https://github.com/testing-library/dom-testing-library/issues/914
  if (typeof setImmediate === 'function') {
```

kentcdodds:
Hope you don't mind I added a little comment to explain why we have the extra check.

kentcdodds:
@all-contributors please add @renatoalencar for code and tests

renatoalencar:
> Hope you don't mind I added a little comment to explain why we have the extra check.

That's awesome, thanks.

kentcdodds:
@all-contributors please add @gaearon for bugs and review

gaearon:
thanks!

Issue #921, opened 2021-03-26
*ByText options has incorrect type


vovanezha:
- `@testing-library/dom` version: 7.28.1
- Testing Framework and version: jest 26.6.1
  
- DOM Environment: jsdom 16.4.0
  

### Relevant code or config:

```js
screen.getByText('hello', { ignore: 'my-selector' });
```

### What you did:

I'm trying to use `ignore` option in `getByText` method 

### What happened:

```
Argument of type '{ ignore: string; }' is not assignable to parameter of type 'SelectorMatcherOptions'.
  Object literal may only specify known properties, and 'ignore' does not exist in type 'SelectorMatcherOptions'
```

### Reproduction:

https://codesandbox.io/s/misty-pond-hg7s7?file=/src/__tests__/index.tsx:0-220

### Problem description:

The option is described in [official docs](https://testing-library.com/docs/queries/bytext#ignore) and exists in [codebase](https://github.com/testing-library/dom-testing-library/blob/master/src/queries/text.js#L20), but not in the types

PR #924, opened 2021-03-30
fix(types): missing ignore parameter for SelectorMatcherOptions

MatanBobi:
**What**: Add missing `ignore` parameter definition in type `SelectorMatcherOptions`. 
From the [docs](https://testing-library.com/docs/queries/bytext/#ignore) and the codebase it looks like it can either be a string or a boolean.

**Why**: Resolves https://github.com/testing-library/dom-testing-library/issues/921

**How**: Added the missing parameter manually.

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/types/query-helpers.d.ts
+++ b/types/query-helpers.d.ts
@@ -3,6 +3,7 @@ import {waitForOptions} from './wait-for'
 
 export interface SelectorMatcherOptions extends MatcherOptions {
   selector?: string
+  ignore?: string | boolean
 }
 
 export type QueryByAttribute = (


timdeschryver (approved):
LGTM

types/query-helpers.d.ts
```diff
@@ -3,6 +3,7 @@ import {waitForOptions} from './wait-for'
 
 export interface SelectorMatcherOptions extends MatcherOptions {
   selector?: string
+  ignore?: string | boolean
```
nickmccurdy, line 6: Perhaps this would help with code completion? I'm not sure if it's `'script'` or `'script, style'` though, the docs are inconsistent.
```suggestion
  ignore?: 'script' | string | boolean
```

nickmccurdy (changes requested):

types/query-helpers.d.ts
MatanBobi: The default value is 'script, style' if that's what you're looking for:
https://github.com/testing-library/dom-testing-library/blob/master/src/config.ts#L44

nickmccurdy: Ah, I thought TypeScript already supported that, as I saw it in use elsewhere

nickmccurdy (approved):

PR #932, opened 2021-04-19
chore: install lz-string from git repo

anc95:
**What**:
I change lz-string install type from npm to git repo

**Why**:
lz-string@1.4.4 use  WTFPL license, it could cause some law issue when use it. 

**How**:
so I take the advice of it's author, install lz-stirng just from git repo to use MIT license. [related issue](https://github.com/pieroxy/lz-string/issues/147)

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "aria-query": "^4.2.2",
     "chalk": "^4.1.0",
     "dom-accessibility-api": "^0.5.4",
-    "lz-string": "https://github.com/pieroxy/lz-string.git",
+    "lz-string": "https://github.com/pieroxy/lz-string.git#b2e0b270a9f3cf330b778b777385fcba384a1a02",
     "pretty-format": "^26.6.2"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "aria-query": "^4.2.2",
     "chalk": "^4.1.0",
     "dom-accessibility-api": "^0.5.4",
-    "lz-string": "^1.4.4",
+    "lz-string": "https://github.com/pieroxy/lz-string.git",
     "pretty-format": "^26.6.2"
   },
   "devDependencies": {


package.json
```diff
@@ -44,7 +44,7 @@
     "aria-query": "^4.2.2",
     "chalk": "^4.1.0",
     "dom-accessibility-api": "^0.5.4",
-    "lz-string": "^1.4.4",
+    "lz-string": "https://github.com/pieroxy/lz-string.git",
```
eps1lon, line 47: Can we install from a tag/branch instead? We're basically giving up SemVer here which could lead to all sorts of different issues.

`lz-string` has been out for years now with 3 million downloads. Not saying that this means it's safe but it looks like it.

@kentcdodds Was license mismatch in dependencies ever a problem for you?

kentcdodds: You can point to a specific commit rather than the main branch which is what I would recommend.

I've never had license issues before.

anc95: Thx for your review, I just pushed a commit to add the lz-string latest commit hash on main branch

MatanBobi: I guess that's also not the best approach because now we're set to a specific release.
Maybe that's the best we can do though in regards to this PR.

anc95: This issue was reported by  my company's `front-end npm's safe detection platform`.  Now, I use yarn resolutions to fixed @testing/library at 7.25.0 which not has `lz-string` as it's dependencies. Can you approve this mr and publish the new version then I can keep @testing/library to latest version, or there would have any other better solution for this issue? Thanks

eps1lon: I'm a bit hesitant to merge that since I don't know if every package manager can resolve that or creates other problems. Considering you already discovered yarn resolutions I would recommend using that for `lz-string`:

```json
"resolutions": {
  "**/lz-string": "https://github.com/pieroxy/lz-string.git"
}
```

eps1lon:
Thanks for the work.

I'm not comfortable with declaring git dependencies. Since there is a workaround (https://github.com/testing-library/dom-testing-library/pull/932#discussion_r619204527) I'm closing.

PR #938, opened 2021-05-01
test: enable modern jest timers test

timdeschryver:
**What**:

Explicit set jest's faker timers to `modern`.

**Why**:

Because in Jest 26.x the default is `legacy`. 
https://jestjs.io/blog/2020/05/05/jest-26#new-fake-timers

**How**:

Changed the modern test  by passing `modern` as the options.
I also had to advance the timers, and had to "clear" the timers after the test.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/fake-timers.js
+++ b/src/__tests__/fake-timers.js
@@ -1,20 +1,26 @@
 import {waitFor, waitForElementToBeRemoved} from '..'
 import {render} from './helpers/test-utils'
 
-async function runWaitFor({time = 300} = {}, options) {
+async function runWaitFor({time = 300, advanceTimers = true} = {}, options) {
   const response = 'data'
   const doAsyncThing = () =>
     new Promise(r => setTimeout(() => r(response), time))
   let result
   doAsyncThing().then(r => (result = r))
-
+  if (advanceTimers) {
+    jest.advanceTimersByTime(time)
+  }
   await waitFor(() => expect(result).toBe(response), options)
 }
+
+afterEach(() => {
+  jest.useRealTimers()
+})
 
 test('real timers', async () => {
   // the only difference when not using fake timers is this test will
   // have to wait the full length of the timeout
-  await runWaitFor()
+  await runWaitFor({advanceTimers: false})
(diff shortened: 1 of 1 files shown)


src/__tests__/fake-timers.js
```diff
@@ -1,20 +1,26 @@
...
-async function runWaitFor({time = 300} = {}, options) {
+async function runWaitFor({time = 300, advanceTimers = true} = {}, options) {
   const response = 'data'
   const doAsyncThing = () =>
     new Promise(r => setTimeout(() => r(response), time))
   let result
   doAsyncThing().then(r => (result = r))
-
+  if (advanceTimers) {
+    jest.advanceTimersByTime(time)
```
eps1lon, line 11: Is this reachable from non-jest environments or where the `jest` global isn't available?

timdeschryver: I don't think so.
But I was thinking that this shouldn't matter since our tests are run with jest?

eps1lon: Oh, I didn't notice this was just a test file.

eps1lon (approved):

src/__tests__/fake-timers.js
eps1lon, line 11: Shouldn't `waitFor` take care of this? 

@timdeschryver I think this is fixed by https://github.com/testing-library/dom-testing-library/pull/966 and https://github.com/testing-library/dom-testing-library/pull/962 without needing to change these particular tests. Can we close this PR in favor of https://github.com/testing-library/dom-testing-library/pull/962?

timdeschryver: Yes of course.
Thanks for getting back to this PR, because I forgot about it.

timdeschryver: > Shouldn't waitFor take care of this?

It could've been a bug in `waitFor` then, the reason why I added this was because the timers weren't advancing on their own. So, I thought that I had to do that manually.

PR #943, opened 2021-05-05
feat(ByRole): Allow filter by `aria-current` state

### LINKED ISSUE #864: getByRole(role, { current: true })

The [getByRole query](https://testing-library.com/docs/dom-testing-library/api-queries#byrole) is very versatile and helpful in helping us put accessibility at the forefront of our testing. It currently includes options like `selected`, `checked` to select based on state. This is great but it seem to be missing out on one of such states that is usually defined by users to determine active/current section: [aria-current](https://www.aditus.io/aria/aria-current/). It will be great if we can pass that option that checks for this aria attribute
### Suggested implementation:

Since `aria-current` may be used in different ways e.g `aria-current="page"`, `aria-current="location"`, `aria-current="step"` we may consider making the option take a string like:
```js
getByRole("button", { current: "page" })
```
but so far in my presumed usage it seem like this can be sufficient as:
```js
getByRole("button", { current: true })
```
checking only the existence of the `aria-current` attribute as it's unlikely more than one type is being tested at a time.

savcni01:
According to fix #864 issue, this PR adds `aria-current` attribute to byRole queries options.

Since `aria-current` may be used in different ways e.g aria-current="page", aria-current="location", aria-current="step" we may consider making the option take a string like:

getByRole("button", { current: "page" })

but so far this RR implemented another sufficient variant:

getByRole("button", { current: true })
checking only the existence of the aria-current attribute as it's unlikely more than one type is being tested at a time.

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs) - see this [docs PR](https://github.com/testing-library/testing-library-docs/pull/822)
- [x] Tests
- [x] Ready to be merged


types/queries.d.ts
```diff
@@ -88,6 +88,11 @@ export interface ByRoleOptions extends MatcherOptions {
    * pressed in the accessibility tree, i.e., `aria-pressed="true"`
    */
   pressed?: boolean
+  /**
+   * If true only includes elements in the query set that are marked as
+   * current in the accessibility tree, i.e., `aria-current="true"`
+   */
+  current?: boolean
```
eps1lon, line 95: I would rather just query against exact values. Then we automatically support newly introduced values. It's not clear to me that you want `aria-current="page"` to match against `current: true` as well as `aria-current="date"`. Would people actually want `aria-current="page"` and `aria-current="date"` in a `getAllByRole(role, { current: true })` query?

src/queries/role.js
```diff
@@ -64,6 +66,13 @@ function queryAllByRole(
     }
   }
 
+  if (current !== undefined) {
```
MatanBobi, line 69: Is there any reason not to merge these two `if`s into one?

eps1lon: Readability. That way we have a block for every option that is trivial to parse.

types/queries.d.ts
savcni01: It was one of the solution suggestions at the original issue. Both suggestions have it pros and cons.

Now we have two options:
1. Merge current implementation and open additional issue to improve/enhance it to yu
2. Fix it right here in this MR. I think it preferred way. I'll try to do it on weekend.

Thanks for the feedback.

eps1lon:
1. Updated branch
2. Use explicit values instead of casting to `true | false` as discussed in https://github.com/testing-library/dom-testing-library/pull/943#discussion_r626443790
3. `aria-current` defaults to `false` if omitted
4. Drop test for unsupported roles as there are none

src/__tests__/ariaAttributes.js
```diff
@@ -177,6 +177,31 @@ test('`pressed: true|false` matches `pressed` elements with proper role', () =>
...
+  '`aria-current="%s"` matches `current: %j` elements with proper role',
+  (ariaCurrentValue, filterByValue) => {
+    const {getByRole} = renderIntoDocument(
+      ` <a href="/" aria-current="${ariaCurrentValue}"></a>`,
+    )
+    expect(getByRole('link', {current: filterByValue})).toBeInTheDocument()
+  },
+)
+
+test('Missing `aria-current` matches `current: true`', () => {
```
eps1lon, line 198:
```suggestion
test('Missing `aria-current` matches `current: false`', () => {
```

src/role-helpers.js
```diff
@@ -3,6 +3,8 @@ import {computeAccessibleName} from 'dom-accessibility-api'
 import {prettyDOM} from './pretty-dom'
 import {getConfig} from './config'
 
+export const ariaCurrentValues = ['date', 'location', 'page', 'step', 'time']
+
```
eps1lon, line 7:
```suggestion
```

types/queries.d.ts
eps1lon, line 95:
```suggestion
  /**
   * Filters elements by their `aria-current` state. `true` and `false` match `aria-current="true"` and `aria-current="false"` respectively.
   */
  current?: boolean | string
```

eps1lon:
@all-contributors add @savcni01 for code

eps1lon:
@savcni01 Much appreciated, thanks!

Issue #929, opened 2021-04-08
ByText finds children of hidden elements 


mancristiana:
- `@testing-library/react` version: 11.1.0
- Testing Framework and version:
   I am using the built-in testing tools that come with `create-react-app`
    `@testing-library/jest-dom`: 5.11.4
    `@testing-library/user-event`: 12.1.10
    `react-scripts`: 4.0.3
     `jest`: 26.6.0 (hoisted from `react-scripts`)
- DOM Environment: `jsdom`: 16.4.0

### Relevant code or config:

Consider this collapsible element with aria-hidden attribute that displays children props
```js
<div
  aria-hidden={isCollapsed}
  className={isCollapsed ? "hide" : "show"}
>
  {children}
</div>
```

This assertion fails
```js
expect(screen.queryByText("children")).toBeNull();
```

### What you did:

When using `screen.queryByText("children")` I expected not to find hidden elements that specify `aria-hidden={isCollapsed}` attribute. 
However, when running the test I could see that the element was found. 

### What happened:
I got the following failed test:
```
  ● collapses and expands children

    expect(received).toBeNull()

    Received: <div aria-hidden="true" class="hide">children</div>

       5 | test('collapses and expands children', () => {
       6 |   render(<Collapsible>children</Collapsible>)
    >  7 |   expect(screen.queryByText("children")).toBeNull();
         |                                          ^
       8 |
       9 |   userEvent.click(screen.getByRole("button", { name: "Expand" }));
      10 |   expect(screen.getByText("children")).toBeInTheDocument();

      at Object.<anonymous> (src/__tests__/index.test.js:7:42)
```
![image](https://user-images.githubusercontent.com/17469404/114029700-145cf300-987a-11eb-9aff-7e9b61aba5ef.png)

### Reproduction:
I reproduced this issue in this repo: https://github.com/mancristiana/dom-testing-library-template

### Problem description:
Since the `findByRole` query supports this scenario, I expected this would be true for other queries.
The current behavior is a problem because it does not reflect the way a user would interact with our application. In this concrete example, a hidden element would not appear for the user, whereas our test finds it in the dom tree.

### Suggested solution:
Seeing how this behavior is supported for `ByRole` queries I suggest implementing something similar for `ByText` query.

MatanBobi:
@eps1lon So do you think we should push this one forwards or not? You've added the `enhancement` label but from your comment it doesn't sound like it should be implemented.
[...]

PR #944, opened 2021-05-05
feat: ignore element that is children of aria-hidden="true"

simcha90:
**What**:
https://github.com/testing-library/dom-testing-library/issues/929

**Why**:
Implement required feature

**How**:
Added additional iteration by dom to find what elements ignore for rtl look up

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/queries/text.js
+++ b/src/queries/text.js
@@ -9,6 +9,31 @@ import {
   buildQueries,
 } from './all-utils'
 
+function ignoreHiddenElements(container, selector, baseArray) {
+  const hiddenElements = [
+    ...Array.from(container.querySelectorAll("[aria-hidden='true']")),
+  ]
+  baseArray.forEach(base => {
+    if (base.getAttribute('aria-hidden') === 'true') {
+      hiddenElements.push(base)
+    }
+  })
+  let foundHiddenChildren = [...hiddenElements]
+  hiddenElements.forEach(hiddenElement => {
+    foundHiddenChildren = [
+      ...foundHiddenChildren,
+      ...Array.from(hiddenElement.querySelectorAll(selector)),
+    ]
+  })
+  const findSameNode = foundElement =>
+    !foundHiddenChildren.find(foundHiddenElement =>
+      foundElement.isSameNode(foundHiddenElement),
+    )
+  return [...baseArray, ...container.querySelectorAll(selector)].filter(
+    findSameNode,
+  )
+}
+
 function queryAllByText(
   container,
   text,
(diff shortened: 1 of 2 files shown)


src/queries/text.js
```diff
@@ -9,6 +9,31 @@ import {
...
 
+function ignoreHiddenElements(container, selector, baseArray) {
+  const hiddenElements = [
+    ...Array.from(container.querySelectorAll("[aria-hidden='true']")),
+  ]
+  baseArray.forEach(base => {
+    if (base.getAttribute('aria-hidden') === 'true') {
+      hiddenElements.push(base)
+    }
+  })
```
fildon, line 20: You can add in the hidden baseArray elements more concisely with an array filter:
```suggestion
  const hiddenElements = [
    ...Array.from(container.querySelectorAll("[aria-hidden='true']")),
    ...baseArray.filter(base => base.getAttribute('aria-hidden') === 'true'),
  ]
```

src/queries/text.js
```diff
@@ -9,6 +9,31 @@ import {
...
+  let foundHiddenChildren = [...hiddenElements]
+  hiddenElements.forEach(hiddenElement => {
+    foundHiddenChildren = [
+      ...foundHiddenChildren,
+      ...Array.from(hiddenElement.querySelectorAll(selector)),
+    ]
+  })
```
fildon, line 27: Rather than need a `let` and doing reassignment, we can do this all in one go with the power of `flatMap`:
```suggestion
  const foundHiddenChildren = [
    ...hiddenElements,
    ...hiddenElements.flatMap(hiddenElement => Array.from(hiddenElement.querySelectorAll(selector))),
  ]
```

src/queries/text.js
```diff
@@ -9,6 +9,31 @@ import {
...
+  const findSameNode = foundElement =>
+    !foundHiddenChildren.find(foundHiddenElement =>
+      foundElement.isSameNode(foundHiddenElement),
+    )
```
fildon, line 31: `.find` returns the first element from an array that matches the provided callback. In this case you don't particularly care what the element is, just whether it can be found or not. That behaviour is better provided by `.some` which just returns a boolean saying whether the given callback could be matched anywhere in the array:
```suggestion
  const findSameNode = foundElement =>
    !foundHiddenChildren.some(foundHiddenElement =>
      foundElement.isSameNode(foundHiddenElement),
    )
```

Additionally the naming of `findSameNode` doesn't quite feel right to me... I think I would call this something more like "isElementAccessible"

src/queries/text.js
eps1lon: let is fine

src/queries/text.js
eps1lon: current implementation is just fine.

eps1lon:
Thanks for the feedback but we're not going to merge this. Hopefully https://github.com/testing-library/dom-testing-library/issues/929#issuecomment-853732346 explains the why to your satisfaction.

PR #946, opened 2021-05-05
fix: not similar behaviour of ByLabelText  and ByRole queries with nbsp in DOM (fix #890, #904)

### LINKED ISSUE #890: getByRole doesn't normalize the node text
- `@testing-library/dom` version: 7.29.4
- Testing Framework and version: jest 26.6.0 (react-scripts 4.0.2)
  
- DOM Environment: jsdom 16.4.0
  

### Relevant code or config:

```js
it("doesn't work", () => {
  render(<button>Click&nbsp;me</button>);

  expect(
    screen.getByRole(
      "button",
      { name: "Click me" }
    )
  ).toBeInTheDocument();
});
```

### What you did:
Tried querying by role. The component text contains `&nbsp;`

### What happened:
```
Unable to find an accessible element with the role "button" and name "Click me"

  button:

  Name "Click me":
  <button />

<body>
  <div>
    <button>
      Click me
    </button>
  </div>
</body>
```

### Reproduction:
https://codesandbox.io/s/billowing-sound-xtnjm?file=/src/role.test.js

### Problem description:
This contradicts the [documentation](https://testing-library.com/docs/queries/about/#normalization)

> Before running any matching logic against text in the DOM, DOM Testing Library automatically normalizes that text. By default, normalization consists of trimming whitespace from the start and end of text, and collapsing multiple adjacent whitespace characters into a single space.

### Suggested solution:
https://github.com/testing-library/dom-testing-library/blob/5bc93643f312d9ca4210b97681686c9aa8a902d7/src/queries/role.js#L156

should probably be
```
matchNormalizer,
```

### LINKED ISSUE #904: getByRole fails if filtering by accessible name with &nbsp; in it

savcni01:
According to the  fix #904 issue - getByRole fails if filtering by accessible name with &nbsp; in it

Probably this PR also can fix #890 

If we have &nbsp; at label name then
ByLabelText('text with nbsp')  and ByRole('textbox', {name: 'text with nbsp'}) queries behave differently
This PR will solve an issue by sync behavior.

Currently, as the first step of solutions representing tests showing different behavior are added.

**Checklist**:

- [ ] Documentation added to the - "N/A"
- [x] Tests
- [ ] Typescript definitions updated - "N/A"
- [x] Ready to be merged


src/queries/role.js
```diff
@@ -153,7 +153,7 @@ function queryAllByRole(
         }),
         element,
         name,
-        text => text,
+        matchNormalizer,
```
eps1lon, line 156: What else does `matchNormalizer` do? If you just want to fix `nbsp` then we should make this explicit.

savcni01: According to request at issue `matchNormalizer` - here sync behavior for `ByLabelText` and `ByRole`.
the idea that we'll have the same behavior/results for the same inputs.

eps1lon:
See https://github.com/testing-library/dom-testing-library/issues/904#issuecomment-905472296

Issue #863, opened 2020-12-27
Unable to find the "window" object for the given node for input event


MattHeard:
- `@testing-library/dom` version: `7.28.1`
- Testing Framework and version: `jest` `7.0.14`
  
- DOM Environment: `jsdom` `16.4.0`
  

Relevant code or config

```typescript
import React from 'react';
import { createEvent, render } from '@testing-library/react';

describe('input', () => {
  it('creates an event', () => {
    const input: HTMLInputElement = render(<input />);
    const event: React.FormEvent<HTMLInputElement> = createEvent.change(input, {})
  });
});
```

What you did:

I ran the above test with `npm test` (which runs `jest`).

What happened:

```
Unable to find the "window" object for the given node. Please file an issue with the code that's causing you to see this error: https://github.com/testing-library/dom-testing-library/issues/new
```

Reproduction repository:
https://github.com/testing-library/dom-testing-library-template

Problem description:

Suggested solution:

eps1lon:
Thanks for the report.

We should check if we're actually given an HTMLElement because that's not the case here. What you probably wanted to do is 

```diff
-const input = render(<input />);
+const { getByRole } = render(<input />);
+const input = getByRole("textbox");
const event = createEvent.change(input, {});
```

MattHeard:
That's right. Sorry. I'm always confusing `HtmlElement` with `HtmlInputElement` and forgetting the `getByRole` query.

mathiassoeholm:
@eps1lon If I understand the code correctly, you're already checking in `getWindowFromNode`, whether `node` is a DOM node.

I think the problem here is that the error message it a bit cryptic. It doesn't hint that the `node` itself is invalid. Perhaps `getWindowFromNode` should throw an error telling that `node` itself is not a valid DOM Node? :-)

eps1lon:
> If I understand the code correctly, you're already checking in getWindowFromNode, whether node is a DOM node.

It doesn't seem that way.

>  Perhaps getWindowFromNode should throw an error telling that node itself is not a valid DOM Node? :-)

Isn't that directly contradicting the statement you made above? 

The question is if we do want to extend runtime type-checking to `fireEvent.change`. Generally, I tend to lean towards using static type-checking instead which catches these mistakes earlier and we don't need to maintain multiple type-checkers. Right now we already have TypeScript in place.

mathiassoeholm:
To be clear, my suggestion is simply to change the error message from

```javascript
[...]

PR #947, opened 2021-05-05
feat: Improve error message when a node was expected

ShayElkana:
**What**: bug

Closes https://github.com/testing-library/dom-testing-library/issues/863

**Why**: yes, to get a more informative error message

**How**: added a new validation

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

This commit is related to this open issue: 
https://github.com/testing-library/dom-testing-library/issues/863

- I extended the validation getWindowFromNode helper function, to check if the node argument object is an HTMLElement type, if not I'm throwing a new error message.

- Regarding the existing test I couldn't find a valid scenario to preserve the previous test, since if you create an HTML element it has an ownerDocument.defaultView, but the validation at the getWindowFromNode must be preserved since we can't know how will some project used framework would create the HTML element

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -105,11 +105,13 @@ function getWindowFromNode(node) {
     throw new Error(
       `It looks like you passed an Array instead of a DOM node. Did you do something like \`fireEvent.click(screen.getAllBy...\` when you meant to use a \`getBy\` query \`fireEvent.click(screen.getBy...\`?`,
     )
-  } else {
-    // The user passed something unusual to a calling function
+  } else if (node instanceof HTMLElement) {
     throw new Error(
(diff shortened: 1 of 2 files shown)


src/helpers.js
```diff
@@ -105,11 +105,13 @@ function getWindowFromNode(node) {
...
     )
-  } else {
-    // The user passed something unusual to a calling function
+  } else if (node instanceof HTMLElement) {
     throw new Error(
       `Unable to find the "window" object for the given node. Please file an issue with the code that's causing you to see this error: https://github.com/testing-library/dom-testing-library/issues/new`,
     )
+  } else {
+    // The user passed something unusual to a calling function
+    throw new Error(`The given node is not an HTMLElement.`)
```
timdeschryver, line 114: I think we need to provide a bit more guidance here.
Could we add the type or the value of `node` to the message?

eps1lon:
Closing since this PR appears to be stale

PR #950, opened 2021-05-05
feat(config): Add option to silence accessibility warnings (addresses #785)

### LINKED ISSUE #785: Add option to silence `accessibility warnings`
The accessibility warning is nice! But if I'm going to update the `testing-playground` chrome extension to use the latest version of `dom-testing-library`, it will start logging errors because of the `console.warn` expressions. The user won't see them in their console. Instead, they appear on the `extensions` page, looking like something in the extension was broken.

Can we add an option to silence that warning? I still want to have that `warning` attribute on the return value, I just don't want it to invoke `console.warn`.

https://github.com/testing-library/dom-testing-library/blob/8eebd9e5b264ed8435fd3f531a7d0eb112d3faa9/src/suggestions.js#L54-L60

Instead of adding an option, it might also make sense to simply move the printing higher up the tree.

image

savcni01:
This PR addresses issue fix #785 

**Checklist**:

- [x] Documentation added to the - will be added in a few days
      [docs site](https://github.com/testing-library/testing-library-docs) - [Docs PR](https://github.com/testing-library/testing-library-docs/pull/838)
- [x] Tests
- [x] Ready to be merged


smeijer:
Is there anything blocking this? If this PR does what it says it does, it might fix https://github.com/testing-library/testing-playground/issues/330

timdeschryver:
@smeijer I know that there was an agreement to add this feature, but is it still needed since you've overwritten the warn method in https://github.com/testing-library/testing-playground/commit/9efc41a18a9b06e15c19e12c82028c8c30cdadd9?

MatanBobi:
As far as I understand, it looks like @smeijer [didn't want to mock out console functions](https://github.com/testing-library/dom-testing-library/issues/785#issuecomment-708971570) so maybe that's just a workaround for now?
Not sure..

smeijer:
I'd indeed still prefer to not mock out the console. If there is real concern for adding this feature, then I'll accept that. But as there is now a pull request, I think we should either merge (my preference), or close this.

I can reserve some time this week to do a review and get it merged, if there are no objections.

eps1lon:
See https://github.com/testing-library/dom-testing-library/issues/785#issuecomment-905476486

PR #954, opened 2021-05-06
Move to TS

simcha90:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged


src/get-user-code-frame.ts
```diff
@@ -29,12 +33,14 @@ function getCodeFrame(frame) {
 
   let rawFileContents = ''
   try {
-    rawFileContents = readFileSync(filename, 'utf-8')
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    rawFileContents = readFileSync?.(filename, 'utf-8')
```
eps1lon, line 37: This would no longer throw if `readFileSync` is not callable. 

Generally, don't change any runtime when porting to TypeScript. Use type casting instead so that we know where we might want to work on.

eps1lon (changes requested):

src/queries/role.ts
```diff
@@ -134,11 +150,14 @@ function queryAllByRole(
       return true
     })
     .filter(element => {
-      return hidden === false
-        ? isInaccessible(element, {
+      return hidden
+        ? true
+        : !isInaccessible(element, {
```
fildon, line 155: If the first branch of your ternary is `true` you can just use `||` to provide the fallback
i.e. `A ? true : B` is equivalent to `A || B`.
```suggestion
      return hidden || !isInaccessible(element, {
```

nickmccurdy:
Hi, do you think you can work on these merge conflicts or do you need help?

Issue #963, opened 2021-05-28
Not compatible with jest 27


lukeapage:
- `@testing-library/dom` version: 7.31.0
- Testing Framework and version:
 Jest 27.0.1
- DOM Environment:
  jsdom: 16.6

### What you did:

After upgrading to jest v27 I get an error:

```
  ● Test suite failed to run

    TypeError: Cannot destructure property 'DOMElement' of '_prettyFormat.default.plugins' as it is undefined.

      at Object.<anonymous> (node_modules/@testing-library/dom/dist/pretty-dom.js:33:3)
      at Object.<anonymous> (node_modules/@testing-library/dom/dist/config.js:11:18)
```

### Problem description:

I believe jest's pretty-format package changed from exporting on "default" to exporting directly. Even though dom testing library requires a older version, jest requires the newer version, so when dom testing library requires it, it gets the newer version...

### Suggested solution:

I can only think to publish a new major version which will not be compatible with jest 26 but will work with jest 27. I believe you just need to remove `.default`.

timdeschryver:
This should be resolved with the [v8 release](https://github.com/testing-library/dom-testing-library/releases/tag/v8.0.0). Could you try it out please?

lukeapage:
I’ve been using it successfully on the v8 alpha releases so I assume v8 final is ok.

btw we still use legacy timers and things “seem to work” - I guess we are not exercising something that might break?

andrevenancio:
It doesn't work for me. I'm running `@testing-library/dom` 8.1.0 and `jest`27.0.6`.

See multiple `TypeError: Cannot destructure property 'AsymmetricMatcher' of '_prettyFormat.default.plugins' as it is undefined.` warnings on my tests.

lukeapage:
Some testing library packages aren’t upgraded to 8 so you might have an old one in your package Json tree - try checking your yarn.lock /package-lock.json

nickmccurdy:
You can also use `npm ls @testing-library/dom` or `yarn why @testing-library/dom` to see what version you have installed without opening your lockfile. If you're still having issues, please let us know your version.

signal-intrusion:
@nickmccurdy @lukeapage 

Sorry. I have the same problem as @andrevenancio and I'm definitely using testing-library/dom 8.1 and jest 27:
```
➜ npm ls @testing-library/dom
my-package@2.0.0 /Users/blahblahblah
└─┬ @testing-library/react@12.0.0
[...]

PR #964, opened 2021-05-29
chore(packages): support pretty-dom esm build

MatanBobi:
**What**:

This adds support for `pretty-dom` 27 as it's moved to export default only the `format` function.

**Why**:

This will fix a break for everyone using jest 27. Resolves https://github.com/testing-library/dom-testing-library/issues/963

**How**:

Upgrade the dependency and import `plugins` where it's needed.

**Checklist**:

I'm pretty sure this shouldn't be a breaking change.

- [ ] Tests - N/A
- [ ] Typescript definitions updated - N/A
- [X] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "chalk": "^4.1.0",
     "dom-accessibility-api": "^0.5.4",
     "lz-string": "^1.4.4",
-    "pretty-format": "^26.6.2"
+    "pretty-format": "^27.0.1"
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^5.11.6",
--- a/src/pretty-dom.js
+++ b/src/pretty-dom.js
@@ -1,4 +1,4 @@
-import prettyFormat from 'pretty-format'
+import {format as prettyFormat, plugins} from 'pretty-format'
 import {getUserCodeFrame} from './get-user-code-frame'
 import {getDocument} from './helpers'
 
@@ -21,7 +21,7 @@ const getMaxLength = dom =>
     ? 0
     : (typeof process !== 'undefined' && process.env.DEBUG_PRINT_LIMIT) || 7000
 
-const {DOMElement, DOMCollection} = prettyFormat.plugins
+const {DOMElement, DOMCollection} = plugins
 
 function prettyDOM(dom, maxLength, options) {
   if (!dom) {


eps1lon:
> This will break for everyone uses jest 27.

This change **will** break or existing code **does** break? What happens to people still on jest 26?

MatanBobi:
> > This will break for everyone uses jest 27.
> 
> This change **will** break or existing code **does** break? What happens to people still on jest 26?

This change will fix the breaking in jest 27.
I still didn't get a chance to understand why it's breaking though if we're using a specific version of `pretty-format`. Based on [this](https://github.com/testing-library/dom-testing-library/issues/963) issue it looks like jest are using their own `pretty-format` and not the one in our dependencies.

If this is in fact the way it works, sounds like it will break for users using jest 26 (though it didn't fail our tests).
I'll try to link this version to a local project I have and see.

eps1lon:
> If this is in fact the way it works, sounds like it will break for users using jest 26 (though it didn't fail our tests).

That's my main confusion here. If this supposedly breaks jest 26, why doesn't it break our test suite? So let's find out first why https://github.com/testing-library/dom-testing-library/issues/963 pulls in a  different version of `pretty-dom`. Might be a package manager bug or issue with the bundling setup.

lukeapage:
Do you want any more info from me?

```
$ yarn why pretty-format
yarn why v1.19.1
[1/4] Why do we have the module "pretty-format"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "pretty-format@27.0.1"
info Has been hoisted to "pretty-format"
info Reasons this module exists
   - "workspace-aggregator-039abf99-d701-42f4-a83c-b114a8390581" depends on it
   - Hoisted from "_project_#jest-diff#pretty-format"
   - Hoisted from "_project_#jest-circus#pretty-format"
   - Hoisted from "_project_#jest-cli#jest-config#pretty-format"
   - Hoisted from "_project_#jest-circus#jest-each#pretty-format"
   - Hoisted from "_project_#jest-circus#jest-matcher-utils#pretty-format"
   - Hoisted from "_project_#jest-circus#jest-message-util#pretty-format"
   - Hoisted from "_project_#jest-circus#jest-snapshot#pretty-format"
   - Hoisted from "_project_#jest-cli#jest-validate#pretty-format"
   - Hoisted from "_project_#jest-cli#jest-config#jest-jasmine2#pretty-format"
   - Hoisted from "_project_#jest-circus#jest-runner#jest-leak-detector#pretty-format"
info Disk size without dependencies: "164KB"
info Disk size with unique dependencies: "332KB"
info Disk size with transitive dependencies: "1.43MB"
info Number of shared dependencies: 14
=> Found "@types/jest#pretty-format@26.6.2"
info Reasons this module exists
   - "_project_#@types#jest" depends on it
   - Hoisted from "_project_#@types#jest#jest-diff#pretty-format"
info Disk size without dependencies: "140KB"
info Disk size with unique dependencies: "308KB"
info Disk size with transitive dependencies: "1.41MB"
info Number of shared dependencies: 14
=> Found "@testing-library/dom#pretty-format@26.6.2"
info This module exists because "_project_#@testing-library#react#@testing-library#dom" depends on it.
```

So, if node imports testing-library first, then it will cache v26 of pretty-format for all requests within that process. But if jest imports it first, it will cache v27 against node. I thought that was just the way node works - but maybe I am missing something that jest does?

When I add a test resolution to this branch to my jest v27 branch
```
"@testing-library/dom": "https://pkg.csb.dev/testing-library/dom-testing-library/commit/771c1901/@testing-library/dom"
```

it works.

When I add the same resolution to my main branch using jest v26 it still works - my guess is that it is getting v27 of pretty-format but the way it is now imported in this PR is backwards compatible with both v26 and v27.

eps1lon:
> Do you want any more info from me?

A cloneable repro. These error messages do not help me identify the issue. They only show symptoms.

> So, if node imports testing-library first, then it will cache v26 of pretty-format for all requests within that process. 

That's not how node's resolution algorithm works. It resolves each package relative to each other (https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders)

If I have jest -> pretty-format@27 and testing-library -> pretty-format@26 then testing-library will always get v26 and jest v27. Everything else is either a bug in node (highly unlikely) or a bug in your package manager/bundler or possibly jest itself. Otherwise semantic versioning would be useless since every package would just get a single version that just happens to be loaded first.

I think we're just going to inline pretty-format. We've had two or three issues with it in the past and by inlining `pretty-format` we can solve https://github.com/testing-library/dom-testing-library/pull/907. It's now clear to me that `pretty-format` is not suiteable for consumption outside of jest packages.

MatanBobi:
>
> That's not how node's resolution algorithm works. It resolves each package relative to each other (https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders)
> 
> If I have jest -> pretty-format@27 and testing-library -> pretty-format@26 then testing-library will always get v26 and jest v27. Everything else is either a bug in node (highly unlikely) or a bug in your package manager/bundler or possibly jest itself. Otherwise semantic versioning would be useless since every package would just get a single version that just happens to be loaded first.

I was under the same impression you were but then I saw `jest-resolve-dependencies` package and wasn't sure what's going on..
I also saw [this one](https://github.com/facebook/jest/issues/5913) and thought that maybe jest are doing something in their code but wasn't able to understand and didn't have a lot of time..

lukeapage:
You are right from a normal node perspective:

https://nodejs.org/api/modules.html#modules_module_caching_caveats

it sounds like you do not need a repro if you already have plans to fix this. If your stuck and nothing will happen without me making a repro I will attempt to make one.

eps1lon:
jest might resolve dependencies overly aggressive. I need pretty format v27 for https://github.com/testing-library/dom-testing-library/pull/907 anyway so it'll be fixed in the next major anyway. Which makes it also safer for people still on jest 26

MatanBobi:
@eps1lon wdym when you say overly aggressive? 
I'm closing this one if you say you already need pretty-format 27 as part of a different fix.

eps1lon:
> @eps1lon wdym when you say overly aggressive?

It's de-deduplicating packages it shouldn't e.g. v26 and v27 are required by different packages but `jest` just gives every package a single version.

MatanBobi:
> > @eps1lon wdym when you say overly aggressive?
> 
> It's de-deduplicating packages it shouldn't e.g. v26 and v27 are required by different packages but `jest` just gives every package a single version.

This sounds like a bug to me.. Do you know if that's the want behavior?

eps1lon:
> This sounds like a bug to me

I said "might". 

> Do you know if that's the wanted behavior?

I wouldn't know

Issue #939, opened 2021-05-01
Opt-out of Jest FakeTimers


timdeschryver:
### Describe the feature you'd like:

I had to pin the DTL version to 7.29.4 in ATL because some tests that were using `waitFor` and `waitForElementToBeRemoved` were failing (I think this was also the root cause for https://github.com/testing-library/dom-testing-library/issues/901). 
This issue still persists to this date (I just tried to upgrade the DTL to its latest version).

After some time debugging, I think that the problem is that DTL falsely assumes that the Angular tests are using jest's fake timers. This is because Angular [patches](https://github.com/angular/angular/blob/master/packages/zone.js/lib/zone-spec/fake-async-test.ts#L60) the timers API.

### Suggested implementation:

I would like to add an option to the config to opt-out of this behavior since I don't see a better way to check if fake timers are used.

### Describe alternatives you've considered:

Could we patch the `useFakeTimers` method, and use that to know if fake timers are used? This would replace the [existing check](https://github.com/testing-library/dom-testing-library/blob/master/src/helpers.js#L43-L45).

### Teachability, Documentation, Adoption, Migration Strategy:

/

timdeschryver:
I think the problem is also a bit larger than using fake timers.
Even calling `jest.useRealTimers()` will mess up some tests. 
Angular has a built-in way to advance time (not by using `jest.advanceTimers`), by invoking `jest.useRealTimers()` the Angular timers are confused and don't advance in time correctly.

eps1lon:
I'm more and more convinced that we should remove all the jest timer handling and replace with a timer API that people configure.

Assuming jest is just not appropriate anymore with our reach and it causes us too often problems. Something like

```js
// test.js
beforeEach(() => {
  domTestingLibrary.useFakeTimers(jest)
})
test(...);
afterEach(() => {
  domTestingLibrary.useRealTimers()
})

// @testing-library/dom 
interface TimersAdapter {
  // whatever methods we currently use from jest
}

export function useFakeTimers(newTimersAdapter: TimersAdapter) {
  timersAdapter = newTimersAdapter;
  // This call is for convenience
  timersAdapter.useFakeTimers();
}

export function useRealTimers() {
  timersAdapter = noopTimers;
  timersAdapter.useRealTimers();
}
```

And then different timer libraries would implement their own `TimersAdapter`.

We can then think about warning if we detect that timers were mocked but no call to `domTestingLibrary.useFakeTimers()` happened.

[...]

Issue #880, opened 2021-01-19
unexpected behavior in waitFor


mlarcher:
- `@testing-library/dom` version: 7.29.4
- Testing Framework and version: Jest 26.6.3 (react-scripts 4.0.1)
- DOM Environment:
  jsdom@16.4.0
- node v14.15.4

### What you did:
I was investigating random test fails we have on the CI (see https://github.com/testing-library/react-testing-library/issues/848 for details)

### What happened:

I noticed that the function I am passing to waitFor was being called even after the timeout occurred.

### Suggested solution:

when going through `handleTimeout`, `clearInterval` and `observer.disconnect()` are triggered in the `onDone`  function at https://github.com/testing-library/dom-testing-library/blob/master/src/wait-for.js#L112  
Before the call to `onDone`, there are a few lines of code 
https://github.com/testing-library/dom-testing-library/blob/master/src/wait-for.js#L162 that are mostly here to copy the stack trace.  
Triggering `clearInterval` and `observer.disconnect()` at the beginning of the `handleTimeout` function makes all the difference between having useless callback invocation after the timeout or not. I suppose that copying the stack trace is long enough to let other calls to the interval sleep through.

Note that in my test case I replaced the setInterval call with a setTimeout calling itself again. When things begin to be dangerously slow, function called with setInterval may run slower than the interval, and begin to ~~run in parallel~~ accumulate in the timers list, slowing things even more... 

On a side note, I was debugging using the compiled js version of the code and editing it directly in my node_modules folder. I noticed that the setTimeout/clearTimeout functions were coming from the `helper.js` file and written as `const overallTimeoutTimer = (0, _helpers.setTimeout)(handleTimeout, timeout);` while the setInterval/clearInterval weren't. It makes sens to be sure we are not using a mocked timer, but shouldn't it be the case for a setInterval too ? I believe a self calling setTimeout is a better choice in all circumstances, but if there is a good reason to use setInterval instead that I'm not aware of, there may be something fishy there.

In the end those adjustments didn't really help solve my case, but I thought they may be worth sharing.
Also, a quick search before submitting this issue led me to interesting tickets that may or may not be related to the point highlighted above:
https://github.com/testing-library/dom-testing-library/issues/820
https://github.com/testing-library/dom-testing-library/issues/698
https://github.com/testing-library/dom-testing-library/issues/808

Please let me know if those points are worth discussing or if any more info is needed.

[...]

Issue #756, opened 2020-08-26
Regression: "ReferenceError: global is not defined"


EvgenyOrekhov:
- `@testing-library/dom` version: 7.22.6
- Testing Framework and version: QUnit 2.11.0
  
- DOM Environment: Chrome 84.0.4147.135 (via Testem)
  

### Relevant code or config:

```js
import { screen } from "@testing-library/dom";
```

### What you did:

I ran my tests: `ember test --server`.

### What happened:

Tests couldn't be executed, the following error was thrown:

```
ReferenceError: global is not defined
    at Object../node_modules/pretty-format/build/plugins/AsymmetricMatcher.js (test-support.js:37305)
    at __webpack_require__ (test-support.js:18993)
    at Object../node_modules/pretty-format/build/index.js (test-support.js:36889)
    at __webpack_require__ (test-support.js:18993)
    at Module../node_modules/@testing-library/dom/dist/@testing-library/dom.esm.js (test-support.js:19861)
    at __webpack_require__ (test-support.js:18993)
    at Module.callback (test-support.js:19106)
    at Module.exports (vendor.js:118)
    at Module._reify (vendor.js:155)
    at Module.reify (vendor.js:142)
```

### Problem description:

`@testing-library/dom` 7.22.3 was working fine.

`@testing-library/dom` 7.22.6 doesn't work.

### Suggested solution:

I can see there was a bump of pretty-format between 7.22.3 and 7.22.6: https://github.com/testing-library/dom-testing-library/commit/9cbc4280d4eceddc5511c21114455f73cd31d196, perhaps it's the culprit.

eps1lon:
Could you include a minimal repro? There are no references to `global` in https://unpkg.com/pretty-format@26.4.2/build/index.js but there are some in https://unpkg.com/@testing-library/dom@7.22.6/dist/@testing-library/dom.esm.js. 

Though the references in the bundled `@testing-library/dom` build have been there for a year now. They assume that we either run in a browser (`typeof window !== 'undefined'`) or in node (where `global` is defined). Though standalone JS engines don't have `global`. Though we don't actually support standalone JS shells. We do require an environment where `setTimeout` is implemented. And as far as I can tell all these environments either have `global` or `window`.

EvgenyOrekhov:
@eps1lon Of course, here: https://github.com/EvgenyOrekhov/ember-with-testing-library.

[...]

PR #962, opened 2021-05-28
chore: Update kcd-scripts to 11.x

eps1lon:
Gradually pushing one major bump at a time.

**What**:

Bump kcd-scripts to 11.x

**Why**:

Verify jest 27 compat

**How**:

Bump kcd-scripts gradually to 11.x

**Checklist**:

- [ ] Documentation added to the
      [docs site](https://github.com/testing-library/testing-library-docs)
- [x] Tests
- ~[ ] Typescript definitions updated~
- [x] Ready to be merged

PR #966, opened 2021-06-01
fix: Only use a single clock

eps1lon:
BREAKING CHANGE:
1. Remove deprecated `waitFormDOMChange`
2. The `timeout` in `waitFor(callback, { interval, timeout } )` now uses the same clock as `interval`
    Previously `timeout` was always using the real clock while `interval` was using the global clock which could've been mocked out. For the old behavior I'd recommend `waitFor(callback, { interval, timeout: Number.PositiveInfinity })` and rely on your test runner to timeout considering real timers.

TODO: 
- [ ] Should `Number.PositiveInfinity` be the new default for `timeout`? 

**What**:

Only run with a single clock.

Closes https://github.com/testing-library/dom-testing-library/issues/939
Closes https://github.com/testing-library/dom-testing-library/issues/756
Closes https://github.com/testing-library/dom-testing-library/issues/880

Hopefully also fixes flaky test failures such as https://github.com/testing-library/dom-testing-library/runs/2727330116 

**Why**:

The previous approach breaks jest 27. I would also consider it suprising behavior (agreeing with Closes https://github.com/testing-library/dom-testing-library/issues/880). If I pass `waitFor(callback, { interval: 50, timeout: 900 })` then I expect both times to run with the same clock. Previously the behavior of `waitFor` was different depending on fake or real timers which I consider problematic.

I also don't think we need to force the `timeout` to run with the real clock. Test runners already implement this behavior. So if people don't want the `timeout` to use the same clock as `interval` then they can use `waitFor(callback, { timeout: Number.PositiveInfinity })`.

**How**:

Backported [`f1a8a14` (#962)](https://github.com/testing-library/dom-testing-library/pull/962/commits/f1a8a14da53f5cc85918ef06003adbb0ceba4fe1) from https://github.com/testing-library/dom-testing-library/pull/962. This ensures that the new approach works with jest 26 (this PR) and jest 27 (https://github.com/testing-library/dom-testing-library/pull/962)

- use globally declared `setTimout` instead of `setImmediate` and `setTimeout` from the real clock
- remove deprecated `waitForDOMChange` since it relied on the now removed `runWithRealTimers`

**Checklist**:

- [x] Tests
- [ ] Ready to be merged
      

/cc @testing-library/all-maintainers Since this might be fairly invasive.


src/helpers.js
```diff
@@ -1,85 +1,21 @@
...
-    clearTimeoutFn: globalObj.clearTimeout,
-    setImmediateFn: globalObj.setImmediate || setImmediatePolyfill,
-    setTimeoutFn: globalObj.setTimeout,
+  /* istanbul ignore else */
+  if (typeof jest !== 'undefined' && jest !== null) {
+    return (
+      // legacy timers
+      setTimeout._isMockFunction === true ||
+      // modern timers
+      Object.prototype.hasOwnProperty.call(setTimeout, 'clock')
```
ph-fritsche, line 12: Neither check guarantees that the timer is not mocked by other means.
We've already been there and it caused some confusion:
https://github.com/testing-library/dom-testing-library/blob/bc65b91fed81d900e8eb2aff1813752a4f487869/src/helpers.js#L22-L51

eps1lon: What other means would that be and why should we support them?

timdeschryver (commented):
I'm no expert here, but I can say that the failing tests in the ATL repository are resolved with this change (on Jest 26).
If this gets released, should we also remove the remaining deprecated methods (`wait` and `waitForElement`) ?

eps1lon:
> If this gets released, should we also remove the remaining deprecated methods (wait and waitForElement) ?

Yep, this PR already targets an alpha branch so we'll collect some more breaking changes if this gets released.

src/helpers.js
9still: Additionally to what @eps1lon said, as long as the name of this function is `**jest**FakeTimersAreEnabled`, I think the check covers all possibilities.  If someone mocks timers through another library, or manually, it seems beyond the scope of this method (feels like it needs to be a separate feature).

kentcdodds: I agree with @9still here.

kentcdodds (approved):
> Previously the behavior of waitFor was different depending on fake or real timers which I consider problematic.

This is enough to convince me. The goal of our utilities is to be able to use the async utilities and have all the tests run the same whether the timers are faked or not. And the workaround for the old behavior is pretty sensible IMO.

This looks good to me, and I love all the code that can be removed with this change 👍

src/helpers.js
ph-fritsche: My concern is that this function might return `true` although there are no jest fake timers in place.
That's what the previous solution fixed.

If anyone later relies on this function to tell if jest timers are active and acts on this cognition, we go full circle.
The side-effect described in #884 happened due to `runWithRealTimers` relying on `getJestFakeTimersType` to do what its name said.

So if we can't get rid of a function called `jestFakeTimersAreEnabled`, we should ensure that it only returns `true` if jest fake timers are enabled.

kentcdodds: Sorry, I'm not sure I understand how this could return `true` without fake timers in place 🤔 It looks like this only would return true in a jest environment if `setTimeout._isMockFunction` or `setTimeout.clock` is defined. I suppose someone could mock `setTimeout` themselves (that's odd) or add a `.clock` to `setTimeout` (also odd). Are you suggesting that this should basically be: `return Boolean(getJestFakeTimersType())`?

ph-fritsche: That's what happened before. `jest.useFakeTimers('modern')` was used in another test and then this helper recognized the fake timer to be enabled. `runWithRealTimers` then acted upon this and caused a leak.

Given `runWithRealTimers` is removed the problem might be less likely to resurface again.
But who knows what might be implemented in the future? I would not want to keep in mind that the `jestFakeTimersAreEnabled` helper is not reliable.

Better remove the helper completely or keep the implementation that is safe to use no matter what the user or any other library did to `global.setTimeout`.

eps1lon: > Better remove the helper completely or keep the implementation that is safe to use no matter what the user or any other library did to global.setTimeout.

I think in the future it might be better to explicitly configure the clock and then just try to detect the clock when using jest in its default configuration. But we should probably do this in a follow-up. 

Right now I just hope we can land this without too much breakage. Especially since we can remove so much code while also closing 3 issues.

eps1lon: > I don't understand why we would risk reintroducing a bug just for removing the lines that have been explicitly introduced to prevent this.

Let's unpack this a bit. I'm not explicitly risking a bug by removing the code. I just think the previous code didn't serve any purpose. This assumption holds with our test suite. Hence why I'm asking what cases I am missing.

I do risk a bug implicitly though just by virtue of changing code. I'm very well aware of that. Considering this changes gets rid of a whopping 300 lines of code while fixing 3 open issues, I'm fairly confident the risk is worth it. Any risk remaining will be tested in the alpha phase.

> Could we not just update the implementation if there are issues with jest27 ?

There are issues with jest 27. No "if". Obviously I would prefer to "just update the implementation". But as far as I can tell the previous  implementation is incompatible with jest 27 while also causing existing issues (https://github.com/testing-library/dom-testing-library/issues/880, https://github.com/testing-library/dom-testing-library/issues/756). The proposed implementation in this PR is compatible with jest 26 and jest 27.

ph-fritsche: I haven't spent enough time examining this to tell if this will actually cause problems or not with the current `waitFor` implementation.

I just stumbled over that revert to the old check and remembered how difficult it was to identify the culprit for the leaking test last time because the code seemed too simple and it needed a specific setup to fail.
I'm not confident that an alpha phase will catch similar things - last time the insufficient check was introduced Sep, 2019 and extended to modern versions (causing the bug) in Jun, 2020 - the issue was tracked down in Feb, 2021 just because the bug was also relevant in this library's own test suite. But it would not throw an error - just subsequent tests were silently ignored. (If it wasn't for test coverage, that might have went unnoticed.)

If we can kick out the remainder of the helper, I'm all for it.
But I would hate to require everybody to remember that this helper should not be used for the thing its name suggests.

> There are issues with jest 27. No "if". Obviously I would prefer to "just update the implementation". But as far as I can tell the previous implementation is incompatible with jest 27 while also causing existing issues (#880, #756). The proposed implementation in this PR is compatible with jest 26 and jest 27.

I haven't seen yet which part of the helper is incompatible with jest27. Could you give me a hint which part is considered incompatible? I'll look into it later and try to come up with a fix.
(I checked out the branch yesterday. When updating the `kcd-scripts` tests would fail everywhere. How did you run the test suite with jest27?)

eps1lon: > I haven't spent enough time examining this to tell if this will actually cause problems or not with the current waitFor implementation.

Would you agree that it may be more constructive to find these issues first, submit a test and then refer to it here instead of voicing concern first and putting the burden on me?

> I haven't seen yet which part of the helper is incompatible with jest27.

Every part.

> How did you run the test suite with jest27?

See https://github.com/testing-library/dom-testing-library/pull/962 as explained in the initial issue description

ph-fritsche: > Would you agree that it may be more constructive to find these issues first, submit a test and then refer to it here instead of voicing concern first and putting the burden on me?

I did find the issue, submitted a test that covered that scenario and you deleted it with this change.
And then I raised the concern that deleting that test and reverting to a code snippet that failed before might fail in the future.

I assumed you know best about the part of the code you edited.
I just know from time previously spent on this that this old helper implementation isn't reliable.
It doesn't exactly do what it suggests. That's why I think it shouldn't be in the codebase.

> > I haven't seen yet which part of the helper is incompatible with jest27.
> 
> Every part.

So much for constructive... :facepalm: 
According to the docs jest-next fake timers work like before - I'll check their code to see if every part changed.

eps1lon: > I did find the issue, submitted a test that covered that scenario and you deleted it with this change.

Could you re-submit it as an integration test? I just deleted the unit test that become obsolete with the removal of `useRealTimers`. As far as I can tell this function isn't part of the public API and therefore removing tests is safe if the function is removed.

> I assumed you know best about the part of the code you edited.

Yeah, you shouldn't do that for any edit I make. I don't pressume to know the code I edit to know "best". Otherwise I wouldn't ask for review or collaborate with other people in general.

> So much for constructive... facepalm

May I remind you to be respectul. I don't know how else to say this. The tests were just failing left and right. Maybe there is one particular line in jest that changed but I shouldn't need to debug their implementation because their implementation is irrelevant to our usage. If you want to work on another implementation then please go ahead. In the end I don't care about the implementation. Only about the observable behavior. But I'm not doing that work for you just because you don't like the current implementation.

I know this is kind of easy to say for me but don't take removal of your code personally. Stuff just changes over time. Just because we remove code, doesn't mean the code was useless. It obviously served a purpose in the past. But it isn't unlikely that implementation details change over time.

ph-fritsche: Keep twisting words. Your standards for respectful behavior seem to differ quite a bit depending on if your addressor or addressee.

You might consider this a fundamentally incorrect assumption, but I'll continue to assume that someone is competent and has thought well about a change before writing it. So unless I'm certain to know better I will not easily disregard their thoughts and rather raise a question than redo their work to prove them wrong.

If you consider questions and acknowledgement of limitations of ones one knowledge as unconstructive and a burden to you, that might constrain the possible collaboration quite a bit.

Now you can go ahead and just delete the code because the only possible reason one might be concerned about the removal is the code being precious and taking the removal personally.

9still: > I haven't seen yet which part of the helper is incompatible with jest27.

@ph-fritsche From the point of view of a user - if you are referring to the implementation added in #887, it completely broke modern timers (& they became the default in Jest 27).   #961 describes the reason in detail, but basically half of our fake timer tests would hang with this implementation.

ph-fritsche: @9still Thanks a lot for that detailed report.

That's one ugly mess with those timers.
The short and simple suggested solution is not feasible. Its the root of [this](https://github.com/ph-fritsche/dom-testing-library/blob/24ebf282d10782b3d0a22b939a1b2950d4b5bd1b/src/__tests__/helpers.js#L86-L118).

`jest.useRealTimers` is the only reliable way to tell if jest's fake timers are enabled - at least the only one I could find. Preserving that `clock` as you suggested might work.

I like the option to plug in your own callback for advancing timers during `waitFor`. It's the cleanest solution, but I guess users would like a sensible fallback.

Maybe the "guess" per the simple solution is enough for the new `waitFor` implementation.
I would just have preferred to name it like that - a guess - or at least slap a code comment on it if we can't fix it or don't need it to be reliable, so that we don't end up running in circles.

eps1lon:
Update: I also noticed flaky test failures such as https://github.com/testing-library/dom-testing-library/runs/2727330116 (also remember one PR) and these instances are hopefully also fixed by this change

kentcdodds:
Looks like everything's been resolved here and considering this is just going to the alpha branch I'm going to go ahead with the merge 👍

Thanks for the collaboration!

ValentinH:
While updating Jest to v27, I faced many broken tests where `waitForElementToBeRemoved()` was used along with fake timers. After finding this PR, I tried to force the usage of `@testing-library/dom@v8.0.0-alpha.1` and it fixed my issues.

As I'm not sure when v8 will be released to the stable channel, I'm wondering if the fix to use a single clock could be back-ported to v7? Or is this fix tightly linked with the removal of `waitFormDOMChange` and must wait for a major version release ?

eps1lon:
> I'm wondering if the fix to use a single clock could be back-ported to v7?

It includes breaking changes that are not coupled with `waitForDOMChange`. See item 2 on the breaking changes list in the PR description:

> The `timeout` in `waitFor(callback, { interval, timeout } )` now uses the same clock as `interval`
    Previously `timeout` was always using the real clock while `interval` was using the global clock which could've been mocked out. For the old behavior I'd recommend `waitFor(callback, { interval, timeout: Number.PositiveInfinity })` and rely on your test runner to timeout considering real timers.

PR #975, opened 2021-06-12
chore: Don't test with node 10 or 15

nickmccurdy:
**What**: Don't test with node 10 or 15

**Why**: Node 10 and 15 reached EOL so let's free up some CI resources.

**How**: Remove 10 and 15 from node test matrix.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -16,7 +16,7 @@ jobs:
     if: ${{ !contains(github.head_ref, 'all-contributors') }}
     strategy:
       matrix:
-        node: [10.14.2, 12, 14, 15, 16]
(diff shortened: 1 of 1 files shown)


MatanBobi:
I'm not 100% familiar with the approach but should we bump a major version stating that from now on we don't support Node 10 and 15?

eps1lon:
> I'm not 100% familiar with the approach but should we bump a major version stating that from now on we don't support Node 10 and 15?

Let's just drop node 10 support in this PR as well and target the `alpha` branch

eps1lon:
Closed in https://github.com/testing-library/dom-testing-library/pull/976

PR #978, opened 2021-06-13
test(waitFor): Add current behavior for legacy fake timers and requestAnimationFrame

eps1lon:
Identified the bug when trying out v8 alpha in react-spectrum. Waiting for release of https://github.com/facebook/jest/pull/11567

**What**:

Add a test for a bug caused by https://github.com/facebook/jest/issues/11565

**Why**:

Hoping that we can work around this bug in any way. Would appreciate if anybody got some insight.

**How**:

Add a test when where we `waitFor` `requestAnimationFrame`

**Checklist**:

- ~[ ]~ Documentation added to the
      [docs site](https://github.com/testing-library/testing-library-docs)
- [x] Tests
- ~[ ]~ Typescript definitions updated
- [x] Ready to be merged

PR #979, opened 2021-06-13
chore: Release v8 as stable

eps1lon:
Recommendations:
1. If you're using `jest` fake timers make sure you use modern timers
  jest 27:
   ```diff
   -jest.useFakeTimers('legacy')
   +jest.useFakeTimers('modern')
    // jest.config.js
   -"timers": "legacy",
   +"timers": "modern",
   ```
   jest 26:
   ```diff
   -jest.useFakeTimers()
   +jest.useFakeTimers('modern')
    // jest.config.js
   -"timers": "legacy",
   +"timers": "modern",
   ```

BREAKING CHANGE: Remove deprecated `waitFormDOMChange`

BREAKING CHANGE: Remove deprecated `waitForElement`

BREAKING CHANGE: The `timeout` in `waitFor(callback, { interval, timeout } )` now uses the same clock as `interval`. Previously `timeout` was always using the real clock while `interval` was using the global clock which could've been mocked out. For the old behavior I'd recommend `waitFor(callback, { interval, timeout: Number.POSITIVE_INFINITY })` and rely on your test runner to timeout considering real timers.

BREAKING CHANGE: `<script />`, `<style />` and comment nodes are now ignored by default in `prettyDOM` .If you whish to return to the old behavior, use a custom `filterNode` function. In this case `prettyDOM(element, { filterNode: () => true })`.

BREAKING CHANGE: node 10 is no longer supported. It reached its end-of-life on 30.04.2021.

**What**:

Prepare new major release

**Why**:

Ran promising tests in

- [x] mui-org/material-ui https://github.com/mui-org/material-ui/pull/26740
- [x] reduxjs/react-redux https://github.com/eps1lon/react-redux/tree/test/bump-testing-library-alpha
- [x] adobe/react-spectrum https://github.com/eps1lon/react-spectrum/tree/test/bump-testing-library
	 This revealed https://github.com/testing-library/dom-testing-library/pull/978 which only affects legacy fake timers in jest 26 i.e. latest jest releases are unaffected
**How**:

**Merge** (not squash&merge) `alpha` into `main`

**Checklist**:

      https://github.com/testing-library/testing-library-docs/pull/866
- [x] Tests
- [ ] Ready to be merged


timdeschryver (commented):
Should we also add a note that `OptionsReceived` is replaced with `PrettyFormatOptions` ?
I bumped against this problem while I was [preparing the upgrade](https://github.com/testing-library/angular-testing-library/pull/226/commits/0f29ca34ef4e1139a0d3015ea8ce5300b4aff38a) for Angular Testing Library.

eps1lon:
> Should we also add a note that OptionsReceived is replaced with PrettyFormatOptions ?

It's not a breaking change, no. You can still pass `prettyFormat.OptionsReceived`. 

But we should rename it to `PrettyDOMOptions`. They're not options to `pretty-format` anymore.

kentcdodds:
I'm currently testing the impact this will have on https://github.com/kentcdodds/bookshelf as well as another final test before giving this the green light.

kentcdodds:
As usual, react-query + jest fake timers is an issue (don't worry about the long time the test is taking, my computer is currently doing some other resource intensive video conversion):

image

I tried my react-query v3 branch and had the same issue. In both cases, if I remove the fake timers then the test passes just fine.

If I try to upgrade to jest 27, I get an error because this is a CRA app.

I'm *guessing* this means that the combination of these tools/versions is problematic, so I'll probably just need to lock the versions if I don't want to make some major changes which is ultimately inevitable anyway, so I will not block this release on that.

All the other tests for that app work without trouble.

Once this is ready to merge then I think we're good.

kentcdodds (approved):

eps1lon:
I think ultimately `jest` is missing an API for flushing the microtask queue. We were kind of hacking it with `Promise.resolve` + `setImmediate` which isn't even something that's available in a browser. The current hack isn't better though.

I'm hoping that we can offer a workaround in a feature release where people can implement their own clock and therefore go back to the old `Promise.resolve(resolve => setImmediate(resolve))` implementation for flushing microtasks if they need to. 

The main goal of this major is to make the behavior more predictable (e.g. no more mixing of clocks, no more usage of `setImmediate` which isn't even available in browsers).

Thanks for the hint about react-query though. I'll take that repo and your bookshelf app into consideration for future integration testing. I was focusing too much on libraries.

timdeschryver (approved):
I tested the latest alpha release in ATL and all tests are passing, using Jest 26.

PR #991, opened 2021-07-05
Check if observer is undefined before attempting to disconnect

robreinhard:
**What**: We are not checking if MutationObserver is still defined. In some cases, the method that calls the waitFor can be overwritten, causing the attempt to disconnect to be null.

**Why**: We are now checking if the observer is still defined before attempting to disconnect.

**How**:  I added a truthy check for calling the disconnect method and ran test coverage. (not sure what to write here)

**Checklist**:

- [ ] Tests N/A
- [x] Ready to be merged

### DIFF
--- a/src/wait-for.js
+++ b/src/wait-for.js
@@ -109,7 +109,7 @@ function waitFor(
 
       if (!usingJestFakeTimers) {
         clearInterval(intervalId)
-        observer.disconnect()
+        observer && observer.disconnect()
       }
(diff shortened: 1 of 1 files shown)


eps1lon (changes requested):
Please add a test. I don't see how `observer` can be `null` at this point.

eps1lon:
Closing since this PR appears to be stale.

Issue #413, opened 2019-12-11
Feature request: Shadow DOM support


apalaniuk:
### Describe the feature you'd like:

Web components are an increasingly popular way of creating reusable elements. Due to the existence of shadow DOM, user-focused testing is more difficult and requires tooling focused around the issue, as in the most permissive case, element queries/observer events (eg. via `MutationObserver`s) stop at shadow boundaries and must explicitly be projected into shadow trees. Without support from tooling, aspects of a web app written using web components are virtually untestable without using methods contrary to the goals of this project.

### Suggested implementation:

* Standard element queries can be modified to be projected into successive shadow trees
* Mutation observers to support eg. waits for DOM tree changes can be attached to all shadow roots of descendants
* Solution must consider related concepts (eg. slots) insofar as they relate to testing goals and principle of least surprise according to their documented behaviour/DOM standards
* No expectation of closed shadow roots to be supported in any way, but ideally would be explicitly documented

### Describe alternatives you've considered:

* As first-class support is required, alternatives involve using/writing an alternative implementation which supports shadow trees explicitly (or potentially, writing a plugin system to allow others to support it)
* As shadow tree concepts are an aspect of WHATWG/W3C DOM standards, it makes sense that they're supported first-class by tooling that supports DOM operations

I'd be happy to add this feature, and support for basic functionality (query, waits) was straightforward to implement, but I wanted to get insight on constraints/concerns in advance as I don't currently use this library.

ashubham:
+1

It will be great to build support for WebComponent based frameworks like `lit-element`. If there is a adapter/plugin API, I am sure the community will jump on to build it.

alexkrolick:
How much of this depends on JSDOM itself. This library doesn't implement the DOM part, just traversal.

apalaniuk:
Hi @alexkrolick, it's a good question. JSDOM currently has shadow DOM support, with a few restrictions. The largest issue for testing right now is that this is almost entirely used in the context of web components, and support for custom elements is in-flight (https://github.com/jsdom/jsdom/pull/2548).

That said, the only changes required in here for shadow DOM support should be unrelated, since this would just need to support automatically traversing through shadow trees for queries/attaching MutationObservers/etc, and shadow roots can be attached to many standard elements.

Since the overall change seems small, it's probably better to just implement in a fork and get more direct feedback based on an implementation.

kentcdodds:
[...]

PR #999, opened 2021-08-02
Include roles from shadowDOM in suggestion output

kilmc:
**What**:
Adds a check which improves Error messages for folks testing Web Components using `getByRole`.

**Note:** This PR is focused on getting a correct Error message and not a passing test. This fix will be most useful to folks using `dom-testing-library` in conjunction with the wrapper library referenced in [this comment](https://github.com/testing-library/dom-testing-library/issues/413#issuecomment-657267140) in #413.

**Why**:
When testing Web Components the returned error is incorrect. The example below walks you through this case. 

```HTML
<div>
  <h2>An example</h2>
  <custom-button />
</div>
```
**Contents of the shadowDOM for `<custom-button>`**
```
<button>Button text</button>
```

Below is a test assertion with a minor typo in it.
```javascript
expect(screen.getByRole('button', { name: 'Button texx' })).toBeVisible() // Throws an error
```
Before the change in this PR the error message returned is:
```
There are no accessible roles. But there might be some inaccessible roles. If you wish to access them, then set the `hidden` option to `true`. Learn more about this here: https://testing-library.com/docs/dom-testing-library/api-queries#byrole
```
After the change the error becomes:
```
"Unable to find an accessible element with the role "button" and name "Button texx"

        button:

        Name "Button text":
        <button />

```

**How**:
Adds a check for `node.shadowRoot` in `flattenDom` to ensure `children` inside of Web Components are traversed.

**Checklist**:
- [x] Tests
- [x] Ready to be merged


kilmc:
Is there anything I can do to move this along, even if it's going to be rejected?

PR #1002, opened 2021-08-06
chore: Convert screen.js to TypeScript

riotrah:
**What**:

Converting the following to TS:

* `src/screen.js` -> `src/screen.ts`

**Why**:

#494 outlines motivations and status of project TS conversion, for which this is the start of my contribution.

**How**:

Renamed file. Found type errors and started to fix them whilst trying to keep runtime logic intact.

**Checklist**:

- [x] Tests
- [x] Ready to be merged 
      

0. Should I squash my commits?
1. Do TS defs need to be updated? From what I understand `kcd-scripts` builds d.ts files from the `npm run build` script.
2. I had some coding style/type strictness questions that I've described in the form of "Grok Flow", at this link:

https://dev.app.trygrok.com/pl/YQHa2AMeAQAeanwo/imO5GqKqQI-sE1jrFFUWjg

It's a mind-map style format for describing code-adjacent metadata and comments. At my company we use these for internal not-documentation for major TS migrations or overhauls and also for PR/MRs to describe changes/questions. We made this beta web viewer version recently after some of wanted to use Flows for our personal OSS contributions. Please let me know if you would prefer that I inline those questions into this issue.

### DIFF
--- a/src/screen.ts
+++ b/src/screen.ts
@@ -50,6 +50,9 @@ export const screen =
   typeof document !== 'undefined' && document.body // eslint-disable-line @typescript-eslint/no-unnecessary-condition
     ? getQueriesForElement(document.body, queries, initialValue)
     : Object.keys(queries).reduce((helpers, key) => {
+        // `key` is for all intents and purposes the type of keyof `helpers`, which itself is the type of `initialValue` plus incoming properties from `queries`
+        // if `Object.keys(something)` returned Array<keyof typeof something> this explicit type assertion would not be necessary
+        // see https://stackoverflow.com/questions/55012174/why-doesnt-object-keys-return-a-keyof-type-in-typescript
         helpers[key as keyof typeof initialValue] = () => {
           throw new TypeError(
             'For queries bound to document.body a global document has to be available... Learn more: https://testing-library.com/s/screen-global-error',
--- a/src/screen.ts
+++ b/src/screen.ts
@@ -20,10 +20,10 @@ function getPlaygroundUrl(markup: string) {
 }
 
 const debug = (
-  element: (Element | HTMLDocument)[],
+  element?: Array<Element | HTMLDocument> | Element | HTMLDocument,
   maxLength?: number,
   options?: OptionsReceived,
-) =>
+): void =>
   Array.isArray(element)
     ? element.forEach(el => logDOM(el, maxLength, options))
     : logDOM(element, maxLength, options)
--- a/src/screen.ts
+++ b/src/screen.ts
@@ -1,33 +1,40 @@
 import {compressToEncodedURIComponent} from 'lz-string'
-import * as queries from './queries'
+import type {OptionsReceived} from 'pretty-format'
 import {getQueriesForElement} from './get-queries-for-element'
-import {logDOM} from './pretty-dom'
 import {getDocument} from './helpers'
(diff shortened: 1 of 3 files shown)


src/screen.ts
```diff
@@ -37,15 +44,22 @@ const logTestingPlaygroundURL = (element = getDocument().body) => {
   )
 }
 
-const initialValue = {debug, logTestingPlaygroundURL}
+const initialValue: {
+  [key in keyof typeof queries | 'debug' | 'logTestingPlaygroundURL']?: Function
+} = {debug, logTestingPlaygroundURL}
```
eps1lon, line 49: Does 
```suggestion
const initialValue = {debug, logTestingPlaygroundURL} as const
```
work as well?

src/screen.ts
```diff
@@ -37,15 +44,22 @@ const logTestingPlaygroundURL = (element = getDocument().body) => {
...
+
 export const screen =
-  typeof document !== 'undefined' && document.body
+  typeof document !== 'undefined' && document.body // eslint-disable-line @typescript-eslint/no-unnecessary-condition
     ? getQueriesForElement(document.body, queries, initialValue)
-    : Object.keys(queries).reduce((helpers, key) => {
+    : typedKeysOf(queries).reduce<typeof initialValue>((helpers, key) => {
```
eps1lon, line 54: There's a reason why `Object.keys` returns `string[]` and not an array of literals: https://stackoverflow.com/questions/55012174/why-doesnt-object-keys-return-a-keyof-type-in-typescript. Let's stick with Object.keys and annotate the types properly (casting, `@ts-expect-error` etc).

eps1lon (changes requested):
Looks good overall except the `Object.keys` helper.

src/screen.ts
riotrah: I think in this case perhaps a `helpers[key as keyof typeof initialValue]` might work best. A @ts-expect-error might work too, however, anyone familiar with the Object.keys caveat may be instantly familiar with the nature and reason of the inline cast. 

Though I'm down for whatever! I'll push my proposed change and you can let me if it works.

In addition, doing the cast here makes the explicit typing of `initialValue` above and the explicit type param passed to `reduce` here unneccessary as well.

riotrah: Double post question: Do y'all care about squashing commits? EG should I squash the aforementioned upcoming commit into the one last pushed?

eps1lon: > Double post question: Do y'all care about squashing commits? EG should I squash the aforementioned upcoming commit into the one last pushed?

During review on github it's usually best to not rebase so that notifications only display the most recent changes. We squash on merge anyway.

eps1lon: > I think in this case perhaps a helpers[key as keyof typeof initialValue] might work best.

Yeah. `@ts-expect-error` would also hide potential future errors. But let's document why we think the cast is safe and link the `Object.keys` caveat.

riotrah: Pushed

eps1lon (approved):
Thanks! I revert the change in snapshots since they are specific to local environment (we're unclear why).

Also matched the types of `screen.d.ts` (`debug` accepts a single `HTMLElement` and `Document`)

eps1lon:
@all-contributors Add @riotrah for code

PR #1008, opened 2021-08-12
TS Conversion > `src/events.js` + edit `src/events.d.ts`

riotrah:
**What**: Conversion of `events.js` to TS per #494 . Also, change `events.d.ts` to reflect what I perceive are more accurate types, and closer to those of the dom-testing-library docs.

**Why**: 1. `events.ts` conversion motivation outlined in #494. 2. `events.d.ts` changes might be unnecessary or innapropriate, 

**How**: 1. Changed filename. 2. Fixed errors. 3. Ran tests. 4. Repeat after necessary changes.

**Checklist**:

- [ ] Documentation added to the
      [docs site](https://github.com/testing-library/testing-library-docs) NOT SURE IF RELEVANT
- [x] Tests
- [x] TypeScript definitions updated
- [ ] Ready to be merged
      

I don't think so, there's some things left in commits for this PR that are really better off changed or even removed, but warrant being present during code review.

1. I am not sure if the events.d.ts is intended to be kept after TS conversions, as it's the primary external facing types that users consume. Its current types are clear, simple and legible, however the changes I've made here are much more verbose and potentially hard to read for those contributing or trying to read it. I'm not sure if changing externally facing types is within acceptable scope for the TS conversion effort, but will react to feedback accordingly.

I have some other comments that serve also as questions and as documentation, committed to the repo in their current form, but after responses to them, I think it's best I edit them to reflect any feedback. I've created them in Grok format as they are not necessarily inline comment-worthy, yet benefit from proximity to code, and their live/web viewable versions are at these links:

1. https://dev.app.trygrok.com/pl/YQHa2AMeAQAeanwo/E4kQYcCbQa61VwiM5-hSgw?uid=zbdKDOqUR0KqJYTWm0dkxA
2. https://dev.app.trygrok.com/pl/YQHa2AMeAQAeanwo/fX-L8uEtSoCe7o0fPSTX8Q?uid=CfePr7tKRb2UFBgXT_6X6w
3. https://dev.app.trygrok.com/pl/YQHa2AMeAQAeanwo/ZtdDbdxxTaGrABrpft0k2g?uid=liEQZPW0S8eVwq6nHKLflg

PR #1009, opened 2021-08-18
TS Conversion > `src/query-helpers.js`

riotrah:
**What**: Convert `src/query-helpers.js` to TS. Includes test snapshot updates, some type declaration file changes.

**Why**: Motivations are outlined in #494 

**How**: Filename changes first. Then error fixes. Then fixes to make tests pass. I also occasionally changed files imported from or exported into from `query-helpers` to include pending changes made from other PRs that made TS changes, just for sanity checks and guiding inspiration whilst working on my changes. Of course, those changes were not included in this PR, nor were they kept around when finally running tests for my changes.

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs) NA
- [ ] Tests NA
- [ ] Ready to be merged
      

No, as I need feedback for the following concerns:

1. There are some sections of code that raise type errors or lint errors, whose solution I cannot tell with what little I know of this library and its usage. Those sections require input from other community members/maintainers, and I've marked them with `// FIXME: ` comments. I'll open threads (if possible, have not used GitHub in this manner before) for review at each point.

2. There are some potentially questionable type definitions I've made where I've tried to square the existing types in `types/query-helpers.d.ts` with the errors raised by inlining those types into `src/query-helpers.ts`. Tests pass without too much finagling with `// @ts-expect-error` or `// eslint-disable`, yet, for the same reason as the previous point, I cannot fully tell if these are acceptable changes/compromises.

3. Similar question to my other TS PRs (#1002  #1008): I've changed public facing type declarations, are these changes acceptable?


eps1lon:
Thanks for the work. I've convered the file in https://github.com/testing-library/dom-testing-library/pull/1016 which was less ambitious to ensure files are considered as moved in GitHub.

PR #1011, opened 2021-08-19
fix: support getting text of inputs of type `reset`

G-Rath:
**What**:

Currently doing `screen.getByText('Reset')` results in an error about being unable to find the text if you're testing for an input like so:

```
export const Form: React.FC<Props> = props => (
  <form onSubmit={props.onSubmit} onReset={props.onReset}>
    {props.children}
    <br />
    <input type="submit" value="Submit" />
    {props.onReset && <input type="reset" value="Reset" />}
  </form>
);
```

But this works fine if you try and select the "submit" input in the same manner: `screen.getByText('Submit')`

**Why**:

https://github.com/testing-library/dom-testing-library/pull/185 added support for treating inputs of type `submit` & `button` as if they have text, but it left out `reset` which is the third "type" that has this behaviour.

**How**:

I've adjusted the query to include `type=reset`.

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs) N/A I think (I couldn't find docs on this specific behaviour)
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -133,15 +133,17 @@ test('can get elements by matching their text across adjacent text nodes', () =>
   expect(queryByText('£24.99')).toBeTruthy()
 })
 
-test('can get input elements with type submit or button', () => {
+test('can get input elements with type submit, button, or reset', () => {
   const {queryByText} = render(`
     <div>
       <input type="submit" value="Send data"/>
+      <input type="reset" value="Clear EVERYTHING"/>
       <input type="button" value="Push me!"/>
       <input type="text" value="user data" />
     </div>
   `)
   expect(queryByText('Send data')).toBeTruthy()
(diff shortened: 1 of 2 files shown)


eps1lon:
`getByRole('button', { name: 'reset' })` should work. If not, please open an issue and fill out the issue template.

eps1lon:
Just saw we do have support for "submit" 😕

eps1lon (dismissed):
Thanks!

eps1lon:
@G-Rath Please fix CI. The checks include the task that are run (https://github.com/testing-library/dom-testing-library/runs/3382484950). To start, run `npm run format` and commit the changes.

eps1lon (approved):

eps1lon:
Thanks!

@all-contributors add @G-Rath for code and test

PR #1016, opened 2021-08-25
chore(types): Convert src/query-helpers to TS

eps1lon:
Built on top of https://github.com/testing-library/dom-testing-library/pull/1009 with the goal of having the file marked as moved (for git-blame and git-diff). https://github.com/testing-library/dom-testing-library/pull/1009 had so many changes the GitHub no longer recognized the file as moved.

**What**:

Type-check `query-helpers.js`

**Why**:

- code safety
- eventually these will be used to create the published declarations

**How**:

Convert `query-helpers.js` to TypeScript. No runtime changes.

**Checklist**:

- ~[ ]~ Tests
  Will investigate further if the declarations for `query-helpers` are actually incorrect
- [x] Ready to be merged


src/queries/alt-text.ts
```diff
@@ -18,16 +18,15 @@ const queryAllByAltText: AllByBoundAttribute = (
...
   `Unable to find an element with the alt text: ${alt}`
 
-const queryAllByAltTextWithSuggestions = wrapAllByQueryWithSuggestion(
-  queryAllByAltText,
-  queryAllByAltText.name,
-  'queryAll',
-)
+const queryAllByAltTextWithSuggestions = wrapAllByQueryWithSuggestion<
+  // @ts-expect-error -- See `wrapAllByQueryWithSuggestion` Argument constraint comment
+  [labelText: Matcher, options?: SelectorMatcherOptions]
```
timdeschryver, line 28: Nit
```suggestion
  [altText: Matcher, options?: SelectorMatcherOptions]
```

src/queries/placeholder-text.ts
```diff
@@ -5,21 +5,17 @@ import {queryAllByAttribute, buildQueries} from './all-utils'
...
   `Unable to find an element with the placeholder text of: ${text}`
 
-const queryAllByPlaceholderTextWithSuggestions = wrapAllByQueryWithSuggestion(
-  queryAllByPlaceholderText,
-  queryAllByPlaceholderText.name,
-  'queryAll',
-)
+const queryAllByPlaceholderTextWithSuggestions = wrapAllByQueryWithSuggestion<
+  // @ts-expect-error -- See `wrapAllByQueryWithSuggestion` Argument constraint comment
+  [placeholder: Matcher, options?: MatcherOptions]
```
timdeschryver, line 17: Another nit,

```suggestion
  [placeholderText: Matcher, options?: MatcherOptions]
```

src/query-helpers.ts
```diff
@@ -116,24 +164,40 @@ const wrapSingleQueryWithSuggestion =
   }
 
 const wrapAllByQueryWithSuggestion =
-  (query, queryAllByName, variant) =>
-  (container, ...args) => {
+  <
+    // We actually want `Arguments extends [args: ...unknown[], options?: Options]`
+    // But that's not supported by TS so we have to `@ts-expect-error` every callsite
+    Arguments extends [...unknown[], WithSuggest],
```
timdeschryver, line 170: I think it would be safer to not use `@ts-expect-error` in the consumers of this type and that they be explicit about the arguments.

```suggestion
    Arguments extends unknown[],
```


The reasoning is that if we make a typo in the typename, or if we forget to import the used type, that our users will see wrong types and that they will be able to invoke the methods with invalid arguments. 

Some examples, for `queryAllByAltTextWithSuggestions` with the current implementation of

```ts
const queryAllByAltTextWithSuggestions = wrapAllByQueryWithSuggestion<
  // @ts-expect-error -- See `wrapAllByQueryWithSuggestion` Argument constraint comment
  [labelText: Matcher, options?: SelectorMatcherOptions]
>(queryAllByAltText, queryAllByAltText.name, 'queryAll')
```

If the `Matcher` type and `SelectorMatcherOptions` are not imported in the file, consumers will see the following types.

```ts
const queryAllByAltTextWithSuggestions: (container: HTMLElement, labelText: any, options?: any) => HTMLElement[]
```

And they will be able to consume this query with invalid params.

```ts
// Date is not a Matcher
queryAllByAltTextWithSuggestions({} as unknown as HTMLElement, new Date())
```

When we remove `@ts-expect-error`, we see the type errors clearly in ouor codebase, and we can also fix them. In this case we can fix it by importing the missing types.

```ts
const queryAllByAltTextWithSuggestions = wrapAllByQueryWithSuggestion<
  [labelText: Matcher, options?: SelectorMatcherOptions]
>(queryAllByAltText, queryAllByAltText.name, 'queryAll')
```

Our users now have the correct types:

```ts
const queryAllByAltTextWithSuggestions: (container: HTMLElement, labelText: Matcher, options?: SelectorMatcherOptions | undefined) => HTMLElement[]
```

And they will also get an error when they try to invoke the query with invalid params:

```ts
queryAllByAltTextWithSuggestions({} as unknown as HTMLElement, new Date())
Argument of type 'Date' is not assignable to parameter of type 'Matcher'.
  Type 'Date' is not assignable to type 'number'.ts(2345)
```

eps1lon: These types currently don't face consumers. There's some follow-up to be done since I'm fairly certain the consumed types are incorrect.

> When we remove @ts-expect-error, we see the type errors clearly in ouor codebase, and we can also fix them. In this case we can fix it by importing the missing types.

I'm pretty sure we cannot fix them. I've explained so in the implementation of `wrapAllByQueryWithSuggestion` so please adress those statements directly if you think they're incorrect.

Either way, discuss this in a follow-up (via alternate PRs) so that we don't have to make guesses but see the TypeScript output for our proposals.

PR #1020, opened 2021-08-30
feat(fireEvent): make read-only event target properties definable

9inpachi:
… testing

**What**:

This makes read-only properties of the target element like `scrollWidth` definable through `fireEvent` or `createEvent`.

**Why**:

Because as of now it's not possible to test events with read-only target properties. If a component for example makes use of `scrollWidth` to calculate the scroll width inside a `scroll` event and thus applied some sort of limit to the scroll, it can't be tested for that event.

For example, this test fails normally but passes after the changes in this PR.

```ts
it("scrolls the list to right", async () => {
  render(<TestComponent length={20} />);
  const list = screen.getByRole("list");

  await act(async () => {
    fireEvent.scroll(list, { target: { scrollWidth: 100, scrollLeft: 10 } });
  });

  expect(screen.getByText("List is scrolled")).toBeTruthy();
});
```

**How**:

By using `Object.defineProperties` and modifying the read-only target properties. The solution is the same as or similar to handling `files` but it's generic for all other properties as well.

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged


9inpachi:
I will add tests and other items on the checklist once I can get some sort of confirmation if this is all right. There may have a reason to keep the read-only target properties as is.

Peace. :)

juanca:
> I will add tests and other items on the checklist once I can get some sort of confirmation if this is all right. There may be a reason to keep the read-only target properties as is that I don't know of. So please let me know if this is fine so I can move forward.
> 
> Peace. :)

As far as I can tell, the an [Element's `scrollWidth` property](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollWidth) is read-only. 

What does the rest of your code do to the `scrollWidth`? I would think it shouldn't change after emitting the event.

9inpachi:
> What does the rest of your code do to the scrollWidth?

The code just reads and uses the `scrollWidth` which comes from a scroll event emitted by the browser.

> I would think it shouldn't change after emitting the event.

It doesn't. But I think the emitted event from the test (as shown in the example) should be able to specify it as part of the event itself. So the read-only usage of `scrollWidth` can be tested.

9inpachi:
What I am trying to test is this scroll behavior.

Before horizontal scroll, the left arrow is invisible:

image

After horizontal scroll, the component uses `scrollLeft` and `scrollWidth` to determine if there is sufficient scroll to show the left arrow:

image

For firing the event with this horizontal scroll, I need to specify `scrollWidth` and `scrollLeft` properties.

PR #1022, opened 2021-09-05
Use same interval handler in all supported clocks

eps1lon:
**What**:

Add `advanceTimersWrapper`

**Why**:

I believe this is required for testing concurrent rendering in React 18 but I need to experiment with it first in https://github.com/testing-library/react-testing-library/pull/937

**How**:

Refactor `waitFor` to have more invariants until it's clearer what is responsible for advancing times

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged


eps1lon:
Going to merge soon-ish to unblock an alpha release of https://github.com/testing-library/react-testing-library/pull/937. It's behind an `unstable` prefix and not documented so it is safe to change the API freely.

PR #1027, opened 2021-09-10
test: Run full matrix even if a single job fails

eps1lon:
**What**:

The validate pipeline previously immediately stopped once a single job failed. Now we complete every job even if another fails.

**Why**:

This made it impossible to know if an issue is invariant across the test matrix or tied to a specific value.

**How**:

Disable `fail-fast`

**Checklist**:

- ~[ ]~ Tests
- [x] Ready to be merged


eps1lon:
Merging since the node 16 failure is due to https://github.com/nodejs/node/issues/40030

PR #1028, opened 2021-09-11
test: Ensure broken node 16.9.0 isn't used

eps1lon:
**What**:

Fix GitHub jobs using node 16

**Why**:

They're currently broken due to using the cached and broken node version 16.9.0 which causes jest to fail due to

**How**:

Cherry-pick using a working version which works in a previously failing job: https://github.com/testing-library/dom-testing-library/runs/3574134403

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -18,7 +18,7 @@ jobs:
       # Otherwise we would not know if the problem is tied to the Node.js version
       fail-fast: false
       matrix:
-        node: [12, 14, '^16.9.1']
+        node: [12, 14, '16.9.1']
     runs-on: ubuntu-latest
     steps:
       - name: 🛑 Cancel Previous Runs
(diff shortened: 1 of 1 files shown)


.github/workflows/validate.yml
```diff
@@ -18,7 +18,8 @@ jobs:
       # Otherwise we would not know if the problem is tied to the Node.js version
       fail-fast: false
       matrix:
-        node: [12, 14, 16]
+        # TODO: relax `'16.9.1'` to `16` once GitHub has 16.9.1 cached. 16.9.0 is broken due to https://github.com/nodejs/node/issues/40030
+        node: [12, 14, '16.9.1']
```
eps1lon, line 22: Can't use `'^16.9.1'` because codecov flags do not support carets:

> Flags Must consist only of alphanumeric characters, '_', '-', or '.' and not exceed 45 characters
-- https://docs.codecov.com/docs/flags

Issue #820, opened 2020-11-13
waitFor + getByRole causing severe delays


herecydev:
- `@testing-library/dom` version: 7.26.6
- `node` version: 14.3.0

### Relevant code or config:

```js
it("Causes a big delay", async () => {
      // This fires a fetch request
      userEvent.click(screen.getByRole("button", { name: "Search"}));

      await waitFor(() => {
        //super fast call
        //screen.getByText("some_link_text");

        // causes massive delays
        screen.getByRole("link", { name: "some_link" });
      });
    });
```

### What you did:

Running this test causes the fetch request promise to resolve a lot longer than usual (around 1.5 - 2s). I initially investigated this by wrapping `console.time` around the request.

![image](https://user-images.githubusercontent.com/11328618/99072037-27a9b200-25ab-11eb-817b-1d5772c0ecad.png)

I also had an additional test which looks very similar to the above, but instead of using `getByRole` it was using `getByText` to search for some text. That test took a fraction of the time:

![image](https://user-images.githubusercontent.com/11328618/99072114-52940600-25ab-11eb-8306-132e2cd3f471.png)

### What happened:

After changing `getByRole` to `getByText` **or** inserting a `getByText` query before the `getByRole` it resolved much faster. My working theory is that `getByRole` is so expensive that it was monopolising the `waitFor` loop and takes a long time to relinquish control to the pending promise that needs to resolve.

nickmccurdy:
The issue is that `getByRole` is fairly slow, but it's also syncronous, so `waitFor` keeps rerunning it and has to wait for each call to finish before other code can run. It could potentially use some performance improvements, but perhaps we could make it asyncronous (or add an alternative for back compatibility) so this can be run asyncronously.

herecydev:
Thanks for the quick reply Nick, does that not suggest that any `waitFor` that is sufficiently slow would trigger this? If so, is `waitFor` problematic in it's design?

What do you think the best short term resolution for this?

nickmccurdy:
I'm guessing `waitFor` may be so eager to quickly rerun longer functions that it might actually be slowed down. You could try increasing the `interval` option, it defaults to only 50 milliseconds.

herecydev:
For reference the `getByRole` was taking 400-500ms and would cause the `waitFor` to loop 3-4 times. The best fix (for time) is to insert a more trivial query before that (like a `getByText`) as that flushes that promise much faster.

[...]

PR #1031, opened 2021-09-14
Tweak `waitFor` to not bottleneck async tasks

nstepien:
**What**: I'm addressing issues with `waitFor` bottlenecking async tasks and ultimately timing out in some cases. Ref: https://github.com/testing-library/dom-testing-library/issues/820#issuecomment-911629625

**Why**: The `MutationObserver` callback can fire quite often, and if the check is slow/expensive, it can prevent async tasks from being run to completion in a timely manner, ultimately leading to `waitFor` timing out. Similarly, `setInterval` can queue a new task before the check is complete, so now the "interval" will be between the end of the previous check and the beginning of the next check, instead of between the start of two checks. The "interval" is where async work can actually run, so it's wise to relinquish the main thread to non-test code during that time.

**How**: Added a delay in the `MutationObserver` callback, swapped `setInterval` for `setTimeout`

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

There are 2 non-covered lines now, let me know if you want me to try and add more tests to cover them.

### DIFF
--- a/src/__tests__/wait-for.js
+++ b/src/__tests__/wait-for.js
@@ -255,3 +255,27 @@ test('the real timers => fake timers error shows the original stack trace when c
 
   expect((await waitForError).stack).not.toMatch(__dirname)
 })
+
+test('allow further async tasks to complete after the MutationObserver callback fired', async () => {
+  jest.useRealTimers()
+  renderIntoDocument(`<div id="async">a</div>`)
+  let waitForCount = 0
+  const el = document.getElementById('async')
+  const update = () => {
+    setTimeout(() => {
+      el.textContent += 'a'
+    }, 1)
+  }
+
+  update()
+  update()
+  update()
+
+  await waitFor(() => {
+    waitForCount++
+    expect(el).toHaveTextContent('aaaa')
+  })
+
+  // initial sync check + mutation check
+  expect(waitForCount).toBe(2)
+})
--- a/src/wait-for.js
+++ b/src/wait-for.js
@@ -15,6 +15,15 @@ function copyStackTrace(target, source) {
   target.stack = source.stack.replace(source.message, target.message)
 }
 
+// `requestIdleCallback` is not available in Node/JSDOM environments,
+// so we fallback to setTimeout instead with a little bit of timeout
+// to allow non-test async tasks to run.
+const requestIdleCallback =
+  // eslint-disable-next-line no-undef
+  globalThis.requestIdleCallback ?? (fn => setTimeout(fn, 5))
+// eslint-disable-next-line no-undef
+const cancelIdleCallback = globalThis.cancelIdleCallback ?? clearTimeout
+
 function waitFor(
   callback,
   {
@@ -43,7 +52,7 @@ function waitFor(
   }
 
   return new Promise(async (resolve, reject) => {
-    let lastError, intervalId, observer
+    let lastError, intervalId, observer, idleCallbackId
     let finished = false
     let promiseStatus = 'idle'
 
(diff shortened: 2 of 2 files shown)


nstepien:
# `jest --runInBand`
#### Before
```
Test Suites: 1 failed, 51 passed, 52 total
Tests:       1 failed, 1 skipped, 223 passed, 225 total
Snapshots:   0 total
Time:        374.919 s
```
## After
```
Test Suites: 52 passed, 52 total
Tests:       1 skipped, 224 passed, 225 total
Snapshots:   0 total
Time:        380.614 s
```
# `jest` (2nd run results)
#### Before
```
Test Suites: 1 failed, 51 passed, 52 total
Tests:       1 failed, 1 skipped, 223 passed, 225 total
Snapshots:   0 total
Time:        91.298 s
```
## After
```
Test Suites: 52 passed, 52 total
Tests:       1 skipped, 224 passed, 225 total
Snapshots:   0 total
Time:        90.435 s
```

**Conclusion:** I get fewer flaky tests without needing `configure({ asyncUtilTimeout: 2500 })`, and the performance is about the same.

src/wait-for.js
```diff
@@ -125,6 +141,10 @@ function waitFor(
     }
 
     function checkRealTimersCallback() {
+      clearTimeout(intervalId)
```
eps1lon, line 144: `clearTimeout` on something called "interval" seems off. Either the method is wrong or the naming is off.

eps1lon (changes requested):
Thanks for the work. Some lines aren't covered with tests though.

nstepien:
@eps1lon Coverage should be good now.
The CI is failing, but it seems to be unrelated.

src/wait-for.js
```diff
@@ -101,9 +110,15 @@ function waitFor(
         reject(e)
         return
       }
-      intervalId = setInterval(checkRealTimersCallback, interval)
```
eps1lon, line 104: Let me know if I have this right: We're still calling this every other time but now by scheduling a timeout every time `callback` fails?

So we're now somewhat drifting over time i.e. we're not checking every `interval` but rather at most every `interval` which is `interval + runtime(callback)`?

nstepien: Yes that's correct. This will avoid `setInterval` scheduling tasks while the `callback` is still running, which leads to `callback` running constantly and not relinquishing the main thread for other async tasks to complete.

src/__tests__/wait-for.js
```diff
@@ -255,3 +255,42 @@ test('the real timers => fake timers error shows the original stack trace when c
...
+})
+
+test('avoid running the next check when the previous check is still pending', async () => {
+  jest.useFakeTimers()
+  let currentlyRunningCount = 0
+
+  await waitFor(async () => {
+    currentlyRunningCount++
+    await new Promise(resolve => {
+      setTimeout(resolve, 1)
```
eps1lon, line 290: Shouldn't this be greater than `interval` to illustrate that the `callback` always completed before another one is started?

nstepien: The test fails as expected when this line is removed 🤷‍♂️:
https://github.com/testing-library/dom-testing-library/blob/45830f580ced3ae5337af9ace4a9bd65067dfe5c/src/wait-for.js#L140

When using fake timers, we use `setTimeout(r, 0)` in a `while` loop, which is why this test works:
https://github.com/testing-library/dom-testing-library/blob/45830f580ced3ae5337af9ace4a9bd65067dfe5c/src/wait-for.js#L92

I can still increase the timer if you'd prefer.

eps1lon: I don't understand how this is related. What are you testing and why is it related to flushing microtasks? My original statement would still apply to real timer usage.

nstepien: I had to add this test to reach 100% testing coverage

https://github.com/testing-library/dom-testing-library/blob/45830f580ced3ae5337af9ace4a9bd65067dfe5c/src/wait-for.js#L140
☝️ this line isn't covered anymore without this test.

I've tweaked the test now, please take another look.

tomalexhughes:
Appreciate the work on this so much, we have a large amount of tests using `@testing-library/react` and `msw` and applying this as a patch seems improves the reliability and performance when running locally without `--runInBand`.

Any ETA on when this will be merged and propagate through to the latest version of RTL?

eps1lon (changes requested):
> The MutationObserver callback can fire quite often, and if the check is slow/expensive, it can prevent async tasks from being run to completion in a timely manner, ultimately leading to waitFor timing out.

There's currently no test for this. If this change relates to https://github.com/testing-library/dom-testing-library/pull/1031, then you want to do some expensive work in `waitFor` (e.g. just running a while loop  for N milliseconds where N is greater than the interval).

Also "avoid running the next check when the current check is still pending" is already passing on `main`. I don't see how this test is testing what it says it does.

PR #1035, opened 2021-09-21
feat: Make findBy/getBy return an non-empty array

icatalina:
**What**:
Make `findAllBy`/`getAllBy` return type a "non-empty" array.

**Why**:

These methods will blow if no element is found, so they already return
a non-empty array. The types don't match though. If you enable
`noUncheckedIndexedAccess` you have to manually check the first element
if you want to act upon it.

**How**:
Defining an array as `[T, ...T[]]` typescript considers the first element to be always defined.

**Checklist**:

- [ ] Tests
- [x] Ready to be merged
      

I'm not sure how/if this has to be tested and if I need to update any documentation.

### DIFF
--- a/types/queries.d.ts
+++ b/types/queries.d.ts
@@ -2,6 +2,8 @@ import {ByRoleMatcher, Matcher, MatcherOptions} from './matches'
 import {SelectorMatcherOptions} from './query-helpers'
 import {waitForOptions} from './wait-for'
 
+type NonEmptyArray<T> = [T, ...T[]]
+
 export type QueryByBoundAttribute<T extends HTMLElement = HTMLElement> = (
   container: HTMLElement,
   id: Matcher,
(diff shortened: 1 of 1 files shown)


types/queries.d.ts
```diff
@@ -2,6 +2,8 @@ import {ByRoleMatcher, Matcher, MatcherOptions} from './matches'
 import {SelectorMatcherOptions} from './query-helpers'
 import {waitForOptions} from './wait-for'
 
+type NonEmptyArray<T> = [T, ...T[]]
```
eps1lon, line 5: What's the lowest TypeScript version that supports this syntax?

icatalina: good point, I don't know, give me a sec

icatalina: The syntax looks supported in 3.0.1, fails in 2.9.2. (what's the oldest version we support?)

The [noUncheckedIndexAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) check was introduced in 4.1, so it won't make a difference in anything older than that, or anyone with the check disabled.

icatalina:
Looking at the test failures

types/queries.d.ts
```diff
@@ -12,14 +14,14 @@ export type AllByBoundAttribute<T extends HTMLElement = HTMLElement> = (
   container: HTMLElement,
   id: Matcher,
   options?: MatcherOptions,
-) => T[]
+) => NonEmptyArray<T>
```
timdeschryver, line 17: I think we can't change this because `queryAllBy` queries return an empty array when there are no matches.
https://codesandbox.io/s/react-testing-library-demo-forked-o2sw6?file=/src/__tests__/hello.js

types/queries.d.ts
eps1lon: > The syntax looks supported in 3.0.1, fails in 2.9.2. (what's the oldest version we support?)

I usually follow DefinitelyTyped here (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#support-window) i.e. 3.6 is the lowest supported.

PR #1040, opened 2021-09-29
test: Fix inline snapshot mismatch on certain machines

bpinto:
**What**:

Two snapshot tests were failing locally while passing on CI, this change fixes it so that tests are green on CI as well as locally.
 
The replaced library was appending extra double quotes at the begging and end of the string.

https://github.com/joaogranado/jest-serializer-ansi/blob/c8c5e8cbc0f804f9fe195dfd27625ea0b3d2a062/index.js#L13-L16

**How**:

Replacing jest-serializer-ansi library with jest-snapshot-serializer-ansi.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/pretty-dom.js
+++ b/src/__tests__/pretty-dom.js
@@ -15,11 +15,11 @@ afterEach(() => {
 test('prettyDOM prints out the given DOM element tree', () => {
   const {container} = render('<div>Hello World!</div>')
   expect(prettyDOM(container)).toMatchInlineSnapshot(`
-    "<div>
+    <div>
       <div>
         Hello World!
       </div>
-    </div>"
+    </div>
   `)
 })
 
@@ -28,11 +28,11 @@ test('prettyDOM supports truncating the output length', () => {
   expect(prettyDOM(container, 5)).toMatch(/\.\.\./)
   expect(prettyDOM(container, 0)).toMatch('')
   expect(prettyDOM(container, Number.POSITIVE_INFINITY)).toMatchInlineSnapshot(`
-    "<div>
+    <div>
       <div>
         Hello World!
       </div>
-    </div>"
+    </div>
   `)
 })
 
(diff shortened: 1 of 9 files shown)


src/__tests__/pretty-dom.js
```diff
@@ -45,16 +45,34 @@ test('prettyDOM defaults to document.body', () => {
   </body>"
 `
   renderIntoDocument('<div>Hello World!</div>')
-  expect(prettyDOM()).toMatchInlineSnapshot(defaultInlineSnapshot)
-  expect(prettyDOM(null)).toMatchInlineSnapshot(defaultInlineSnapshot)
+  expect(prettyDOM()).toMatchInlineSnapshot(
```
bpinto, line 48: Hmmm... not sure what happened here, I need to investigate this change.

bpinto: Okay, fixed, it was a problem with automatically updating inline snapshots and then the previous expectation had the extra double quotes.

eps1lon:
That's awesome! Will check tomorrow if this fixes it for me locally. And then we probably want to apply this fix in all testing-library repos

bpinto:
Let me know how it goes! Regarding other testing-library repos, I could not find any other repository making use of ansi snapshot tests, I found this [PR](https://github.com/testing-library/jest-dom/pull/16) removing some snapshot tests on jest-dom repository, maybe this was experiencing a similar issue?

timdeschryver:
If you need someone on a Windows environment to test this, feel free to ping me.

bpinto:
@timdeschryver sorry if the comment wasn't for me, but if I may, I think running tests on Windows could do no harm and would, in fact, being more confidence to these changes. So please, if you could run on Windows, I think that would be great.

timdeschryver (approved):
Works on my machine 👍

eps1lon (approved):
Awesome!

PR #1034, opened 2021-09-20
feat: add generics to screen queries

timdeschryver:
**What**:

Adding generics to screen queries.
Follow up to #1023, which added generics to query methods
Closes #1033

**Why**:

So users that use screen queries can also use generics.

```ts
screen.getByRole('button') as HTMLButtonElement
// becomes
screen.getByRole<HTMLButtonElement>('button')
```

**How**:

- Added a type test and saw it fail.
- Refactored `BoundFunction`  to a generic type that removes the first argument (the container)
- Manually added all of the queries to the `screen` type because it isn't possible to reuse the queries while keeping the generic

**Checklist**:

- [ ] Documentation added to the
      [docs site](https://github.com/testing-library/testing-library-docs)
- [x] Tests
- [x] TypeScript definitions updated
- [] Ready to be merged

PR #1044, opened 2021-09-30
test: add type tests for within

timdeschryver:
**What**:

Follow up to #1034

**Why**:

I wanted to add generics to the within queries, but they already accept a generic.
I think adding a few tests could be helpful for potential changes in the future. 

**How**:

Added a few tests and saw that the tests were passing

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

cc @nstepien

### DIFF
--- a/types/__tests__/type-tests.ts
+++ b/types/__tests__/type-tests.ts
@@ -230,15 +230,15 @@ export async function testWaitFors() {
 }
 
 export async function testWithin() {
-  const sut = within(document.body)
-  sut.queryAllByLabelText('Some label')
+  const container = within(document.body)
+  container.queryAllByLabelText('Some label')
 
-  sut.getByText('Click me')
-  sut.getByText<HTMLButtonElement>('Click me')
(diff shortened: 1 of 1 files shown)


types/__tests__/type-tests.ts
```diff
@@ -228,6 +229,18 @@ export async function testWaitFors() {
   await waitFor(async () => {})
 }
 
+export async function testWithin() {
+  const sut = within(document.body)
```
eps1lon, line 233: What is "sut"?

timdeschryver: It's an abbreviation for "System Under Test", I couldn't think of a better name.
I can change it if you want though.

eps1lon: please 😄

timdeschryver: No problem, renamed to `container`.

eps1lon (approved):

PR #1046, opened 2021-10-04
perf(byRole): check only elements that could be a match

ph-fritsche:
**What**:

Let `querySelectorAll` pre-filter elements before running our own filters on them in `queryAllByRole`.

**Why**:

`byRole` is slow.

**How**:

If `role` parameter is a string, we can predetermine which elements might match our filter functions.
By dismissing other elements in `querySelectorAll` we reduce the time spent in our filter functions.

**Checklist**:

- [N/A] Tests
- [N/A] TypeScript definitions updated
- [x] Ready to be merged

**Additional notes**

I ran `getByAllRoles(container, 'table')` on the [example document for `role-helpers`](https://github.com/testing-library/dom-testing-library/blob/8c5c05dad04f8b976875cb45b2bdfb29b2f8b63d/src/__tests__/role-helpers.js#L19-L69).
The avg time spent dropped from `9.5ms` to `4.2ms`.

### DIFF
--- a/src/queries/role.js
+++ b/src/queries/role.js
@@ -1,5 +1,5 @@
 import {computeAccessibleName} from 'dom-accessibility-api'
-import {roles as allRoles} from 'aria-query'
+import {roles as allRoles, roleElements} from 'aria-query'
 import {
   computeAriaSelected,
   computeAriaChecked,
@@ -99,7 +99,12 @@ function queryAllByRole(
     return subtreeIsInaccessibleCache.get(element)
   }
 
-  return Array.from(container.querySelectorAll('*'))
+  return Array.from(
+    container.querySelectorAll(
+      // Only query elements that can be matched by the following filters
+      makeRoleSelector(role, exact, normalizer ? matchNormalizer : undefined),
+    ),
+  )
     .filter(node => {
       const isRoleSpecifiedExplicitly = node.hasAttribute('role')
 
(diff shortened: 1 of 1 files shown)


MatanBobi:
If I'm not mistaken here, the thing is that we have roles that are implicitly defined (by the elements) and we'll probably lose those if we filter using explicit role selector won't we?

eps1lon:
How did you verify that this improves performance? Please share the results so that we can judge whether this optimization is worth the added code.

ph-fritsche:
> If I'm not mistaken here, the thing is that we have roles that are implicitly defined (by the elements) and we'll probably lose those if we filter using explicit role selector won't we?

Handled here:
https://github.com/testing-library/dom-testing-library/blob/77d329107e9fd74aa185dccc869280063e66957c/src/queries/role.js#L190-L192

The matching on attributes with implicit roles as performed by the filter could not be handled by `querySelectorAll` because the query runs on content attributes and we support querying roles per IDL attributes. Therefore I could not skip the following filter function completely.

But narrowing the selection according to the element types which might be matched is possible.

ph-fritsche:
> How did you verify that this improves performance? Please share the results so that we can judge whether this optimization is worth the added code.

https://github.com/ph-fritsche/dom-testing-library/blob/a63a64f2bc73bfb89feba576eca7fe014ef9d2b8/src/__tests__/role-perf.js

eps1lon:
So this was only tested in JSDOM? Would be nice if you could share a fixture that we can run with this PR in any environment and post the findings you have.

ph-fritsche:
In the browser just the `queryAllByRole`:

[old](https://codesandbox.io/s/perf-byrole-old-2fje3?file=/src/index.js) avg: 3.1051999999284745 vs [new](https://codesandbox.io/s/perf-byrole-new-jd871?file=/src/index.js) avg: 0.20780000004172325

PR #1048, opened 2021-10-05
feat(alt-text): query AMP-IMG by alt text, too

AJ-Jambit:
For my customer, I have to implement <amp-img /> in React. I would love to be able to use the existing RTL tooling.

This is my first PR against this project. I have read the CONTRIBUTING.md and looked at some PRs before opening this one (using GitHub's web UI).

As I am not sure, what the project thinks about AMP (Accelerated Mobile Pages), I am only touching the code lightly before investing too much energy.
We have a workaround in place right now, that sidesteps RTL.

**What**:

The CONTRIBUTING.md does not indicate to open an issue before a PR, so there is no reference to be made.
This PR adds `amp-img` to the elements to query for selection by alt text.

**Why**:

Right now, there is no other way (not even by means of a plugin) to use React Testing Library for AMP projects.

**How**:

I added `amp-img` to the selector list.

**Checklist**:

I left these open for now to sense the waters first.

- [ ] Documentation added to the
      [docs site](https://github.com/testing-library/testing-library-docs)
- [ ] Tests
- [ ] TypeScript definitions updated
- [ ] Ready to be merged

PR #1049, opened 2021-10-05
feat: Support all tags for alt-text

SantoJambit:
**What**: Made the ByAltText methods tag-agnostic

**Why**: ByAltText does currently not support `<amp-img>` (and possibly other custom-elements).

**How**: by using the existing `queryAllByAttribute` function and thus dropping the limitation to only 'img,input,area' tags.

**Checklist**:

- [x] Tests
- [ ] TypeScript definitions updated (no changes required)
- [x] Ready to be merged
      

**Note:**
`queryAllByAltText` and `queryAllByAttribute` are mostly identical, so I chose to just call `queryAllByAttribute`.
However queryAllByAltText had an additional call to `checkContainerType`, which I left in place. Not sure why `queryAllByAttribute` does not have this check.

Alternatively, I could leave the old logic in place and simply add the tag amp-img to the allowed list, but I think this way is better.

I will create an update PR for the docs when it's clear, that this will be merged.

### DIFF
--- a/src/queries/alt-text.ts
+++ b/src/queries/alt-text.ts
@@ -1,21 +1,22 @@
-import {wrapAllByQueryWithSuggestion} from '../query-helpers'
+import {
+  queryAllByAttribute,
+  wrapAllByQueryWithSuggestion,
+} from '../query-helpers'
 import {checkContainerType} from '../helpers'
-import {AllByBoundAttribute, GetErrorFunction} from '../../types'
-import {matches, fuzzyMatches, makeNormalizer, buildQueries} from './all-utils'
+import {
+  AllByBoundAttribute,
+  GetErrorFunction,
+  MatcherOptions,
+} from '../../types'
+import {buildQueries} from './all-utils'
 
 const queryAllByAltText: AllByBoundAttribute = (
   container,
   alt,
-  {exact = true, collapseWhitespace, trim, normalizer} = {},
+  options: MatcherOptions = {},
 ) => {
   checkContainerType(container)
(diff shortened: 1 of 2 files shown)


SantoJambit:
Sorry, there was a mixup on who was supposed to create a PR for this issue.. feel free to choose between this and #1048

src/queries/alt-text.ts
```diff
@@ -1,21 +1,22 @@
...
 ) => {
   checkContainerType(container)
-  const matcher = exact ? matches : fuzzyMatches
-  const matchNormalizer = makeNormalizer({collapseWhitespace, trim, normalizer})
-  return Array.from(
-    container.querySelectorAll<HTMLElement>('img,input,area'),
-  ).filter(node =>
-    matcher(node.getAttribute('alt'), node, alt, matchNormalizer),
-  )
+  return queryAllByAttribute('alt', container, alt, options)
```
eps1lon, line 19: If this PR only targets custom elements then let's restrict it to known and custom elements. Otherwise `<p alt />` becomes queryable.

eps1lon (changes requested):
Do we have a test that e.g. `<div alt />` is not queried? If not then we should add one now that the behavior is not trivial anymore.

SantoJambit:
@eps1lon thanks for the review and the docs PR. I've reviewed the docs PR.

timdeschryver (approved):

eps1lon:
@all-contributors add @SantoJambit for code

Issue #742, opened 2020-08-17
Queries can not find elements in open shadow dom (& weird prettyDOM output)


aryzing:
- `@testing-library/dom` version: `"7.22.0"`
- Testing Framework and version: Jest `"26.2.2"`
  
- DOM Environment: `"16.3.0"`
  

### Relevant code or config:

After appending a custom element to the DOM,
```ts
test("Renders content when attached to DOM", async (): Promise<void> => {
  const testComponent = new TestComponent({ label: "counter" }, { count: 0 });

  if (!testComponent.shadowRoot) {
    throw new Error("Expected an open shadow DOM to be attached.");
  }

  document.body.append(testComponent);

  // Works! But Typescript complains about the type of shadowRoot.
  // Recommendation taken from https://github.com/testing-library/dom-testing-library/issues/413#issuecomment-660039128
  // Unsure if this is even correct usage of testing-library (this is not documented).
  expect(queryAllByText(testComponent.shadowRoot, "counter")).toHaveLength(1);

  // Fails... :(
  expect(screen.queryAllByText("counter")).toHaveLength(1);
});
```

### What you did:

Queried for text that is in the DOM.

### What happened:

Assertion fails b/c query is unable to find element in shadow DOM. Also, interestingly when using `prettyDOM`, the custom element is shown as an object rather than with its tag, `<test-component>`:

```
      <body>
        TestComponent {
          "props": Object {
            "label": "counter",
          },
          "setProps": [Function],
          "setState": [Function],
          "state": Object {
            "count": 0,
          },
        }
      </body>
```

### Reproduction:

(do you have a non-React Codesandbox I could use?)

Managed to pull this together in the one suggested, but even the `expect` that is working on my local fails in the codesandbox. https://codesandbox.io/s/react-testing-library-demo-forked-126wx?file=/src/__tests__/hello.js

### Problem description:

The queries are described as being able to find relevant nodes in DOM, yet are unable to find elements when they are in a custom element's shadow DOM.

### Suggested solution:

I'm sorry, i am unable to provide any suggestions beyond having queries meet the expectation that they find elements in the DOM, including when they are inside the shadow DOM of a custom element.

kentcdodds:
Hi @aryzing,

Check out this thread: https://github.com/testing-library/dom-testing-library/issues/413

TL;DR, the only way to get this support today is to use this: https://www.npmjs.com/package/testing-library__dom

[...]

PR #1054, opened 2021-10-13
feat(queryconfig): Allows configuring a different element query mechanism

MatthiasKainer:
Commit opens the possibility to configure the library to use more sophisticated/different queries on elements, for instance enabling shadow dom specific queries via libraries like https://github.com/Georgegriff/query-selector-shadow-dom

**What**:

Provides the ability to specify a different element query mechanism other than `querySelector` and `querySelectorAll`. 

An example of the usage with Georgegriff/query-selector-shadow-dom looks like this: 

```
import { querySelectorAllDeep, querySelectorDeep } from 'query-selector-shadow-dom';
configure({
    queryElement: querySelectorDeep,
    queryAllElements: querySelectorAllDeep,
})
```

After this line, a query will also drill into the shadow dom of elements on the page

**Why**:

In our project, we are using a design system build with web components (via stenciljs), and different projects in different frameworks (lit, react, vue) want to use this library. However, due to the issues already highlighted in #742 #413 and in a wider part also #999 I tried to come up with an approach that would allow extension for such scenarios without having to introduce modification in the core library.

**How**:

I implemented the changes in a way they are non-breaking. In the `types/config.d.ts` I added the fields `queryElement` and `queryAllElements` as optional so that any existing typings in projects won't break. In the `config.ts` I added it as required fields to the `InternalConfig` to ensure that it is always available. I set them to `querySelector` and `querySelectorAll` to make sure it's doing the same as before.

Then, in the files that were using `querySelector` and `querySelectorAll` I replaced the calls with the method exposed by the config.

The next step would probably be to make the prettyDOM implementation configurable as well, ie to allow fixing of #999 

Ideally, I was thinking about introducing a field "domHandling" that would contain all that, but it felt overblown for now. Open for discussion I guess.

**Checklist**:

- [x] Tests
I added one test, as I felt it wouldn't provide a lot of value given the core functionality is the same. I'd expect the providers of extension to test it. Would you agree?
- [x] Ready to be merged
      

Looking forward to your feedback!

### DIFF
--- a/src/queries/text.ts
+++ b/src/queries/text.ts
@@ -13,7 +13,6 @@ import {
   makeNormalizer,
   getNodeText,
   buildQueries,
-  getConfig,
 } from './all-utils'
 
 const queryAllByText: AllByText = (


MatthiasKainer:
@eps1lon I was thinking about this quite a bit, and looking at the test adjustments needed by directly including a query mechanism for shadow dom let me to the conclusion that it might be a breaking change for many (see https://github.com/testing-library/dom-testing-library/pull/1069/files#diff-fd31e410b487b45eac250e578eaa42419a3321f9664ce1db42f08fd8cc21aefb for an example)

I would thus change this PR to no longer be a draft and propose the merge, and from there continue with the following direction:

- For the next major release, switch the element's query method to deep queries, with the option to provide an option to gracefully switch back to the old query. 
- For the release after that, remove the fallback option and the indirection.

I will remove the draft status now, if this approach is okay for you and the PR can be merged, I will close #1069

src/config.ts
```diff
@@ -1,9 +1,27 @@
...
+  element: T,
+  selector: string,
+) => element.querySelectorAll(selector)
+
 interface InternalConfig extends Config {
   _disableExpensiveErrorDiagnostics: boolean
+  /**
+   * Returns the first element that is a descendant of node that matches selectors.
+   */
+  queryElement: QueryElement
```
eps1lon, line 20: If this is not for public use (see declaration in `InternalConfig` not `Config`) let's prefix it with `_`. Otherwise move its declaration to `Config`

types/__tests__/type-tests.ts
```diff
@@ -249,11 +249,11 @@ export async function testWithin() {
...
 
   container.getByText('Click me')
-  container.getByText<HTMLButtonElement>('Click me')
-  container.getAllByText<HTMLButtonElement>('Click me')
+  container.getByText('Click me')
+  container.getAllByText('Click me')
 
   await container.findByRole('button', {name: /click me/i})
-  container.getByRole<HTMLButtonElement>('button', {name: /click me/i})
+  container.getByRole('button', {name: /click me/i})
```
eps1lon, line 256: Why did these change?

src/__node_tests__/index.js
```diff
@@ -77,6 +85,89 @@ test('works without a browser context on a dom node (JSDOM Fragment)', () => {
...
+    ...element.querySelectorAll(query),
+    ...[...element.children].reduce(
+      (result, child) => [
+        ...result,
+        ...queryMeAndChildrenAndShadow(child, query),
+      ],
+      [],
+    ),
+    ...(element.shadowRoot?.querySelectorAll(query) ?? []),
+  ]
```
eps1lon, line 143: Wouldn't this duplicate elements in e.g. querying for `'button'` in `<div><button /></div>`?

src/config.ts
eps1lon, line 20: What's the use case for `queryElement`? The tests only show `queryAllElement` usage.

eps1lon (changes requested):
Documentation is missing as well. This should be added beforehand by illustrating the problem it's trying to solve.

types/__tests__/type-tests.ts
MatthiasKainer: That's somehow done by the precommit hook locally. I'll either try fix it locally or prep a commit at the end that reverts this change and skip the precommit there and create a task to track it.  

- [ ] Undo this changes

src/config.ts
MatthiasKainer: That's a tricky one. `queryElement` is currently only used by the label helpers. Now in theory the input might be placed in a different shadow dom than the label (for instance, because the input would be wrapped). However, this would lead to a broken link between the label and the element (see https://github.com/whatwg/html/issues/3219) and the test will fail anyways. This discussion has gained some traction again and has led to https://github.com/WICG/webcomponents/issues/917, in which case this function would be needed. 

Theoretically though, if I understand the logic in https://github.com/testing-library/dom-testing-library/blob/main/src/queries/label-text.ts#L142 correct, if you add the queryElement selector then the error message will be nicer. If you want I can play around and see if I can create a test that expect the failure, and add the constraint of the shadow boundary to the error message. The notice would be weird because it only works if you do query into the shadow dom.

src/__node_tests__/index.js
MatthiasKainer: It is supposed to be a naive implementation :D 

I added ":scope" to only query children in https://github.com/testing-library/dom-testing-library/pull/1054/commits/28168530dad4e274268a6df78503c9483bc7e413

MatthiasKainer:
Created docs here: https://github.com/testing-library/testing-library-docs/pull/964

Broke the git history, will clean that up next

src/config.ts
MatthiasKainer: I added the case that would need the `queryElement` here: https://github.com/testing-library/dom-testing-library/pull/1054/commits/eabc513519fe0a690450e0ef1b036f4707ad829d 

As said, from a custom-element perspective it's a rather sad case atm as its supposed to fail

MatthiasKainer:
@eps1lon do you need anything else to move this forward?

surfer19:
hello everyone, is it possible to provide this feature also to @react-testing-library? Does @react-testing somehow derives functionalities from main @dom-testing or is it 100% separated? 
I'm having same scenario that I need to resolve. 

I would expect querying elements behind shadow dom will be more and more common for all kind of projects.
At example very common scenario is to have design system components (based on lit-element) consumed by other app (React,..).

Thank you!

MatthiasKainer:
@surfer19 Yes, react-testing-tools is only a wrapper around the dom-testing-tools (see https://github.com/testing-library/react-testing-library/blob/main/src/pure.js#L119). 

Integrating a react app with react-testing-tools with a stenciljs design system (with shadow dom) is the original usecase for which I created the PR. I tested this scenario with this addition and it works.

surfer19:
@MatthiasKainer cool, does it support nested tree of the web components?
Like:
```
#shadow-root
 ...
   #shadow-root
      ....
         #shadow-root
```

MatthiasKainer:
@surfer19 jop, if you use [query-selector-shadow-dom](https://github.com/Georgegriff/query-selector-shadow-dom) as shown in the description it will reach into arbitrary deep shadow doms. 

If this request is merged I would also use the opportunity to create a pr to change the default query mechanism to shadow-dom, so you will no longer be required to use another library; But as this is a breaking change (ie the order of elements might be changed, see https://github.com/testing-library/dom-testing-library/pull/1069/files#diff-fd31e410b487b45eac250e578eaa42419a3321f9664ce1db42f08fd8cc21aefbR134) and performance might be worse, my proposal after creating both PRs was to go for a stepwise approach. Does that make sense?

MatthiasKainer:
Are there any updates? Is there something else I should add?

surfer19:
@eps1lon @MatthiasKainer 
Any updates? Based on community opinion, it seems to be something that might be worth to finish.
https://github.com/testing-library/dom-testing-library/issues/413#issuecomment-1018774053
https://github.com/testing-library/dom-testing-library/issues/413#issuecomment-1018772952

MatthiasKainer:
I just merged the latest changes main, and the build fails. I will fix this and try again

MatthiasKainer:
okay, I tried to follow codesandbox to the letter, including installing the same node version

```
❯ git clone https://github.com/testing-library/dom-testing-library.git
Cloning into 'dom-testing-library'...
remote: Enumerating objects: 3523, done.
remote: Counting objects: 100% (511/511), done.
remote: Compressing objects: 100% (329/329), done.
remote: Total 3523 (delta 294), reused 303 (delta 167), pack-reused 3012
Receiving objects: 100% (3523/3523), 1.55 MiB | 3.97 MiB/s, done.
Resolving deltas: 100% (2467/2467), done.
❯ cd dom-testing-library
❯ git fetch --force origin pull/1054/head:remotes/origin/pull/1054
remote: Enumerating objects: 46, done.
remote: Counting objects: 100% (46/46), done.
remote: Compressing objects: 100% (12/12), done.
remote: Total 46 (delta 33), reused 46 (delta 33), pack-reused 0
Unpacking objects: 100% (46/46), 8.70 KiB | 387.00 KiB/s, done.
From https://github.com/testing-library/dom-testing-library
 * [new ref]         refs/pull/1054/head -> origin/pull/1054
❯ git checkout -qf remotes/origin/pull/1054
❯ node -v
v16.13.0
❯ nvm install v12.22.7
Downloading and installing node v12.22.7...
Downloading https://nodejs.org/dist/v12.22.7/node-v12.22.7-linux-x64.tar.xz...
################################################################################################################################ 100.0%
Computing checksum with sha256sum
Checksums matched!
Now using node v12.22.7 (npm v6.14.15)
❯ node -v
v12.22.7
❯ yarn -v
1.22.17
❯ yarn install
yarn install v1.22.17
info No lockfile found.
[1/5] Validating package.json...
[2/5] Resolving packages...
warning @testing-library/jest-dom > css > source-map-resolve@0.6.0: See https://github.com/lydell/source-map-resolve#deprecated
warning kcd-scripts > cpy > globby > fast-glob > micromatch > snapdragon > source-map-resolve@0.5.3: See https://github.com/lydell/source-map-resolve#deprecated
warning kcd-scripts > rollup-plugin-node-builtins > browserify-fs > level-filesystem > level-sublevel > xtend > object-keys@0.2.0: Please update to the latest object-keys
warning kcd-scripts > cpy > globby > fast-glob > micromatch > snapdragon > source-map-resolve > resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
warning kcd-scripts > cpy > globby > fast-glob > micromatch > snapdragon > source-map-resolve > source-map-url@0.4.1: See https://github.com/lydell/source-map-url#deprecated
warning kcd-scripts > cpy > globby > fast-glob > micromatch > snapdragon > source-map-resolve > urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
[3/5] Fetching packages...
[4/5] Linking dependencies...
[5/5] Building fresh packages...
success Saved lockfile.
Done in 37.14s.
❯ yarn run build
yarn run v1.22.17
$ kcd-scripts build  --no-ts-defs --ignore "**/__tests__/**,**/__node_tests__/**,**/__mocks__/**" && kcd-scripts build --no-ts-defs --bundle --no-clean
Successfully compiled 29 files with Babel (969ms).
[cjs]
[cjs] /home/mkainer/projects/private/testing-tools/branch/dom-testing-library/src/index.js → dist/@testing-library/dom.cjs.js...
[umd.min]
[umd.min] /home/mkainer/projects/private/testing-tools/branch/dom-testing-library/src/index.js → dist/@testing-library/dom.umd.min.js...
[umd]
[umd] /home/mkainer/projects/private/testing-tools/branch/dom-testing-library/src/index.js → dist/@testing-library/dom.umd.js...
[esm]
[esm] /home/mkainer/projects/private/testing-tools/branch/dom-testing-library/src/index.js → dist/@testing-library/dom.esm.js...
[esm] (!) Plugin replace: @rollup/plugin-replace: 'preventAssignment' currently defaults to false. It is recommended to set this option to `true`, as the next major version will default this option to `true`.
[esm] created dist/@testing-library/dom.esm.js in 2.1s
[esm] cross-env BUILD_ROLLUP=true BUILD_FORMAT=esm BUILD_MINIFY=false NODE_ENV=development BUILD_PREACT=false BUILD_SIZE_SNAPSHOT=undefined BUILD_NODE=false BUILD_REACT_NATIVE=false rollup --config exited with code 0
[cjs] (!) Plugin replace: @rollup/plugin-replace: 'preventAssignment' currently defaults to false. It is recommended to set this option to `true`, as the next major version will default this option to `true`.
[cjs] created dist/@testing-library/dom.cjs.js in 2.2s
[cjs] cross-env BUILD_ROLLUP=true BUILD_FORMAT=cjs BUILD_MINIFY=false NODE_ENV=development BUILD_PREACT=false BUILD_SIZE_SNAPSHOT=undefined BUILD_NODE=false BUILD_REACT_NATIVE=false rollup --config exited with code 0
[umd] (!) Plugin replace: @rollup/plugin-replace: 'preventAssignment' currently defaults to false. It is recommended to set this option to `true`, as the next major version will default this option to `true`.
[umd] (!) Circular dependency
[umd] node_modules/dom-accessibility-api/dist/util.mjs -> node_modules/dom-accessibility-api/dist/getRole.mjs -> node_modules/dom-accessibility-api/dist/util.mjs
[umd] created dist/@testing-library/dom.umd.js in 6.5s
[umd] cross-env BUILD_ROLLUP=true BUILD_FORMAT=umd BUILD_MINIFY=false NODE_ENV=development BUILD_PREACT=false BUILD_SIZE_SNAPSHOT=undefined BUILD_NODE=false BUILD_REACT_NATIVE=false rollup --config --sourcemap exited with code 0
[umd.min] (!) Plugin replace: @rollup/plugin-replace: 'preventAssignment' currently defaults to false. It is recommended to set this option to `true`, as the next major version will default this option to `true`.
[umd.min] (!) Circular dependency
[umd.min] node_modules/dom-accessibility-api/dist/util.mjs -> node_modules/dom-accessibility-api/dist/getRole.mjs -> node_modules/dom-accessibility-api/dist/util.mjs
[umd.min] created dist/@testing-library/dom.umd.min.js in 8.3s
[umd.min] cross-env BUILD_ROLLUP=true BUILD_FORMAT=umd BUILD_MINIFY=true NODE_ENV=production BUILD_PREACT=false BUILD_SIZE_SNAPSHOT=undefined BUILD_NODE=false BUILD_REACT_NATIVE=false rollup --config --sourcemap exited with code 0
Done in 10.68s.
```

Finished successfully. Can someone with permissions retrigger the codesandbox build?

MatthiasKainer:
@eps1lon any updates on how we can proceed here?

MatthiasKainer:
@eps1lon everything should be resolved - can we merge this eventually?

MatthiasKainer:
If there's something missing I'd be super happy to tackle it. Pls let me know how to proceed

manuscriptmastr:
@eps1lon @MatthiasKainer just wanted to bump this PR ✨ I'm really hoping this gets through soon so I can recommend Testing Library for testing web components workshop at my company!

CreativeTechGuy:
While I really support this change, one thing I see missing is the handling of slots. Simply querying for all elements won't work correctly as the slotted content won't be correctly attached. I don't think we can solve this simply by changing the querySelector method but would need special handling for each Testing Library query.

For example, `textContent` of a slot will be whitespace. You'd first need to do `slot.assignedNodes({ flatten: true })` to get the contents of the slot and then traverse those elements.

I don't see why this cannot be a flag in the core library, something like `traverseShadowDOM: true` in the config which would then swap out the handling of each query internally to handle these things. I have a strong need for this and would be happy to contribute if we can get some signal that this would be merged if solved.

alexkrolick:
@CreativeTechGuy thanks for pointing this out. I've noticed that libraries like Playwright and Cypress have ShadowDOM-piercing support. Maybe we should look into what they do.

MatthiasKainer:
The library i mentioned above is used by playwright - https://github.com/Georgegriff/query-selector-shadow-dom 

I also had a PR that was using that directly, and closed it because lack of downwards compatibility

manuscriptmastr:
@alexkrolick @MatthiasKainer Any updates on this PR?

MatthiasKainer:
> @alexkrolick @MatthiasKainer Any updates on this PR?

I don't know what else I could do. From my side this PR could be merged. As it only creates the possibility to extend queries, it shouldn't have any impact to existing installations. As you can use it with a library like https://github.com/Georgegriff/query-selector-shadow-dom you can use it to enable deep queries into shadow Dom's.

If there's anything I should add it would be great if someone could let me know, otherwise I'm just waiting for someone to accept it

odanado:
@alexkrolick Any updates on this PR?

Issue #698, opened 2020-07-15
Performance issue with the byRole query


MatheusPoliCamilo:
- `@testing-library/jest-dom` version: 5.11.0
- `node` version: 12.14.1
- `yarn` version: 1.21.1

### What you did:

I converted several tests of the project I work on, which were using manual queries with `container` and `querySelector`, to use `screen` prioritizing the query `byRole` as [recommended](https://testing-library.com/docs/guide-which-query).

### What happened:

There are some timeout errors when I run some tests after the conversion to use `screen` and the `byRole` selector.

![image](https://user-images.githubusercontent.com/25781749/87480555-eb1d7800-c603-11ea-81b6-7c19095b3d59.png)

### Reproduction:

I made a [small example](https://codesandbox.io/s/byrole-vs-bytext-9irkb?file=/src/__tests__/index.js) to demonstrate the difference in ms in a small case, like a table with 8 cells, when using the query byRole compared to byText.

### Problem description:

At the moment there is a certain performance problem with the `byRole` selector, which even causes a timeout problem in tests that end up simulating a relatively large accessibility tree, a problem that does not happen with other selectors such as byText.

Below is the time it takes me to run a test on the project I work on when using the `byRole` query (in which there is a timeout error in some moments for taking too long).

![image](https://user-images.githubusercontent.com/25781749/87481645-215bf700-c606-11ea-94c3-d3b5f38c26c4.png)
![image](https://user-images.githubusercontent.com/25781749/87482548-05595500-c608-11ea-81a3-3460e1e15a71.png)

Then, the time it takes to perform the same test, but using the `byText` selector. I have run the test several times this way and with this selector I did not give a timeout because it is fast.

![image](https://user-images.githubusercontent.com/25781749/87481691-3a64a800-c606-11ea-8563-5d57926f742a.png)
![image](https://user-images.githubusercontent.com/25781749/87482621-3043a900-c608-11ea-9a2b-3797e49eff4e.png)

### Suggested solution:

I believe that for now the solution would be not to suggest the query `byRole` as a priority, at least until there are some more performance improvements to get closer to the other selectors.

I apologize because I did not get to look at the code in depth to try to understand a little if there is a way to bring about an improvement in `byRole`. I am also aware of the benefits that this selector brings and I am willing to waste more time to run the tests to use it, however the time difference is still large and causing timeout errors. Thank you!!!

eps1lon:
[...]

PR #1068, opened 2021-10-23
feat: Prioritize accessible names check higher than inaccessibility check in `byRole` queries

kevin940726:
**What**:

Related to https://github.com/testing-library/dom-testing-library/issues/698 and https://github.com/testing-library/dom-testing-library/issues/820.

`*ByRole` queries are filtering inaccessibility (`hidden: false`) earlier than accessible names. This PR changes that by prioritizing accessible names over inaccessibility. Hopefully to achieve better performance in most cases.

**Why**:

As stated in https://github.com/testing-library/dom-testing-library/issues/820#issuecomment-726936225, one of the _escape hatches_ to improve the performance of the `*ByRole` queries is to pass `hidden: true` to the options. The reason is that the visibility check is more expensive in most cases. People are already using this technique to improve their queries' performance by setting the default options to true globally. The downside of this technique is obviously that we'd lose the visibility check for better and predictable queries.

What if instead of removing the inaccessibility check altogether, we just have to move the order lower? Since the query is really just a set of filter functions, if we can run the faster filter first and short-circuiting most of the candidates, then hopefully we don't have to run the slower inaccessibility check on so many elements.

For a naive test with 1000 buttons all with different names in jsdom environment, selecting the last element takes around 2 seconds. After applying this changes, it only takes around 1 second in average on my machine. It's not a lot but seems like a free win to me.

```js
const {getByRole} = render(
  Array.from({length: 1000})
    .map((_, index) => `<button>${index}</button>`)
    .join(''),
)
expect(getByRole('button', {name: '999'})).not.toBeNull()
```

Obviously it doesn't reflect the real world usage, but I'd expect the usage of accessible names check are more frequent than the inaccessibility check. It'd only be slower if there are multiple elements with the same name but only a part of them are accessible/visible. I can't think of a case where this would be a common scenario in testing. Additionally, people are already opting-out of the inaccessibility check globally, this should have little impact for them.

Preferably, it'd be nice to have it tested in some real world repos to see if it actually improves the performance though.

**How**:

Move the inaccessibility check lower to the last.

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs) (I don't think this needs to be added to the documentation site. A record in changelog should be fine?)
- [ ] Tests (N/A)
- [ ] TypeScript definitions updated (N/A)
- [x] Ready to be merged

### DIFF
--- a/src/queries/role.js
+++ b/src/queries/role.js
@@ -154,13 +154,6 @@ function queryAllByRole(
       return true
     })
     .filter(element => {
-      return hidden === false
-        ? isInaccessible(element, {
-            isSubtreeInaccessible: cachedIsSubtreeInaccessible,
-          }) === false
-        : true
-    })
-    .filter(element => {
       if (name === undefined) {
         // Don't care
         return true
@@ -175,6 +168,13 @@ function queryAllByRole(
         name,
         text => text,
       )
+    })
+    .filter(element => {
+      return hidden === false
+        ? isInaccessible(element, {
+            isSubtreeInaccessible: cachedIsSubtreeInaccessible,
+          }) === false
+        : true
     })
 }
 


kevin940726:
Any plans to merge this? Or should we get more approvals from maintainers?

eps1lon:
Just wanted to let this sit so that everybody has a chance to review

timdeschryver (approved):
This LGTM. 
Although I agree with @markerikson that it could be made faster (even if it's just a little bit) if all `filter` statements are merged into 1 statement.

markerikson:
Just as an FYI, I tried specifically adding a dep on `@testing-library/dom` to bump us to 8.11, and ran our longest UI test file, an "integration"-style test using `next-page-tester`.

Before:
```
 PASS  client/features/programDetails/characteristicsTab/tests/Characteristics.tests.tsx (107.118s)
  ProgramCharacteristicsContent
    √ Renders and allows editing Program Characteristics (98432ms)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        109.2s, estimated 124s
```

After:

```
 PASS  client/features/programDetails/characteristicsTab/tests/Characteristics.tests.tsx (90.958s)
  ProgramCharacteristicsContent
    √ Renders and allows editing Program Characteristics (82187ms)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        93.06s, estimated 106s
```

No other code changes.

So, this distinctly made _some_ improvement.

PR #1069, opened 2021-10-25
Draft: feat(query-deep): Using a query selector that supports queries into the shadow dom of elements

MatthiasKainer:
**What**:

_This is the second RFC-PR for solving the shadow-dom problems. It is created as a follow up on #1054, this time showing a less generic approach - by building the shadow dom query functionality right into the framework, rather than opening it to consumers. After discussions, the goal would be to get a review for either of the two to merge it._

This commit will automatically resolve queries that have to reach into the shadow dom of any of the children to find requested elements. You can see an example in the test here: https://github.com/MatthiasKainer/dom-testing-library/commit/03e1477fcfb51633077d360d3911dc0bd8885fcb#diff-03031dc49abaa5227e8e4e13fd7459557e494345a85e4a29e7a78c10d7046cfeR87

This test would fail with the old version and works with this change. It will fix issues #742 & #413 and every other that's related to shadow dom.

**Why**:

In our project, we are using a design system built with web components (via stenciljs), and different projects in different frameworks (lit, react, vue) want to use this library. However, due to the issues already highlighted in #742 #413 and the fact that the current [workaround](https://github.com/testing-library/dom-testing-library/issues/742#issuecomment-674987855) is not really a long-term solution as it wraps an old version of this library, this change will solve this issue once and for all. 

**How**:

As said, unlike in  #1054, this change replaces the query mechanism of the core library. It uses the library [Georgegriff/query-selector-shadow-dom](https://github.com/Georgegriff/query-selector-shadow-dom) - which, to my knowledge, is also the goto solution for other testing libraries that struggle with shadow-dom with 200k downloads a week. 

The change gravitates around the following lines of code: https://github.com/MatthiasKainer/dom-testing-library/commit/03e1477fcfb51633077d360d3911dc0bd8885fcb#diff-75da2f8ae17b208e9a92d6eee57458b0a5e95f89ced8e4c400bfeb5e0d66ee85R18 

It's a wrapper for the query method, that is to be used by the application rather than using the `querySelector` and `querySelectorAll` methods directly. 

Given it changes the query mechanism this might be considered a potentially breaking change. I lack the knowledge of all the use cases of this library, but one example can already be found in the adjustments in the node tests: https://github.com/MatthiasKainer/dom-testing-library/commit/03e1477fcfb51633077d360d3911dc0bd8885fcb#diff-03031dc49abaa5227e8e4e13fd7459557e494345a85e4a29e7a78c10d7046cfeR4

Basically, non-browser environments might have issues with this change as the library will need access to the document, and to the Node interface. The reason is performance optimisation (no need to search everything if no shadow element exists) and testing if an element is a node.

**Checklist**:

- [x] Tests
- [ ] Ready to be merged
      

Pinging @eps1lon @alexkrolick to follow up on the discussions from #1054

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -17,6 +17,7 @@ module.exports = {
     ...watchPlugins,
     require.resolve('jest-watch-select-projects'),
   ],
+  transformIgnorePatterns: ['node_modules/(?!(query-selector-shadow-dom)/)'],
   projects: [
     require.resolve('./tests/jest.config.dom.js'),
     require.resolve('./tests/jest.config.node.js'),
--- a/package.json
+++ b/package.json
@@ -46,7 +46,8 @@
     "chalk": "^4.1.0",
     "dom-accessibility-api": "^0.5.9",
     "lz-string": "^1.4.4",
-    "pretty-format": "^27.0.2"
+    "pretty-format": "^27.0.2",
+    "query-selector-shadow-dom": "^1.0.0"
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^5.11.6",
--- a/src/__node_tests__/index.js
+++ b/src/__node_tests__/index.js
@@ -1,6 +1,13 @@
 import {JSDOM} from 'jsdom'
 import * as dtl from '../'
 
+beforeEach(() => {
+  const dom = new JSDOM()
+  global.document = dom.window.document
+  global.window = dom.window
+  global.Node = dom.window.Node
+})
+
 test('works without a global dom', async () => {
   const container = new JSDOM(`
     <html>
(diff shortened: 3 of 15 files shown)


alineDiesel:
👀

w0wka91:
Is this still in progress?

MatthiasKainer:
@w0wka91 
This is waiting for https://github.com/testing-library/dom-testing-library/pull/1054 - it's a breaking change, so I was proposing a stepwise approach.

alineDiesel:
Hi!
Sorry about that
We change the test approach cause access the shadow dom was impossible  :(
I’m really sad with this decision, but, the product can’t wait.

very thanks. ❤️

brownieboy:
> Hi! Sorry about that We change the test approach cause access the shadow dom was impossible :( I’m really sad with this decision, but, the product can’t wait. very thanks. ❤️

Can I ask what your new "test approach" is?

alineDiesel:
Yeap.

We are testing the form flow state with unit test and functional testing in
the critics flows.

At the first we thought of doing only integration tests…

You have another recommendation?

Aline

brownieboy:
> Yeap. We are testing the form flow state with unit test and functional testing in the critics flows. At the first we thought of doing only integration tests… You have another recommendation? Aline

Not at this stage, sadly.

PR #1073, opened 2021-11-10
fix(wait-for): Don't queue microtasks after condition is met

eps1lon:
**What**:

Fixes integration with of `@testing-library/react@alpha` in React Spectrum when using React 18.

**Why**:

Otherwise context restoration in `unstable_advanceTimersWrapper` becomes unpredictable leading to fixes like [`d375ff3` (#2526)](https://github.com/adobe/react-spectrum/pull/2526/commits/d375ff359347d7818ed4902f3e37281bb7b5d61e#diff-c19475ebc27556f4091d71e843529459da93e9ebe0179c9650978422ec3b2942L2099-R2099) that seemingly have nothing to do with the test/implemenation but magically fix tests.

The implementation change fixed 
```js
it('should handle when a non drop target element is added', async () => {
      let setShowInput2;
      let Test = () => {
        let [showInput2, _setShowInput2] = React.useState(false);
        setShowInput2 = _setShowInput2;
        return (<>
          <Draggable />
          <input />
          <Droppable />
          {showInput2 &&
            <input />
          }
        </>);
      };

      let tree = render(<Test />);

      let draggable = tree.getByText('Drag me');

      expect(tree.getAllByRole('textbox')).toHaveLength(1);

      fireEvent.focus(draggable);
      fireEvent.click(draggable);
      act(() => jest.runAllTimers());

      expect(() => tree.getAllByRole('textbox')).toThrow();

      act(() => setShowInput2(true));
      // MutationObserver is async
      await waitFor(() => expect(() => tree.getAllByRole('textbox')).toThrow());

      fireEvent.click(draggable);
      expect(tree.getAllByRole('textbox')).toHaveLength(2);
    });
```
-- https://github.com/adobe/react-spectrum/blob/8dc33dcdab3bb41518daf023c18939e844e843eb/packages/%40react-aria/dnd/test/dnd.test.js#L2072-L2105

**How**:

Exit the loop once the callback no longer throws instead of unconditionally queuing another microtask.

This may be an issue with how `@testing-library/react` uses `unstable_advanceTimersWrapper`. 
Alternatively, we could flush microtasks before we check the callback. But this was intentionall though we don't have any test/repro for that at the moment /cc @kentcdodds 

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/wait-for.js
+++ b/src/__tests__/wait-for.js
@@ -11,6 +11,17 @@ function deferred() {
   return {promise, resolve, reject}
 }
 
+let originalConfig
+beforeEach(() => {
+  originalConfig = getConfig()
+})
+
+afterEach(() => {
+  configure(originalConfig)
+  // restore timers
+  jest.useRealTimers()
+})
+
 test('waits callback to not throw an error', async () => {
   const spy = jest.fn()
   // we are using random timeout here to simulate a real-time example
@@ -138,7 +149,6 @@ Ignored nodes: comments, <script />, <style />
 })
 
 test('should delegate to config.getElementError', async () => {
-  const originalConfig = getConfig()
   const elementError = new Error('Custom element error')
   const getElementError = jest.fn().mockImplementation(() => elementError)
   configure({getElementError})
@@ -153,7 +163,6 @@ test('should delegate to config.getElementError', async () => {
 
   expect(getElementError).toBeCalledTimes(1)
   expect(error.message).toMatchInlineSnapshot(`Custom element error`)
-  configure(originalConfig)
 })
 
 test('when a promise is returned, it does not call the callback again until that promise rejects', async () => {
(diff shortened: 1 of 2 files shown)


src/__tests__/wait-for.js
```diff
@@ -255,3 +264,71 @@ test('the real timers => fake timers error shows the original stack trace when c
...
+      }
+    },
+    {interval: 50},
+  )
+
+  expect(context).toEqual('initial')
+
+  await Promise.resolve()
+
+  expect(context).toEqual('initial')
```
eps1lon, line 333: Was `no-act` without the change. We had the following schedule-resolution order:


Without this change:
1. start `asyncWrapper`
1. start `unstable_advanceTimersWrapper`
1. leave `asyncWrapper`
1. leave `unstable_advanceTimersWrapper`

With this change:
1. start `asyncWrapper`
1. start `unstable_advanceTimersWrapper`
1. leave `unstable_advanceTimersWrapper`
1. leave `asyncWrapper`

In other words, before `asyncWrapper` and `unstable_advanceTimersWrapper` were interleaved instead of nested.

kentcdodds (commented):
I didn't realize `unstable_advanceTimersWrapper` was a thing 😅 The test looks a bit complicated, but I understand why so that's fine. This looks good to me 👍

kentcdodds (approved):

eps1lon:
RE: https://github.com/testing-library/dom-testing-library/pull/667#issuecomment-965091530

Bookshelf app tests path with this PR (https://github.com/testing-library/dom-testing-library/pull/1073).
But they fail if I revert https://github.com/testing-library/dom-testing-library/pull/667 so let's stick with the approach in this PR.

Ultimately we should add a test that illustrates what's happening in the bookshelf app. Hopefully we can refactor our implementation to fix all tests but keep the code simpler.

kentcdodds:
Sounds good to me 👍

PR #1086, opened 2022-01-14
feat(ByRole): improved byRole query performance

kalmi:
**What**: byRole query logic is optimised for performance

**Why**: We experienced timeouts while using byRole queries, and decided to investigate by performing profiling. We found that each and every byRole query was taking at least hundred(s) of milliseconds in our test cases _(even when accessibility/visibility checks were disabled by mocking out getComputedStyle to always return undefined for any property)_. Our test cases have medium sized trees, nothing excessive.

**How**: the byRole query logic is refactored, so that less work done in the tight loop of byRole queries. That loop is executed for each and every node and role combination for every element for every byRole query. Some code is moved out of the loop into the surrounding closure, so that some values are precomputed. The most important change is that a cheap tagName equality check is done, and we skip most of the work if the it can't possibly be a match. These changes were done in a way that we now avoid most allocations in the tightest part of the loop.

These changes result in a 2 orders of magnitude speed up for byRole queries according to my profiling results in out test cases. _(This was measured while accessibility/visibility checks were disabled (both before and after) by mocking out getComputedStyle to always return undefined for any property.)_

**Checklist**:

- [X] Documentation added to the
- [X] Tests
- [X] TypeScript definitions updated N/A
- [X] Ready to be merged


eps1lon:
> These changes result in a 2 orders of magnitude speed up for byRole queries according to my profiling results in out test cases. (This was measured while accessibility/visibility checks were disabled (both before and after) by mocking out getComputedStyle to always return undefined for any property.)

Could you share a specific setup that was used to run these benchmarks? Don't forget to include your environment (e.g. browser versions, dependency lockfiles etc.)

src/role-helpers.js
```diff
@@ -102,27 +102,40 @@ function buildElementRoleList(elementRolesMap) {
...
-        // not using splice to not mutate the attributes array
-        attributes = [
-          ...attributes.slice(0, typeTextIndex),
-          ...attributes.slice(typeTextIndex + 1),
-        ]
-        if (node.type !== 'text') {
-          return false
-        }
+      if (upperCasedTagName && node.tagName !== upperCasedTagName) {
+        // Short-circuit if tag name does not match,
```
eps1lon, line 129: Why do we need that here? Aren't we just repeating what is implemented in `node.matches`?

eps1lon (changes requested):
Thanks. Nicely spotted!

Hoisting some of the code up so that it doesn't need to access closures definitely makes sense. Not just from a big-O notation standpoint but from a JS engine perspective as well.

I'm not so sure about the introduce short-circuits though. Could you clarify that?

src/role-helpers.js
kalmi: Hm, this is probably not as important as I initially wrote. This was my first approach that yielded results. I will recheck if this actually helps, and if it does, how much.

eps1lon: I removed the short-circuit to razor-focus this perf change

eps1lon:
@all-contributors add @kalmi for code

eps1lon:
Work towards https://github.com/testing-library/dom-testing-library/issues/698 and https://github.com/testing-library/dom-testing-library/issues/820

Issue #980, opened 2021-06-15
Error: Role "treeitem" cannot have "level" property.


lindapaiste:
### Summary

`@testing-library/jest-dom` explicitly disallows the use of the `level` property in `getByRole` for all roles except `heading`. 
 This needs to be broadened as W3 also supports this property on the  `treeitem`, `listitem`, and `row` roles.

### Relevant code or config:

-  ` "@testing-library/jest-dom": "^5.14.1"`
-  `"@testing-library/react": "^11.2.7"`

```js
test("Can select by role treeitem with level", () => {
  render(<Hello />);
  expect(
    screen.getByRole("treeitem", { name: "project-3B.docx", level: 3 })
  ).toBeInTheDocument();
  expect(
    screen.getByRole("treeitem", { name: "project-3B.docx", level: 1 })
  ).not.toBeInTheDocument();
});
```

### What you did:

I am trying to select an element with `role="treeitem"` based on its `level`. In my actual app I have the same names appearing as both a parent and a child.

### What happened:

The test cannot be executed due to an error:

>  Role "treeitem" cannot have "level" property.
>
>  at queryAllByRole (node_modules/@testing-library/dom/dist/queries/role.js:72:13)
>
>  at node_modules/@testing-library/dom/dist/query-helpers.js:87:17
>
>  at node_modules/@testing-library/dom/dist/query-helpers.js:62:17
>
>  at getByRole (node_modules/@testing-library/dom/dist/query-helpers.js:111:19)
>
>  at Object.<anonymous> (src/components/bubbles/BubblesScreen.test.tsx:24:28)

### Reproduction:

[Created a CodeSandbox](https://codesandbox.io/s/react-testing-library-aria-role-tree-vgisr?file=/src/__tests__/hello.js:332-630)

### Problem description:

The `aria-level` attribute and implicit levels are supported by W3 for multiple roles, [as seen here](https://www.w3.org/TR/wai-aria-1.2/#aria-level):
- `heading`
- `listitem`
- `row`
- `treeitem`

However `@testing-library/dom` does a validation check which only allows for `level` to be used with the `heading` role.

The problematic code appears on [lines 67-72 of `/src/queries/roles.js`](https://github.com/testing-library/dom-testing-library/blob/56a4c759a0237b6ca9dcec3cffcd71832c00e736/src/queries/role.js#L67-L72):
```
if (level !== undefined) { 
   // guard against using `level` option with any role other than `heading`
    if (role !== 'heading') {
[...]

PR #1088, opened 2022-01-15
feat(roleHelpers): compute tree implicit arialevel

MatanBobi:
**What**: Adding the option to implicitly get the `aria-level` of a `treeitem`

**Why**: Resolves https://github.com/testing-library/dom-testing-library/issues/980

**How**: At the moment, I'm using the naive approach and climb up the tree until I reach a node with explicit role `tree`.

**Checklist**:

- [X] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/role-helpers.js
+++ b/src/__tests__/role-helpers.js
@@ -3,6 +3,7 @@ import {
   logRoles,
   getImplicitAriaRoles,
   isInaccessible,
+  computeAriaLevel,
 } from '../role-helpers'
 import {render} from './helpers/test-utils'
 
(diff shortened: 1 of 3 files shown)


eps1lon:
Should I review this considering it's in draft and you said it's not ready yet?

cmorten:
> At the moment, I'm using the naive approach and climb up the tree until I reach a node with explicit role tree.

Some care needed here (if this is picked up again) as the implicit level (if I’m interpreting correctly) is not necessarily the depth in the DOM tree, but rather the depth in the accessibility tree from the parent node (including concepts such as `aria-owns`), though aria simply says "document structure" so maybe it’s ambiguous.

The core aam spec I think clarifies this: "If [aria-level](https://w3c.github.io/aria/#aria-level) is not provided or inherited for an element of [role](https://www.w3.org/TR/wai-aria/#dfn-role) [treeitem](https://w3c.github.io/aria/#treeitem) or [comment](https://w3c.github.io/aria/#comment), user agents implementing IAccessible2 or ATK/AT-SPI MUST compute it by following the explicit or computed RELATION_NODE_CHILD_OF relations."

In the majority of cases this is likely equivalent, but there are patterns where application of `role="presentation"` or `role="none"` complicates things, e.g. see https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/, or where using groups adds extra DOM hierarchy that need be ignored, e.g. see https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-1a/

> Adding the option to implicitly get the aria-level of a treeitem

Worth considering `listitem`, `row` (but only when in a `treegrid`), and `comment` (but only if want to support wai-aria 1.3 which is draft atm) for this as they too have implicit levels.

REF:
- https://www.w3.org/TR/wai-aria-1.2/#aria-level
- https://www.w3.org/TR/core-aam-1.2/#mapping_additional_position

---

By way of precedent, I have (I think) a working implementation in another project which can be used by way of example (if useful) https://github.com/guidepup/virtual-screen-reader/blob/main/src/getNodeAccessibilityData/getAccessibleAttributeLabels/getLabelFromImplicitHtmlElementValue/index.ts#L27

PR #1089, opened 2022-01-15
fix: improve error message when no window found

MatanBobi:
**What**: Improves error message when window cannot be found

**Why**: Resolves https://github.com/testing-library/dom-testing-library/issues/863

**How**: Narrow the `else` fallback

**Checklist**:

- [X] Tests
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/helpers.js
+++ b/src/__tests__/helpers.js
@@ -25,22 +25,22 @@ describe('window retrieval throws when given something other than a node', () =>
       `It looks like you passed an Array instead of a DOM node. Did you do something like \`fireEvent.click(screen.getAllBy...\` when you meant to use a \`getBy\` query \`fireEvent.click(screen.getBy...\`?`,
     )
   })
-  test('HTMLElement as node', () => {
+  test('window is not available for node', () => {
     const elem = document.createElement('div')
-    Object.defineProperty(elem, 'ownerDocument', {
+    Object.defineProperty(elem.ownerDocument, 'defaultView', {
       get: function get() {
         return null
       },
     })
 
     expect(() => getWindowFromNode(elem)).toThrowErrorMatchingInlineSnapshot(
-      `Unable to find the "window" object for the given node. Please file an issue with the code that's causing you to see this error: https://github.com/testing-library/dom-testing-library/issues/new`,
+      `It looks like the window object is not available for the provided node.`,
     )
   })
 
   test('unknown as node', () => {
     expect(() => getWindowFromNode({})).toThrowErrorMatchingInlineSnapshot(
-      `The given node is not an HTMLElement, the node type is: object`,
+      `The given node is not an Element, the node type is: object.`,
     )
   })
 })
--- a/src/__tests__/helpers.js
+++ b/src/__tests__/helpers.js
@@ -25,6 +25,19 @@ describe('window retrieval throws when given something other than a node', () =>
       `It looks like you passed an Array instead of a DOM node. Did you do something like \`fireEvent.click(screen.getAllBy...\` when you meant to use a \`getBy\` query \`fireEvent.click(screen.getBy...\`?`,
     )
   })
+  test('HTMLElement as node', () => {
+    const elem = document.createElement('div')
+    Object.defineProperty(elem, 'ownerDocument', {
+      get: function get() {
+        return null
+      },
+    })
+
+    expect(() => getWindowFromNode(elem)).toThrowErrorMatchingInlineSnapshot(
+      `Unable to find the "window" object for the given node. Please file an issue with the code that's causing you to see this error: https://github.com/testing-library/dom-testing-library/issues/new`,
+    )
+  })
+
   test('unknown as node', () => {
     expect(() => getWindowFromNode({})).toThrowErrorMatchingInlineSnapshot(
       `The given node is not an HTMLElement, the node type is: object`,
--- a/src/__tests__/helpers.js
+++ b/src/__tests__/helpers.js
@@ -27,7 +27,7 @@ describe('window retrieval throws when given something other than a node', () =>
   })
   test('unknown as node', () => {
     expect(() => getWindowFromNode({})).toThrowErrorMatchingInlineSnapshot(
-      `Unable to find the "window" object for the given node. Please file an issue with the code that's causing you to see this error: https://github.com/testing-library/dom-testing-library/issues/new`,
+      `The given node is not an HTMLElement, the node type is: object`,
     )
   })
 })
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -33,6 +33,10 @@ function getWindowFromNode(node) {
   } else if (node.window) {
     // node is window
     return node.window
+  } else if (node.ownerDocument && !node.ownerDocument.defaultValue) {
+    throw new Error(
+      `It looks like the window object is not available for the provided node.`,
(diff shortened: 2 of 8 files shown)


MatanBobi:
The validate is failing since coverage is not 100%. I tried to simulate a test when an HTMLElement is detached but I wasn't able to.
Any ideas how we can imitate that? I tried this:
```
  test.only('HTMLElement with no window', () => {
    const element = document.createElement('div')
    const removedElement = element.ownerDocument.removeChild(element)
    expect(() =>
      getWindowFromNode(removedElement),
    ).toThrowErrorMatchingInlineSnapshot()
  })
```

But it's failing saying that `element` is not a child of the `ownerDocument`.

timdeschryver (commented):
I think we can do something like this:

```ts
  test('htmlElement as node', () => {
    const elem = document.createElement('div')
    Object.defineProperty(elem, 'ownerDocument', {
      get: function get() {
        return null;
      },
    });
  
    expect(() => getWindowFromNode(elem)).toThrowErrorMatchingInlineSnapshot(
      `Unable to find the "window" object for the given node. Please file an issue with the code that's causing you to see this error: https://github.com/testing-library/dom-testing-library/issues/new`,
    )
  })
```

MatanBobi:
> I think we can do something like this:
> 
> ```ts
>   test('htmlElement as node', () => {
>     const elem = document.createElement('div')
>     Object.defineProperty(elem, 'ownerDocument', {
>       get: function get() {
>         return null;
>       },
>     });
>   
>     expect(() => getWindowFromNode(elem)).toThrowErrorMatchingInlineSnapshot(
>       `Unable to find the "window" object for the given node. Please file an issue with the code that's causing you to see this error: https://github.com/testing-library/dom-testing-library/issues/new`,
>     )
>   })
> ```

Good idea, that did the trick :)
I'm still wondering if that's even a valid use-case that can happen, but I pushed it :)
Thanks!

src/__tests__/helpers.js
```diff
@@ -25,9 +25,22 @@ describe('window retrieval throws when given something other than a node', () =>
       `It looks like you passed an Array instead of a DOM node. Did you do something like \`fireEvent.click(screen.getAllBy...\` when you meant to use a \`getBy\` query \`fireEvent.click(screen.getBy...\`?`,
     )
   })
+  test('HTMLElement as node', () => {
+    const elem = document.createElement('div')
+    Object.defineProperty(elem, 'ownerDocument', {
+      get: function get() {
+        return null
+      },
+    })
```
eps1lon, line 34: This is basically mocking `document`. I'm not sure what this particular pattern is simulating.

src/helpers.js
```diff
@@ -48,11 +48,16 @@ function getWindowFromNode(node) {
     throw new Error(
       `It looks like you passed a \`screen\` object. Did you do something like \`fireEvent.click(screen, ...\` when you meant to use a query, e.g. \`fireEvent.click(screen.getBy..., \`?`,
     )
-  } else {
+  } else if (node instanceof HTMLElement) {
```
eps1lon, line 51: I don't understand how this would ever be the case. Particularly because `instanceof` would fail for nodes from other windows.

eps1lon (changes requested):

timdeschryver (dismissed):
> I'm still wondering if that's even a valid use-case that can happen, but I pushed it :)

Me neither 😅 - but since someone opened an issue, there must be a few of those cases. Now, I hope this message will help them.

src/helpers.js
MatanBobi: Didn't think about that point.
How would you catch an instance of different windows then? And does that even a valid use-case to test for?
We can just remove this `else if` and keep the new else instead.
I wasn't sure when was the `else` part happening before, if you have an idea, please let me know :)

eps1lon: > How would you catch an instance of different windows then? And does that even a valid use-case to test for?

What would we want to catch about them?

> And does that even a valid use-case to test for?

Why would it not be valid?

MatanBobi: The wording I used was problematic. What I meant was that before this change, we assumed that if we're getting to the `else` section, we're probably in a scenario where we have an element from a different Document. How would you assert on that?

And my thought was that this use-case can't happen. How can one create an element that has no document?

src/__tests__/helpers.js
timdeschryver: It's a workaround to skip over the [`node.ownerDocument && node.ownerDocument.defaultView` condition](https://github.com/testing-library/dom-testing-library/blob/main/src/helpers.js#L30). I don't know what we could do otherwise, because the created element is attached to the `document`. 
What I've tried as alternatives: removing the created element from the document, assigning the `ownerDocument` or `defaultView` to `null`.

eps1lon: > What I've tried as alternatives: removing the created element from the document,

What does "removing from the document" mean? Note that an element always has `ownerDocument`. Whether it's connected (`element.isConnected`) or not, does not matter.

src/helpers.js
eps1lon: >  we assumed that if we're getting to the else section, we're probably in a scenario where we have an element from a different Document. How would you assert on that?

I think the wording was just ambiguous. If the node is from a different document we could still get its `Window`.

The correct assumption that was never tested for is whether we have an `Element` at all. And we can't do that without duck typing. I would just adjust the error message and say at the bottom that argument doesn't look like it's an element at all. And for the earlier check for `defaultView` we should special case if `defaultView` is `null`. Because if we have a Node (`ownerDocument` property exists) but `defaultView` is `null` the window is truly not available. Though I don't know how to create a test case for `defaultView` being `null`. So that would be a scenario where we want to mock `defaultView` (but not `ownerDocument`).

src/__tests__/helpers.js
timdeschryver: > Note that an element always has ownerDocument. Whether it's connected (element.isConnected) or not, does not matter

That's good to know, that's where I got it wrong. I assumed the issue had an issue with `render(<input />)`.

src/helpers.js
MatanBobi: That makes since, thanks @eps1lon.
I pushed another iteration, can you please have a look to see if that's what you meant?

src/helpers.js
```diff
@@ -33,6 +33,10 @@ function getWindowFromNode(node) {
   } else if (node.window) {
     // node is window
     return node.window
+  } else if (node.ownerDocument && !node.ownerDocument.defaultValue) {
```
eps1lon, line 36: Why are we checking for `defaultValue` here?

Edit: You probably meant `defaultView`. We should really check for an explicit `null` here. I'm not aware of a valid scenario where `defaultView` is `undefined`.

eps1lon (changes requested):

src/helpers.js
MatanBobi: Oops, that's my bad sorry. Fixed :)

eps1lon (approved):

MatanBobi:
@eps1lon anything preventing us from merging this?

timdeschryver (approved):
I'll also approve this PR since my review was still pending, just in case I was block this.

eps1lon:
Just cant merge from the app. Will do when I'll have access to a desktop

PR #1094, opened 2022-01-21
fix(waitForElementToBeRemoved): Make initial check work with getBy*.

### LINKED ISSUE #1093: Behavior of waitForElementToBeRemoved when first call throws
- `@testing-library/dom` version: 8.11.2

### Problem description:

[There is currently no exception catching for the initial call inside `waitForElementToBeRemoved`](https://github.com/testing-library/dom-testing-library/blob/d578c7ea285d16609a09a96081055a49997938bc/src/wait-for-element-to-be-removed.js#L30). When that call of `callback` throws, instead of getting the error `Error('Timed out in waitForElementToBeRemoved.')` we will get the error thrown from the callback, even when it is a `TestingLibraryElementError`. This contradicts [the documented behavior](https://testing-library.com/docs/dom-testing-library/api-async/#waitforelementtoberemoved):

```js
waitForElementToBeRemoved(() => getByText(/not here/i)).catch(err =>
  console.log(err),
)
// Error: The element(s) given to waitForElementToBeRemoved are already removed. waitForElementToBeRemoved requires that the element(s) exist(s) before waiting for removal.
```

### Suggested solution:

Currently writing a PR.

make-github-pseudonymous-again:
This commit includes a regression test. Fixes #1093.

### DIFF
--- a/src/wait-for-element-to-be-removed.js
+++ b/src/wait-for-element-to-be-removed.js
@@ -12,13 +12,28 @@ function initialCheck(elements) {
   }
 }
 
+function wrapFunctionCallback(callback) {
+  return () => {
+    try {
+      return callback()
+    } catch (error) {
+      if (error.name === 'TestingLibraryElementError') {
+        return null
+      }
+      throw error
+    }
+  }
+}
+
 async function waitForElementToBeRemoved(callback, options) {
   // created here so we get a nice stacktrace
   const timeoutError = new Error('Timed out in waitForElementToBeRemoved.')
-  if (typeof callback !== 'function') {
-    initialCheck(callback)
+  if (typeof callback === 'function') {
+    callback = wrapFunctionCallback(callback)
+  } else {
     const elements = Array.isArray(callback) ? callback : [callback]
     const getRemainingElements = elements.map(element => {
+      if (!element) return () => null
       let parent = element.parentElement
       if (parent === null) return () => null
       while (parent.parentElement) parent = parent.parentElement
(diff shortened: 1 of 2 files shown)


src/wait-for-element-to-be-removed.js
```diff
@@ -12,13 +12,28 @@ function initialCheck(elements) {
...
+      throw error
+    }
+  }
+}
+
 async function waitForElementToBeRemoved(callback, options) {
   // created here so we get a nice stacktrace
   const timeoutError = new Error('Timed out in waitForElementToBeRemoved.')
-  if (typeof callback !== 'function') {
-    initialCheck(callback)
```
make-github-pseudonymous-again, line 19: This use of `initialCheck` is replaced by a simpler check.

src/wait-for-element-to-be-removed.js
```diff
@@ -12,13 +12,28 @@ function initialCheck(elements) {
...
+  if (typeof callback === 'function') {
+    callback = wrapFunctionCallback(callback)
+  } else {
     const elements = Array.isArray(callback) ? callback : [callback]
     const getRemainingElements = elements.map(element => {
+      if (!element) return () => null
```
make-github-pseudonymous-again, line 36: This line replaces the removed call to `initialCheck`.

src/wait-for-element-to-be-removed.js
```diff
@@ -12,13 +12,28 @@ function initialCheck(elements) {
   }
 }
 
+function wrapFunctionCallback(callback) {
```
eps1lon, line 15: "wrap" it with what? It would help a lot if we could come up with a less generic name for this function.

src/wait-for-element-to-be-removed.js
```diff
@@ -12,13 +12,28 @@ function initialCheck(elements) {
...
+  return () => {
+    try {
+      return callback()
+    } catch (error) {
+      if (error.name === 'TestingLibraryElementError') {
+        return null
```
eps1lon, line 21: The original code return `undefined`. Is there a reason to change the return type now?

eps1lon (changes requested):
Thanks. It does look like the new behavior is intended. 

Though the implementation needs some work.

PR #1102, opened 2022-02-03
refactor: Leverage early return

afontcu:
This simple PR leverages the early return pattern (which is already heavily used across the codebase).


MatanBobi (approved):
Looks good :)

PR #1116, opened 2022-03-22
fix: reset regexg to prevent regexg with global flags to not match every element

### LINKED ISSUE #1115: getAllByRole witha regex matcher is returning the wrong count
- `@testing-library/dom` version: 8.11.3
- Testing Framework and version: Jest 26.6.3
  
- DOM Environment: jsdom 16.6.0
  

### Relevant code or config:

```js
import { getQueriesForElement } from '@testing-library/dom';

test('should use get all by role matcher', () => {
  const container = document.createElement('div');
  container.innerHTML = `<button
    type="button"
    title="Low">Low</button>
  <button
    type="button"
    title="Medium">Medium</button>
  <button
    type="button"
    title="High">High</button>`;
  const { getAllByRole } = getQueriesForElement(container);
  expect(getAllByRole('button', { name: /(low|medium|high)/gi }).length).toBe(3); // error - only found 2
});
```

### What you did:

I was trying to match a set of elements using the getAllByRole matcher within React Testing Library, and the length of the matched results was coming back wrong.

### What happened:

Instead of finding all 3 buttons, it only found 2 - the Low and High buttons

### Reproduction:

The code above is a self-contained minimal reproduction of this issue within DOM Testing Library (we found this with RTL).

### Problem description:

All 3 buttons should be found, rather than only finding 2.

### Suggested solution:

I wasn't able to dig too far into the code for this. I imagine something is happening either in the filters for the queryAllByRole method, or within the matches method.

timdeschryver:
**What**: Resolves #1115

**Why**: 

The instance of a regex that's using the global flag hold a `lastIndex` state.
This might give false positive results.

For example:

```
const g = /foo/g
g.test('foo') // |> true
g.test('foo') // |> false because lastIndex is 3
g.test('foo') // |> true because lastIndex is 0 again
g.test('foo') // |> false because lastIndex is 3
```

**How**:

Added a reproduction test case.
Recreate a new regex instance from the original regex.

An alternative would be to reset the `lastIndex` to `0` after the check.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/matches.js
+++ b/src/__tests__/matches.js
@@ -15,6 +15,22 @@ test('matchers accept regex', () => {
   expect(fuzzyMatches('ABC', node, /ABC/, normalizer)).toBe(true)
 })
 
+// https://stackoverflow.com/questions/1520800/why-does-a-regexp-with-global-flag-give-wrong-results
+test('a regex with the global flag consistently (re-)finds a match', () => {
+  const regex = /ABC/g
+  const spy = jest.spyOn(console, 'warn').mockImplementation()
+
+  expect(matches('ABC', node, regex, normalizer)).toBe(true)
+  expect(fuzzyMatches('ABC', node, regex, normalizer)).toBe(true)
+
+  expect(spy).toBeCalledTimes(2)
+  expect(spy).toHaveBeenCalledWith(
+    `To match all elements we had to reset the lastIndex of the RegExp because the global flag is enabled. We encourage to remove the global flag from the RegExp.`,
+  )
+
+  console.warn.mockClear()
+})
+
 test('matchers accept functions', () => {
   expect(matches('ABC', node, text => text === 'ABC', normalizer)).toBe(true)
   expect(fuzzyMatches('ABC', node, text => text === 'ABC', normalizer)).toBe(
--- a/src/matches.ts
+++ b/src/matches.ts
@@ -36,7 +36,7 @@ function fuzzyMatches(
   } else if (typeof matcher === 'function') {
     return matcher(normalizedText, node)
   } else {
-    return matcher.test(normalizedText)
+    return matchRegExp(matcher, normalizedText)
   }
 }
 
@@ -56,7 +56,7 @@ function matches(
   if (matcher instanceof Function) {
     return matcher(normalizedText, node)
   } else if (matcher instanceof RegExp) {
-    return matcher.test(normalizedText)
+    return matchRegExp(matcher, normalizedText)
   } else {
     return normalizedText === String(matcher)
   }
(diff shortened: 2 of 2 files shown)


src/matches.ts
```diff
@@ -56,7 +56,7 @@ function matches(
   if (matcher instanceof Function) {
     return matcher(normalizedText, node)
   } else if (matcher instanceof RegExp) {
-    return matcher.test(normalizedText)
+    return new RegExp(matcher).test(normalizedText)
```
eps1lon, line 59: does this preserve all flags? I didn't look at our existing tests so if we already have tests for all flags then this is good to go.

timdeschryver: Good catch, the tests on `matches` don't have any flags, but indirectly a few flags are used throughout the test suite. I'll provide more test cases for `matches`.

eps1lon: Honestly I would just throw on regexp with a global flag (if we can detect that). It doesn't make sense to pass anyway and complicates things for us needlessly

timdeschryver: We can detect if it has a global flag.

Throwing would result in a new major version. Even a fix like the proposed one, or another one, would also lead to a new major because the behaviour will change. 

The more I think abou it, doing nothing and keeping the current implementation might be the best here. We could add a note to the docs. Using a global flag is rare and it doesn't do anything for these cases.

eps1lon: > Throwing would result in a new major version.

I don't see it that way. The old behavior with global flags was clearly wrong anyway. We can do `console.warn` though to make it clearer. 

> Using a global flag is rare and it doesn't do anything for these cases.

Note that a global flag does something as is apparent by the issue. It returns the wrong number of results which is a bug.

timdeschryver: I agree that it's wrong, but I can imagine at there are a few global regexes to just find a single element. This currently works, but would result that it throws in a new version. 

After looking at the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#instance_properties), `lastIndex` is only used for globals, so that would be a safe fix.

I would prefer that over throwing, `console.warn` is also fine for me - this could also become stricter (as in an error) in the future when there are more changes to create a new major.


> Note that a global flag does something as is apparent by the issue. It returns the wrong number of results which is a bug.

True, I formulated the sentence badly. What I meant was that to find an element it doesn't matter if a global flag is added to the regex. 


I've updated the commit with the `lastIndex` implementation, if you prefer the `console.warn` let me know and I'll change it. Thanks for your help.

src/__tests__/matches.js
```diff
@@ -15,6 +15,14 @@ test('matchers accept regex', () => {
   expect(fuzzyMatches('ABC', node, /ABC/, normalizer)).toBe(true)
 })
 
+// https://stackoverflow.com/questions/1520800/why-does-a-regexp-with-global-flag-give-wrong-results
+test('matchers recreate regex to prevent global mistakes', () => {
```
eps1lon, line 19: It's not really about re-creating regexps. That's an implementation detail which is evident by the fact that we no longer re-create but mutate.

eps1lon (changes requested):
This is fine but let's make sure we have a good follow-up strategy in case this doesn't work

eps1lon (approved):
Looks nice, thanks!

Blame me if somebody complains about noisy warnings added in a patch release 😄

eps1lon:
@timdeschryver  I think you need to push an empty commit or rebase and drop the merge commit. Codecov may not like merge commits which is a result of updating the branch from the GitHub UI.

timdeschryver:
@eps1lon Done.
No problem, it was a team effort so we have me to blame aswell 
This change had a bit more catches than forseen 😅

eps1lon (approved):

PR #1120, opened 2022-04-01
feat(byRole): Add description filter

PaquitoSoft:
**What**:

Add a new `description` option to `ByRole` queries.
This helps finding elements which are aria described.

**Why**:

We were implementing a React component to display the usual popup notifications and, when reading about accessibility considerations, we found we should use `ariadialog` role and we also should use `aria-describedby` to reference the message ([spec documentation](https://www.w3.org/TR/wai-aria-1.1/#alertdialog)).

When writing some tests we wanted to have one where we could close an specific notification when a list of them were rendered in a view.

`ByRole` queries currently support filtering results by accessible name, but no filtering by **accessible description**.

**How**:

Since this project is already using the [dom-accessibility-api](https://github.com/eps1lon/dom-accessibility-api) dependency to resolve the accessible name and that library also expose a function to resolve the accessible description, I would like to introduce this proposal where we implement **filtering by description** when using `ByRole` queries.

Given some HTML markup like this one:
```
<ul>
  <li role="alertdialog" aria-describedby="notification-id-1">
    <div><button>Close</button></div>
    <div id="notification-id-1">You have unread emails</div>
  </li>
  <li role="alertdialog" aria-describedby="notification-id-2">
    <div><button>Close</button></div>
    <div id="notification-id-2">Your session is about to expire!</div>
  </li>
</ul>
```

Now we can find `alertdialog` elements with this kind of query:
```
const notification = getByRole('alertdialog', {
  description: 'Your session is about to expire!',
});
```

I used [this PR](https://github.com/testing-library/dom-testing-library/pull/408) (when `name`filtering was added to `ByRole` queries) to guide my proposed implementation.

I just wanted to create the proposal with some actual code so maybe it's easier to reason about and maybe there are other thing to take into account that we might discuss as a follow-up of this starting point.

**Checklist**:

- [x] Tests
- [ ] Ready to be merged
      

**References**
These are the documents I used as reference for this proposal:
* [RTL docs](https://testing-library.com/docs/queries/byrole)
* [`alertdialog` spec](https://www.w3.org/TR/wai-aria-1.1/#alertdialog)
* [What is an accessible name?](https://www.tpgi.com/what-is-an-accessible-name/)
* [Accessible Name and Description spec](https://www.w3.org/TR/accname-1.1/)

### DIFF
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -572,3 +572,177 @@ test('should find the input using type property instead of attribute', () => {
   const {getByRole} = render('<input type="124">')
   expect(getByRole('textbox')).not.toBeNull()
 })
+
+test('can be filtered by accessible description', () => {
+  const targetedNotificationMessage = 'Your session is about to expire!'
+  const {getByRole} = renderIntoDocument(
+    `
+<ul>
+  <li role="alertdialog" aria-describedby="notification-id-1">
+    <div><button>Close</button></div>
+    <div id="notification-id-1">You have unread emails</div>
+  </li>
+  <li role="alertdialog" aria-describedby="notification-id-2">
+    <div><button>Close</button></div>
+    <div id="notification-id-2">${targetedNotificationMessage}</div>
+  </li>
+</ul>`,
+  )
+
+  const notification = getByRole('alertdialog', {
+    description: targetedNotificationMessage,
+  })
+
+  expect(notification).not.toBeNull()
+  expect(notification).toHaveTextContent(targetedNotificationMessage)
+
+  expect(
+    getQueriesForElement(notification).getByRole('button', {name: 'Close'}),
+  ).not.toBeNull()
+})
+
+test('error should include description when filtering a no results are found', () => {
+  const targetedNotificationMessage = 'Your session is about to expire!'
+  const {getByRole} = renderIntoDocument(
+    `<div role="dialog" aria-describedby="some-id"><div><button>Close</button></div><div id="some-id">${targetedNotificationMessage}</div></div>`,
+  )
+
+  expect(() =>
+    getByRole('alertdialog', {description: targetedNotificationMessage}),
+  ).toThrowErrorMatchingInlineSnapshot(`
+    Unable to find an accessible element with the role "alertdialog" and description "Your session is about to expire!"
+
+    Here are the accessible roles:
+
+      dialog:
+
+      Name "":
+      Description "Your session is about to expire!":
+      <div
+        aria-describedby="some-id"
+        role="dialog"
+      />
+
+      --------------------------------------------------
+      button:
+
+      Name "Close":
+      Description "":
+      <button />
+
+      --------------------------------------------------
+
+    Ignored nodes: comments, <script />, <style />
+    <body>
+      <div
+        aria-describedby="some-id"
+        role="dialog"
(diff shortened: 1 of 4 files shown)


PaquitoSoft:
> Very nice, thanks!
> 
> We need documentation in https://github.com/testing-library/testing-library-docs/edit/main/docs/queries/byrole.mdx (under "Options") for this to be mergable.

Hey @eps1lon you mention to add the new option under the _Options_ section but I noticed the `name` property is not there (it is explained before that section) so [I added on that](https://github.com/testing-library/testing-library-docs/pull/1035).

Should I move that explanation to the _Options_ section?
Should another PR (maybe out of current's scope) do the same for `name` prop?

Thanks!

eps1lon:
> Hey @eps1lon, you mention to add the new option under the Options section but I noticed the name property is not there (it is explained before that section) so https://github.com/testing-library/testing-library-docs/pull/1035.

We should move the `name` section to `options` as well. But let's do that in a follow-up

src/role-helpers.js
```diff
@@ -191,7 +194,20 @@ function prettyRoles(dom, {hidden}) {
             computedStyleSupportsPseudoElements:
               getConfig().computedStyleSupportsPseudoElements,
           })}":\n`
+
           const domString = prettyDOM(el.cloneNode(false))
+
+          if (includeDescription) {
```
timdeschryver, line 200: @eps1lon not related to this PR, but should we also do a `includeName`?
It seems like we're already providing the property [here](https://github.com/testing-library/dom-testing-library/blob/main/src/queries/role.js#L229) but we're not doing anything with it.

src/__tests__/role.js
```diff
@@ -572,3 +572,177 @@ test('should find the input using type property instead of attribute', () => {
...
+
+  expect(notification).not.toBeNull()
+  expect(notification).toHaveTextContent(targetedNotificationMessage)
+
+  expect(
+    getQueriesForElement(notification).getByRole('button', {name: 'Close'}),
+  ).not.toBeNull()
+})
+
+test('error should include description when filtering a no results are found', () => {
```
timdeschryver, line 604:
```suggestion
test('error should include description when filtering and no results are found', () => {
```

src/role-helpers.js
eps1lon: I think we originally decide to always include it to help people discover the `name` option (would have to check the original PR though).

Though I don't think this necessarily applies to `description` since it's
1. more likely to not be set
2. quite noisy (concerning length) if you don't care about it

timdeschryver: I'll take a look at the history.
If we indeed decided to always include it (which makes sense imho), I'll remove the code where we now add the `includeName` argument.

PaquitoSoft:
Thanks for accepting this little feature and for your help 👍

vctormb:
I had some tests like this:
```ts
<button title="this is the title">click me</button>

screen.getByRole('button', { name: /this is the title/i }) // <- before it was getting by its title
```

now it only gets by its children
```ts
<button title="this is the title">click me</button>

screen.getByRole('button', { name: /this is the title/i }) // <- failing
screen.getByRole('button', { name: /click me/i }) // <- OK
```

is this problem related to this change?

thanks!

PR #1138, opened 2022-06-15
feat: add `defaultIgnore` config option

robin-drexler:
**What**:

Adds `defaultIgnore` config option for `byText` in a similar vein to what already exists with `defaultHidden` for `byRole`.

**Why**:
Allows to set a default `ignore` globally for a project.

**How**:
Adds new config option. Also uses potentially overridden config value in places where the the default const was used before as this seemed to make sense to me.

Is this a change you'd be willing to accept?
If yes, I'd be happy to add docs etc. Thanks! :)

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs): https://github.com/testing-library/testing-library-docs/pull/1088#pullrequestreview-1011701676
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/types/config.d.ts
+++ b/types/config.d.ts
@@ -12,6 +12,8 @@ export interface Config {
   asyncUtilTimeout: number
   computedStyleSupportsPseudoElements: boolean
   defaultHidden: boolean
+  /** default value for the `ignore` option in `ByText` queries */
+  defaultIgnore: string
   showOriginalStackTrace: boolean
   throwSuggestions: boolean
   getElementError: (message: string | null, container: Element) => Error
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -1080,6 +1080,18 @@ test('getByText ignores script tags by default', () => {
   expect(getAllByText(/hello/i, {ignore: false})).toHaveLength(3)
 })
 
+test('the default value for `ignore` can be configured', () => {
+  configure({defaultIgnore: 'style'})
+
+  const {getAllByText} = render(
+    '<script>Hello</script><div>Hello</div><style>.Hello{}</style>',
+  )
+  const noStyle = getAllByText(/hello/i)
+  expect(noStyle).toHaveLength(2)
+  expect(noStyle[0].tagName).toBe('SCRIPT')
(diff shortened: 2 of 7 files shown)


eps1lon (dismissed):
Thanks. This was planned as a follow-up but I didn't have time.

Could you make a PR to the docs as well?

types/config.d.ts
```diff
@@ -12,6 +12,7 @@ export interface Config {
   asyncUtilTimeout: number
   computedStyleSupportsPseudoElements: boolean
   defaultHidden: boolean
+  defaultIgnore: string
```
timdeschryver, line 15: Could be in a follow-up, but I think it would be helpful if we add some JS-Docs to the types.
Currently the text `default value for the `ignore` option in `ByText` queries` isn't visible to the users.

timdeschryver (approved):

robin-drexler:
> Thanks. This was planned as a follow-up but I didn't have time.
> 
> Could you make a PR to the docs as well?

Opened the PR here: https://github.com/testing-library/testing-library-docs/pull/1088

not sure about how to release and in what order. 😅 

Is that something the maintainers will do? 

Thanks again!

types/config.d.ts
robin-drexler: Added a comment here: https://github.com/testing-library/dom-testing-library/pull/1138/files#diff-19584312cba053f9726eeb4b0766364b873cd974e552656c557ae30ee45f9d00R15

Is that what you were looking for?

timdeschryver:
@robin-drexler You don't have to worry about that. Thanks in advance to create both PRs!

> Is that something the maintainers will do?

robin-drexler:
> @robin-drexler You don't have to worry about that. Thanks in advance to create both PRs!
> 
> > Is that something the maintainers will do?

Perfect, thanks a lot!
Looking forward to seeing it land. :D 

Let me know if I can do anything to support.

eps1lon:
> not sure about how to release and in what order. sweat_smile

I just click two buttons and change the order every so often 😅

eps1lon (approved):

eps1lon:
@all-contributors add @robin-drexler for code

PR #1141, opened 2022-06-22
fix: log ignored selectors correctly on error

robin-drexler:
**What**:

default `getElementError` implementation always logged that script and style are being ignored even when the default has been overridden with the `defaultIgnore` option.

It now logs the correct selector.
Since the selector could be any css selector, we can't just print `<script />` etc. Instead we just print the selector as is.

**Checklist**:

- [x] Tests
- [ ] Ready to be merged


src/__tests__/element-queries.js
```diff
@@ -1,6 +1,16 @@
...
 import {render, renderIntoDocument} from './helpers/test-utils'
 
+// set original config
+let originalConfig
+beforeAll(() => {
+  originalConfig = getConfig()
+})
+
+afterEach(() => {
+  configure(originalConfig)
```
robin-drexler, line 11: Ensure tests don't rely on previous tests changing the config in a certain way.

src/__tests__/element-queries.js
```diff
@@ -1092,6 +1102,19 @@ test('the default value for `ignore` can be configured', () => {
   expect(noStyle[1].tagName).toBe('DIV')
 })
 
+test('the default value for `ignore` is used in errors', () => {
```
robin-drexler, line 1105: New test that proves that overriding `defaultIgnore` produces a new error message

src/__tests__/wait-for.js
```diff
@@ -194,14 +194,14 @@ test('when a promise is returned, if that is not resolved within the timeout, th
   await sleep(5)
 
   expect((await waitForError).message).toMatchInlineSnapshot(`
-Timed out in waitFor.
-
-Ignored nodes: comments, <script />, <style />
```
eps1lon, line 199: Can we keep the original formatting? "comments, scripts, style" looks like they all refer to the same thing while "scripts, style" specifically refers to elements.

eps1lon: Should be fine. Alternative is another `.split().map().join()` which is probably too expensive.

eps1lon (approved):
Thanks!

eps1lon:
@all-contributors add @robin-drexler for code

PR #1144, opened 2022-07-04
feat: add a return value from logTestingPlaygroundURL

### LINKED ISSUE #1125: [feature] add an option to return value from `logTestingPlaygroundURL` instead of printing to console
`logTestingPlaygroundURL` is such a powerful feature 💪!
The only issue with it is always printed to the console if you're using test runners such as (wallaby.js) this is extremely inconvenient.

it'd be great to have a similar `getTestingPlaygroundURL` method or add Options parameters to the original method.

### Suggested implementation:

would you be open for a PR?

[Test](https://github.com/testing-library/dom-testing-library/blob/f27d14c18c02c412635d33fdaa1cfdaa44c14a5a/src/__tests__/screen.js#L43)

once this is implemented we will be able to work faster with the value both in VSCode and other IDE's

Lirlev48:
Adding a return value to logTestingPlaygroundURL function besides the log - Resolves #1125

logTestingPlaygroundURL() used to prints the value directly to the stdout of the running process, 
The values was hidden in the log window

Adding a return statement to the logTestingPlaygroundURL function

**Checklist**:

- [X] Tests
- [X] TypeScript definitions updated
- [X] Ready to be merged
      

Will add documentation to testing-library-docs after the PR will be completed.

### DIFF
--- a/src/screen.ts
+++ b/src/screen.ts
@@ -42,6 +42,7 @@ const logTestingPlaygroundURL = (element = getDocument().body) => {
   console.log(
     `Open this URL in your browser\n\n${getPlaygroundUrl(element.innerHTML)}`,
   )
+  return getPlaygroundUrl(element.innerHTML)
 }
 
 const initialValue = {debug, logTestingPlaygroundURL}
--- a/src/screen.ts
+++ b/src/screen.ts
@@ -39,10 +39,9 @@ const logTestingPlaygroundURL = (element = getDocument().body) => {
     console.log(`The provided element doesn't have any children.`)
     return
   }
-  console.log(
-    `Open this URL in your browser\n\n${getPlaygroundUrl(element.innerHTML)}`,
-  )
-  return getPlaygroundUrl(element.innerHTML)
+  const playgroundUrl = getPlaygroundUrl(element.innerHTML)
+  console.log(`Open this URL in your browser\n\n${playgroundUrl}`)
+  return playgroundUrl
 }
 
 const initialValue = {debug, logTestingPlaygroundURL}
--- a/types/screen.d.ts
+++ b/types/screen.d.ts
@@ -13,7 +13,7 @@ export type Screen<Q extends Queries = typeof queries> = BoundFunctions<Q> & {
     options?: OptionsReceived,
   ) => void
   /**
-   * Convenience function for `Testing Playground` which logs URL that
+   * Convenience function for `Testing Playground` which logs & return URL that
    * can be opened in a browser
    */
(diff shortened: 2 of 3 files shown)


types/screen.d.ts
```diff
@@ -13,7 +13,7 @@ export type Screen<Q extends Queries = typeof queries> = BoundFunctions<Q> & {
     options?: OptionsReceived,
   ) => void
   /**
-   * Convenience function for `Testing Playground` which logs URL that
+   * Convenience function for `Testing Playground` which logs & return URL that
    * can be opened in a browser
    */
   logTestingPlaygroundURL: (element?: Element | HTMLDocument) => void
```
timdeschryver, line 19: Because it returns the URL.

```suggestion
  logTestingPlaygroundURL: (element?: Element | HTMLDocument) => string
```

types/screen.d.ts
```diff
@@ -13,7 +13,7 @@ export type Screen<Q extends Queries = typeof queries> = BoundFunctions<Q> & {
     options?: OptionsReceived,
   ) => void
   /**
-   * Convenience function for `Testing Playground` which logs URL that
+   * Convenience function for `Testing Playground` which logs & return URL that
```
timdeschryver, line 16:
```suggestion
   * Convenience function for `Testing Playground` which logs and returns the URL that
```

Lirlev48:
@timdeschryver  can you please approve? thanks

timdeschryver (dismissed):
LGTM, leaving it to @MatanBobi

src/screen.ts
```diff
@@ -42,6 +42,7 @@ const logTestingPlaygroundURL = (element = getDocument().body) => {
   console.log(
     `Open this URL in your browser\n\n${getPlaygroundUrl(element.innerHTML)}`,
   )
+  return getPlaygroundUrl(element.innerHTML)
```
MatanBobi, line 45: Small nit, can we save the `getPlaygroundUrl in a const instead of calling the function twice? Not sure what the performance aspects for it, but better safe than sorry :)

MatanBobi (changes requested):
Thanks @Lirlev48 for the contribution :) Left a small comment.

@timdeschryver thanks for reviewing this one :)
Do you see any issue with this? I'm not sure if this change can somehow break something, wdyt?

timdeschryver:
I think this is safe @MatanBobi .

MatanBobi (approved):
Thanks @Lirlev48 :)

MatanBobi:
@all-contributors please add @Lirlev48 for code :)

Issue #498, opened 2020-03-24
Should we add a script load event ? 


JulienKode:
Thank you for this nice library 

Currently it's possible to trigger a `load` event on a image:
Example:
```javascript
const { getByTestId, debug } = render(<Stuff/>);

fireEvent.load(getByTestId('my-stuff'));

expect(stuff).toHaveBeenCalledWith(true);
```

But is there any way to do that with `script` ?
Should we implement it here https://github.com/testing-library/dom-testing-library/blob/6c4486c5c290aac515f103314428a9ab2e2873fb/src/__tests__/events.js#L109

kentcdodds:
What would it take to support this? It's it not already supported?

gertsonderby:
It works for me with this test:
```js
test('can set load event handler', () => {
    const handler = jest.fn();
    // This component creates a script with a usdeEffect hook
    render(<TestComponent options={{ onLoad: handler }} src="https://example.com/test.js" />);
    const script = document.head.querySelector('script')!;
    expect(handler).not.toHaveBeenCalled();
    fireEvent.load(script);
    return waitFor(() => expect(handler).toHaveBeenCalled());
});
```

gertsonderby:
I'm digging a little. This test fails (using react testing lib):
```js
test('can load a script', () => {
    const handler = jest.fn();
    render(<script onLoad={handler} src="https://example.com/test.js" />);
    const script = document.querySelector('script');
    expect(handler).not.toHaveBeenCalled();
    fireEvent.load(script);
    expect(handler).toHaveBeenCalled(); // Was not called.
});
```
But this one succeeds:
```js
test('can load a script', () => {
    const handler = jest.fn();
    render(<script src="https://example.com/test.js" />);
    const script = document.querySelector('script');
    script.onload = handler;
    expect(handler).not.toHaveBeenCalled();
    fireEvent.load(script);
    expect(handler).toHaveBeenCalled();
});
```
As you can see, the only difference is that the handler is set indirectly. So there may be an issue wrt. setting the handler?

JulienKode:
@gertsonderby I didn't notice that if I set the `onload` directly in the script it work

kentcdodds:
Sounds like this may be an issue with React's support of `onLoad`?

gertsonderby:
Possibly. Although I don't think they're all too interested in supporting script tags like that, it seems out of scope.

Personally, I would add them to `<head>` with a`useEffect` hook instead of rendering them directly.

[...]

PR #1147, opened 2022-07-09
fix: load and error dispatch `Event` instead of `UIEvent`

eps1lon:
**What**:

`fireEvent.load` (et. al.)  dispatches a `Event` now instead of `UIEvent`
Adds a test for https://github.com/testing-library/dom-testing-library/issues/498

**Why**:

I imagine that the most common targets are `` and `<script />` and neither of these receive load events as `UIEvent`: https://codesandbox.io/s/react-script-does-not-work-the-way-you-think-it-does-stc66c

**How**:

Changed event type and added tests for `<script />`. Ieft a note that our abstraction doesn't support firing events with different types.

Also  removed the notion of "Image events" from comments. We refer to DOM interfaces in all other cases. But there is no such thing as `ImageEvent`.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/events.js
+++ b/src/__tests__/events.js
@@ -72,6 +72,16 @@ const eventTypes = [
     elementType: 'div',
   },
   {
+    type: '',
+    events: ['load', 'error'],
+    elementType: 'img',
+  },
+  {
+    type: '',
+    events: ['load', 'error'],
+    elementType: 'script',
+  },
+  {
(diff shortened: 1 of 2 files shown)


MatanBobi (approved):
This looks good to me. Was there an issue about this somewhere? I see nothing linked.

PR #1156, opened 2022-08-09
chore: Convert `helpers.js` to TypeScript

eps1lon:
**What**:

Part of #494

**Why**:

To use generated TypeScript declaration files eventually

**How**:

Fairly low level helpers where most types are `unknown`. Uses `any` for pragmatic reasons.

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,9 @@
       "plugin:import/typescript"
     ],
     "rules": {
+      "@typescript-eslint/prefer-optional-chain": "off",
+      "@typescript-eslint/no-explicit-any": "off",
+      "@typescript-eslint/no-unsafe-member-access": "off",
       "@typescript-eslint/prefer-includes": "off",
       "import/prefer-default-export": "off",
       "import/no-unassigned-import": "off",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -65,6 +65,9 @@
       "plugin:import/typescript"
     ],
     "rules": {
+      "@typescript-eslint/prefer-optional-chain": "off",
```
eps1lon, line 68: Stylistic rule whose rational is not suitable for a lint rule (false-positive rate needs to be next to zero).

package.json
```diff
@@ -65,6 +65,9 @@
...
+      "@typescript-eslint/no-explicit-any": "off",
+      "@typescript-eslint/no-unsafe-member-access": "off",
```
eps1lon, line 70: Once this rule can distiginguish between explicit `any` and implicit/external `any` we can disable. Until then its false-positive rate is too high.

timdeschryver (approved):
👍 the turned off eslint rules also makes sense to me

Issue #846, opened 2020-12-10
*ByRole reads hidden element names as ""


tpict:
- `@testing-library/dom` version: 7.28.1
- Testing Framework and version: Jest 26.6.3
- DOM Environment: jsdom 16.4.0

### Relevant code or config:

```js
import { render } from "@testing-library/react";
import { getByRole } from "@testing-library/dom";

const { container } = render(
  <>
    <input aria-label="Input" style={{ display: "none" }} />
    <button aria-label="Button" style={{ display: "none" }} />
  </>,
);

expect(
  getByRole(container, "textbox", { name: "Input", hidden: true }),
).toBeInTheDocument();
expect(
  getByRole(container, "button", { name: "Button", hidden: true }),
).toBeInTheDocument();

```

### What you did:

Attempted to test the presence of accessible elements with `display: none;`

### What happened:

The Testing Library error reporter claims the name of these elements is `""` and fails to match them against the provided name:

```
    TestingLibraryElementError: Unable to find an element with the role "textbox" and name "Input"

    Here are the available roles:

      textbox:

      Name "":
      <input
        aria-label="Input"
        style="display: none;"
      />

      --------------------------------------------------button:

      Name "":
      <button
        aria-label="Button"
        style="display: none;"
      />

      --------------------------------------------------
```

### Reproduction:

Hopefully the provided code is sufficient, let me know if not

### Problem description:

The example I've given is a little simplified–the real code I'm testing contains elements that are visible in desktop-width viewports, and if I understand correctly media query parsing is out-of-scope for Testing Library–but the issue is the same: these elements will become part of the accessibility tree under certain circumstances and I'm ignoring their invisibility in order to check that they have the correct accessible names. I would expect that I'd be able to do that with the `hidden: true` option set.

### Suggested solution:

I don't have the time to spare to figure out the cause of this right now :( it might be related to recent changes around https://github.com/testing-library/dom-testing-library/pull/804 ?

eps1lon:
I lean towards leaving this as undefined behavior. Since the accessible name is affected by styling, ignoring it leaves you in an awkward spot. It makes sense in this specific case but the code is going to be really hairy getting all the edge cases right if the element isn't actually part of the a11y tree. For example what happens if  you use `aria-labelledby`, pass `hidden: true` but the referenced element by `aria-labelledby` is also hidden?

>  these elements will become part of the accessibility tree under certain circumstances 

[...]

PR #1161, opened 2022-08-31
fix(ByRole): filter by name or description when hidden

viniciuslagedo:
**What**: Fixes the ByRole methods to accept being filtered using name and description. Also fixes the logs to show the accessible name in case of not found error. Related to https://github.com/testing-library/dom-testing-library/issues/846

**Why**: If we have two hidden elements with the same role we can't filter that using the name. Also when we have hidden elements on logs they don't have the name filled, always equal `""`

**How**: By adding the hidden option when use `computeAccessibleDescription` and `computeAccessibleName` to get the accessible description and name

**Checklist**:

- [x] Tests
- [x] Ready to be merged


src/__tests__/role.js
```diff
@@ -26,7 +26,7 @@ test('by default logs accessible roles when it fails', () => {
 })
 
 test('when hidden: true logs available roles when it fails', () => {
-  const {getByRole} = render(`<div hidden><h1>Hi</h1></div>`)
+  const {getByRole} = render(`<h1 hidden>Hi</h1>`)
```
eps1lon, line 29: Let's keep this test. I don't think it's relevant to the change. If there is something important here, lets's add a new test instead so that we can better track what changed

viniciuslagedo: The idea was to test the case where an element had a name available to be filtered but is hidden, before that change the name was always empty. May we can add a test specifically for that case? What do you think?

eps1lon: Always add test instead of re-purposing existing tests. Existing tests should only change their assertions if they relied on buggy behavior.

viniciuslagedo: Thank you! I'm learning about all that tests concepts and patterns.

viniciuslagedo: @eps1lon can you check this again? :3

eps1lon:
The problem is that the accessible name is different if the element is hidden. See https://github.com/testing-library/dom-testing-library/issues/846#issuecomment-742156607

So this adds a footgun where you just flip `hidden` and expect `name` to return the same set. I wasn't convinced by https://github.com/testing-library/dom-testing-library/issues/846 fully to implement this so let's add some concrete, high level use cases to https://github.com/testing-library/dom-testing-library/issues/846 first (e.g. what are you testing?).

Issue #1169, opened 2022-09-19
Nullish coalescing operator (??) in dist in 8.18.0


leroydev:
@testing-library/dom@8.18.0 has a nullish coalescing operator in the distributed build in the file `dist/@testing-library/dom.esm.js`.
This results in only Node.js 14 and up being able to execute this code and in errors in Cypress 9.1.0:
![image](https://user-images.githubusercontent.com/2467392/190972987-0b09d383-cb60-4e81-aa42-5ab8c1761087.png)

Check here, 8.17.1, lines 409 to 413: https://unpkg.com/browse/@testing-library/dom@8.17.1/dist/@testing-library/dom.esm.js

And in 8.18.0, lines 428 to 430: https://unpkg.com/browse/@testing-library/dom@8.18.0/dist/@testing-library/dom.esm.js

rmoralp:
It seems to be related to last [kcd-scripts](https://github.com/kentcdodds/kcd-scripts) updates.

Could @kentcdodds check it?

sneridagh:
@kentcdodds This is quite important, probably breaking the build of a lot of people...

MichaelDeBoey:
@rmoralp @sneridagh It can't be because the latest changes in `kcd-scripts` as `@testing-library/dom` is still using v11 & `kcd-scripts` is already at v12.3

https://github.com/testing-library/dom-testing-library/blob/ab8182cfc5164f5d0b0fea143cfde350369488c9/package.json#L59

sneridagh:
It's because of Cypress... no kcd-scripts involved...

MichaelDeBoey:
Since I can't assign this ticket to a team, I'll ping @testing-library/dom to look into this

eps1lon:
I'm on it. Will pin kcd-scripts until. Future updates to tooling need a more thorough audit

sneridagh:
@MichaelDeBoey @eps1lon The culprit is in `@testing-library/cypress` since it has `@testing-library/dom` as a dependency with caret:

https://github.com/testing-library/cypress-testing-library/blob/main/package.json#L44

MichaelDeBoey:
@eps1lon Pinning `kcd-scripts` won't help, as we're not using latest version yet (as I described in https://github.com/testing-library/dom-testing-library/issues/1169#issuecomment-1251329766)

eps1lon:
> The culprit is in @testing-library/cypress since it has @testing-library/dom as a dependency with caret:

[...]

PR #1170, opened 2022-09-19
fix: Pin supported target environment

eps1lon:
**What**:

Closes https://github.com/testing-library/dom-testing-library/issues/1169

**Why**:

A minor release created a build with that contained syntax that we don't support yet.

**How**:

Run `npx browserslist defaults` with a version of `caniuse-lite` that was used at the time of the latest good release. 
That snapshot is now contained in our manifest and ensures build tools transpile to this target.
Having "defaults" as the target environment is too weak of a contract to have. 
Now it's apparent that changing the target environment is considered a breaking change.
We can decide in the next major release what a good matrix for supported runtimes is. 

**Checklist**:

- [x] Tests `rm -rf node_modules/; npm install; npm run build`, check `dist/@testing-library/dom.esm.js`, no more `??`
- [ ] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,25 @@
   "engines": {
     "node": ">=12"
   },
+  "browserslist": [
+    "and_chr 103",
+    "and_ff 101",
+    "and_qq 10.4",
+    "and_uc 12.12",
+    "android 103",
+    "chrome 102",
+    "edge 102",
+    "firefox 91",
+    "ios_saf 12.2-12.5",
+    "kaios 2.5",
+    "op_mini all",
+    "op_mob 64",
+    "opera 88",
+    "safari 15.5",
+    "samsung 17.0",
+    "samsung 16.0",
+    "node 12.0"
+  ],
   "scripts": {
     "build": "kcd-scripts build  --no-ts-defs --ignore \"**/__tests__/**,**/__node_tests__/**,**/__mocks__/**\" && kcd-scripts build --no-ts-defs --bundle --no-clean",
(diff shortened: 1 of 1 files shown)


MichaelDeBoey (commented):
@eps1lon Pinning `kcd-scripts` won't help, as we're not using latest version yet

eps1lon:
> Pinning kcd-scripts won't help, as we're not using latest

Please see the PR title and description. I'm not pinning kcd-scripts but everything.

package.json
```diff
@@ -23,6 +23,25 @@
...
+    "firefox 91",
+    "ios_saf 12.2-12.5",
+    "kaios 2.5",
+    "op_mini all",
+    "op_mob 64",
+    "opera 88",
+    "safari 15.5",
+    "samsung 17.0",
+    "samsung 16.0",
+    "node 12.0"
```
eps1lon, line 43: `kcd-scripts` does not use the `engines` field in its "--bundle" target. This is incorrect since it conflates the `module` entry with the `browser` entry. `module` and `main` are only for the used module system. Their runtime target needs to be the same. This isn't formally specified but conventionally used by bundlers (see https://esbuild.github.io/api/#main-fields).

MichaelDeBoey: CC/ @kentcdodds

MatanBobi:
@eps1lon, it looks like `op_mob 64` isn't available and we should upgrade this to 73 but I'm not sure if that will add the nullish coalescing issue again.

CC: @timdeschryver

PR #1184, opened 2022-10-26
add `excludeHidden` for *ByText queries

rluvaton:
**What**: add `excludeHidden` option for `*ByText` queries

**Why**: #196

**How**: used the already existing `isInaccessible` function

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged
      

Before I start adding tests, what do you think about this?

-----

Cc @kentcdodds 
> I'll reopen this and I'd be willing to look at a PR for this.
> https://github.com/testing-library/dom-testing-library/issues/196#issuecomment-628268484

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -24,6 +24,8 @@ let config: InternalConfig = {
   eventWrapper: cb => cb(),
   // default value for the `hidden` option in `ByRole` queries
   defaultHidden: false,
+  // default value for the `hidden` option in `ByText` queries
+  defaultExcludeHiddenText: false,
   // default value for the `ignore` option in `ByText` queries
   defaultIgnore: 'script, style',
   // showOriginalStackTrace flag to show the full error stack traces for async errors
--- a/src/queries/role.js
+++ b/src/queries/role.js
@@ -13,7 +13,7 @@ import {
   getImplicitAriaRoles,
   prettyRoles,
   isInaccessible,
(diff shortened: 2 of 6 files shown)


rluvaton:
Before I start adding tests, what do you think about this?

-----

Cc @kentcdodds 
> I'll reopen this and I'd be willing to look at a PR for this.
> https://github.com/testing-library/dom-testing-library/issues/196#issuecomment-628268484

eps1lon (commented):
Why not use the same name for the option as ByRole does?

Also note that this has considerable perf impact. ByText etc are a faster, less accurate helper and with this change we're making the distinction between By* and ByRole less clear.

kentcdodds:
I think that's fair @eps1lon. As I'm no longer maintaining Testing Library, I don't think it's fair for me to make decisions here. I think the best would be to have this supported in all the queries and the "faster, less accurate helper" can be offered by using the option.

eps1lon:
I'm not fundamentally opposed. But the added dimension due to the naming of the option is a bigger concern for me right now.

rluvaton:
I used a different name to avoid confusion with `*ByRole` as that `hidden: true` there means to show hidden as well
While in the `*ByText` we want to exclude them, and if we use the same configuration we would break our users as currently `defaultHidden` is false - meaning that we don't return hidden results and which is not what we currently do for `*ByText` queries...

PR #1186, opened 2022-10-28
chore: Migrate `ByRole` to TypeScript

DaniAcu:
**What**:
Migrate role.js to role.ts.

**Why**:

Because is the only file that is in js. I wold like to add more inteligence to ByRole, but for any other thing we need to add it first as ts file.

**How**:

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/queries/role.ts
+++ b/src/queries/role.ts
@@ -2,7 +2,7 @@ import {
   computeAccessibleDescription,
   computeAccessibleName,
 } from 'dom-accessibility-api'
-import {roles as allRoles, roleElements, ARIAPropertyMap} from 'aria-query'
+import {roles as allRoles, roleElements, ARIAPropertyMap, ARIARoleDefintionKey} from 'aria-query'
 import {
   computeAriaSelected,
   computeAriaChecked,
@@ -52,21 +52,21 @@ const queryAllByRole: AllByRole = (
 
   if (selected !== undefined) {
     // guard against unknown roles
-    if (getRoleModelFor(role).getProp('aria-selected') === undefined) {
+    if (allRoles.get(role as ARIARoleDefintionKey)?.props['aria-selected'] === undefined) {
       throw new Error(`"aria-selected" is not supported on role "${role}".`)
     }
   }
 
   if (checked !== undefined) {
     // guard against unknown roles
-    if (getRoleModelFor(role).getProp('aria-checked') === undefined) {
+    if (allRoles.get(role as ARIARoleDefintionKey)?.props['aria-checked'] === undefined) {
       throw new Error(`"aria-checked" is not supported on role "${role}".`)
     }
   }
 
   if (pressed !== undefined) {
     // guard against unknown roles
-    if (getRoleModelFor(role).getProp('aria-pressed') === undefined) {
+    if (allRoles.get(role as ARIARoleDefintionKey)?.props['aria-pressed'] === undefined) {
       throw new Error(`"aria-pressed" is not supported on role "${role}".`)
     }
   }
@@ -76,7 +76,7 @@ const queryAllByRole: AllByRole = (
     // guard against unknown roles
     // All currently released ARIA versions support `aria-current` on all roles.
     // Leaving this for symetry and forward compatibility
-    if (getRoleModelFor(role).getProp('aria-current') === undefined) {
+    if (allRoles.get(role as ARIARoleDefintionKey)?.props['aria-current'] === undefined) {
       throw new Error(`"aria-current" is not supported on role "${role}".`)
     }
   }
@@ -88,7 +88,7 @@ const queryAllByRole: AllByRole = (
 
   if (expanded !== undefined) {
     // guard against unknown roles
-    if (getRoleModelFor(role).getProp('aria-expanded') === undefined) {
+    if (allRoles.get(role as ARIARoleDefintionKey)?.props['aria-expanded'] === undefined) {
       throw new Error(`"aria-expanded" is not supported on role "${role}".`)
     }
   }
@@ -218,16 +218,6 @@ function makeRoleSelector(role: ByRoleMatcher, exact: boolean, customNormalizer?
     .concat(Array.from(implicitRoleSelectors))
     .join(',')
 }
-
-function getRoleModelFor (role: ByRoleMatcher) {
-  return {
-    getProp<N extends keyof ARIAPropertyMap>(propName: N) {
-      if(typeof role !== "string") return;
-      return allRoles.get(role)?.props?.[propName] as ARIAPropertyMap[N];
-    }
-  }
-}
-
 
 const getNameHint = (name: ByRoleOptions["name"]): string => {
   let nameHint = ''
--- a/src/queries/role.ts
+++ b/src/queries/role.ts
@@ -151,7 +151,7 @@ const queryAllByRole: AllByRole = (
         return matcher(firstWord, node, role, matchNormalizer)
       }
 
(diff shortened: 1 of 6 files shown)


eps1lon (changes requested):
For TS migration please don't change runtime behavior (yet). This makes reviewing harder than it needs to be. 

If the type-checker has issues with things, we can use `@ts-expect-error` or `as any` for now and then slowly migrate to a more type-safe version.

DaniAcu:
@eps1lon Ohh okay, thanks. I didn't know that but that make sense. Let me update the PR to just have ts migration here. Thanks, again

eps1lon (changes requested):
Lint is failing. Is there anything else that's unclear to you from the CI logs?

CI needs to be green for this to be mergeable. If you're stuck at some point, feel free to ping me and we can go through the CI logs to see what went wrong.

DaniAcu:
@eps1lon I delayed more in make the fixes becuase lint is not working locally. 
image

So, I used the github jobs to check and fix it. 

Related to the changes, I think that the types are not completly clean or organized. I will try to get into it a little bit more now that roles are also in ts. 

Thanks for awesome welcoming to contribute here, I'm happy to could do it.

DaniAcu:
@eps1lon there anything more that i need to change? 🤔

DaniAcu:
@eps1lon I updated based on `main` and fix a error after merge. Could you re check it?

src/queries/role.ts
```diff
@@ -247,7 +282,10 @@ const getMissingError = (
   }
 
   let roles = ''
-  Array.from(container.children).forEach(childElement => {
+  Array.from(
+    /* istanbul ignore next */
+    container?.children ?? [],
```
eps1lon, line 287: Same here. Let's not change runtime and use type casting/assertions instead.

types/__tests__/type-tests.ts
```diff
@@ -178,15 +178,9 @@ export async function testByRole() {
   console.assert(queryByRole(element, 'button', {name: /^Log/}) === null)
   console.assert(
     queryByRole(element, 'button', {
-      name: (name, el) => name === 'Login' && el.hasAttribute('disabled'),
+      name: (name, el) => name === 'Login' && !!el?.hasAttribute('disabled'),
```
eps1lon, line 181: why does this need to change?

It's also pretty hard to grok what this is doing.

types/__tests__/type-tests.ts
```diff
@@ -178,15 +178,9 @@ export async function testByRole() {
...
     }) === null,
   )
-
-  // allow to query for a role that isn't included in the types
-  console.assert(queryByRole(element, 'foo') === null)
-  console.assert(queryByRole(element, /foo/) === null)
-  console.assert(screen.queryByRole('foo') === null)
-  console.assert(screen.queryByRole(/foo/) === null)
```
eps1lon, line 189: Why remove these?

types/queries.d.ts
```diff
@@ -111,24 +116,17 @@ export interface ByRoleOptions extends MatcherOptions {
   /**
    * Only considers elements with the specified accessible name.
    */
-  name?:
-    | RegExp
-    | string
-    | ((accessibleName: string, element: Element) => boolean)
+  name?: RegExp | string | MatcherFunction
```
eps1lon, line 119: The original signature included useful type hints (e.g. `accessibleName`). These are lost with a generic `MatcherFunction`.

Could we instead use a concrete `NameMatcherFunction` to get type hints and keep the improved union?

types/queries.d.ts
```diff
@@ -111,24 +116,17 @@ export interface ByRoleOptions extends MatcherOptions {
...
-    | ((accessibleDescription: string, element: Element) => boolean)
+  description?: RegExp | string | MatcherFunction
 }
 
-export type AllByRole<T extends HTMLElement = HTMLElement> = (
-  container: HTMLElement,
-  role: ByRoleMatcher,
-  options?: ByRoleOptions,
-) => T[]
+export type AllByRole = QueryMethod<
```
eps1lon, line 126: Same here. By using the generic `QueryMethod` we loose type hints. Let's type it out properly.

src/queries/role.ts
```diff
@@ -75,27 +97,31 @@ function queryAllByRole(
...
   }
 
-  if (level !== undefined) {
-    // guard against using `level` option with any role other than `heading`
-    if (role !== 'heading') {
-      throw new Error(`Role "${role}" cannot have "level" property.`)
-    }
+  // guard against using `level` option with any role other than `heading`
+  if (level !== undefined && role !== 'heading') {
+    throw new Error(`Role "${role}" cannot have "level" property.`)
```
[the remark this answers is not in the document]
DaniAcu: Ohh! It's to be cosisten with the other queries right? Sounds good.

types/__tests__/type-tests.ts
DaniAcu: Name require a boolean value. `el` variable could be undefined based on typing. The boolean casting is because type result could be udefined by `el` or boolean by `hasAttribute`

Maybe is more readable like that

```ts
name: (name, el) => name === 'Login' && !!el && el.hasAttribute('disabled'),
```

types/queries.d.ts
DaniAcu: Oh okay, I think QueryMethod was the right way. Do you know what are the type hints that we lose that you mention.

DaniAcu: Just to take in care when I work in that

types/__tests__/type-tests.ts
eps1lon: Why is `el` nullable? Shouldn't that always be defined?

types/__tests__/type-tests.ts
eps1lon: Why is this not allowed? `query*` can return a nullable value

types/queries.d.ts
eps1lon: In editor suggestions:
original:
Screenshot 2023-01-10 at 17 46 40

your change:

Screenshot 2023-01-10 at 17 46 53

Notice the name of the parameters

types/__tests__/type-tests.ts
DaniAcu: Because "foo" is not a valid role according to types

types/queries.d.ts
DaniAcu: ohhh!

DaniAcu: Thank you!

src/queries/role.ts
```diff
@@ -129,7 +156,7 @@ function queryAllByRole(
         return matcher(firstWord, node, role, matchNormalizer)
       }
 
-      const implicitRoles = getImplicitAriaRoles(node)
+      const implicitRoles = getImplicitAriaRoles(node) as string[]
```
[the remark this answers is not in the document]
DaniAcu: role-helpers need migration to typescript so it return any[]

DaniAcu:
@eps1lon hi, i keep the changes simple and i dismiss any change that affect run time. There some ts comments, but I think it's fine for the migration. There others things that need to be change in general, but for the proposal of this PR, i think that with this changes is enough.

Let me know if there any thing that you look weird in this changes.

And thanks in advance for the time that you take to revisit it

eps1lon (approved):
Added some finishing touches. Thank you for sticking with it!

eps1lon:
@all-contributors add @DaniAcu for code

Issue #1202, opened 2023-01-09
ByRole: Constrain API further


eps1lon:
### Describe the feature you'd like:

As part of https://gist.github.com/bvaughn/d3c8b8842faf2ac2439bb11773a19cec I'm revisiting a lot of selector APIs from different libraries.

The API of `ByRole` stood out with regard to role selection and its options. 

### Suggested implementation:

1. Constraint `role` to `string`
2. ~match fallback roles (e.g. `byRole('none') === byRole('presentation')` in `<div role="none presentation" />`~ Leaving for now. Not so clear what's the best default here
3. Remove `exact` option
4. Remove `normalizer` option

### Describe alternatives you've considered:

Document concrete use cases (motivated by real-world tests). If you have some, please let me know.

### Teachability, Documentation, Adoption, Migration Strategy:

We can document fallback role matching in a footnote. I expect that most queries will not be concerned with fallback roles.

Constraining the API makes teaching this query easier.

eps1lon:
Not dropping the `queryFallbacks` for now. Not clear to me if we should restrict that and what the best default would be.,

MatanBobi:
@eps1lon can we close this issue? I think that all of the points were covered as part of #1211.

eps1lon:
Closes in https://github.com/testing-library/dom-testing-library/pull/1211
Released in `@testing-library/dom@9.0.0`

charliematters:
Sorry to dig up a closed comment, but we've just upgrade to the versino with this fix in, and this query no longer works. Maybe we're abusing it, and there's a better way, but I can't find any sugestions from the docs? We use this to wait for a modal with a given string to appear 

```typescript
 const modal = await screen.findByRole(
    (content, element) => {
      return (
        content === 'dialog' &&
        within(element as HTMLElement).queryByText(title) !== null
      )
    }
    ```

PR #1207, opened 2023-01-28
feat: Drop support for Node.js 12.x

eps1lon:
BREAKING CHANGE: Minimum supported Node.js version is now 14.x

**What**:

Drop support for Node.js 12.x

**Why**:

Reached EOL in April 2022. Tooling is starting to drop Node.js 12.x (see JSDOM) as well so it's time to upgrade to be able to test compat with latest test frameworks (e.g. Jest).

**How**:

Targetting alpha for now. Want to batch this with https://github.com/testing-library/dom-testing-library/issues/1202
Look for occurence of "12" and update where appropriate. Also introducing Node.js 18.x into test matrix since that is now in Active LTS

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,5 +1,5 @@
 {
   "installCommand": "install:csb",
   "sandboxes": ["github/kentcdodds/react-testing-library-examples"],
-  "node": "12"
+  "node": "14"
 }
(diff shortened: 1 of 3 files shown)


timdeschryver (approved):
LGTM 👍
Do you think we should also add v19 to the test matrix?

eps1lon:
> Do you think we should also add v19 to the test matrix? 

Sure

PR #1211, opened 2023-02-06
fix(ByRole): Constrain API 

eps1lon:
BREAKING CHANGE: Only allow `string` as a `role`.
Drop support for `exact`, `trim`, `collapseWhitespace`, and `normalizer` options.

**What**:

Closes https://github.com/testing-library/dom-testing-library/issues/1202

**Why**:

Options are not needed as far as I can tell. From the tests we have, it seems to me this encourages rather bad queries. Will ship this soon-ish (since it's just an alpha) to test.

**How**:

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs): https://github.com/testing-library/testing-library-docs/pull/1214
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -788,33 +788,26 @@ test('queryAllByRole returns semantic html elements', () => {
     </form>
   `)
 
-  expect(queryAllByRole(/table/i)).toHaveLength(1)
-  expect(queryAllByRole(/tabl/i, {exact: false})).toHaveLength(1)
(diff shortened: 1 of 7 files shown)


timdeschryver (approved):
👍 I like the change, I think this guards us against misusage of it and also remove some confusion

remidej:
Hello! I know byRole often has perf issues when used on large dom nodes. Do you expect this PR to improve the performance?

PR #1213, opened 2023-02-09
release: 9.0.0

eps1lon:
Will be fast-forward merge on 16.02.2023 unless we need to revise some changes.

Tested latest alpha on the Klarna monorepo and this caught some bad queries assuming `exact` applied to the `name` option. But `exact` applies to the `role`. This is why these options were removed: they're misleading and most likely not needed.

### Bug Fixes

* **ByRole:** Constrain API ([#1211](https://github.com/testing-library/dom-testing-library/issues/1211)) ([746def6](https://github.com/testing-library/dom-testing-library/commit/746def61471e51ba43a519f68c51c57fb0c84958))

### BREAKING CHANGES

* **ByRole:** Only allow `string` as a `role`.
Drop support for `exact`, `trim`, `collapseWhitespace`, and `normalizer` options.

### Features

* Drop support for Node.js 12.x ([#1207](https://github.com/testing-library/dom-testing-library/issues/1207)) ([c37ea2e](https://github.com/testing-library/dom-testing-library/commit/c37ea2e711af79184a4a71c68feefc08a53618b5))

### BREAKING CHANGES

* Minimum supported Node.js version is now 14.x


timdeschryver (approved):
I tested this on my codebase without breaking change 🚀

eps1lon:
Small update: It did catch some breaking changes during type-checking. Specifically, we had a bunch of queries using `exact: false`. It looks like it was assumed `exact` applies to `name` not `role` considering that removing the option did not break runtime tests. If it was intended (e.g. including `role="spinbutton"`) then I'm ok with this breaking change. People should make it explicit what roles they expect instead of relying on perfect knowledge of all possible roles (which is especially problematic since new roles can be added to ARIA).

PR #1215, opened 2023-02-17
test: add missing test for suggestions

MatanBobi:
**What**: Add a test that was missing for throwing suggestions

**Why**: While working on the docs requested in the [discussion between](https://github.com/testing-library/testing-library-docs/pull/1214#discussion_r1099027874) @eps1lon and @timdeschryver, I saw that based on the implementation, there's a missing test.

**Checklist**:

- [X] Tests
- [X] Ready to be merged

### DIFF
--- a/src/__tests__/suggestions.js
+++ b/src/__tests__/suggestions.js
@@ -7,7 +7,7 @@ beforeAll(() => {
 })
 
 afterEach(() => {
-  configure({testIdAttribute: 'data-testid'})
+  configure({testIdAttribute: 'data-testid', throwSuggestions: true})
   console.warn.mockClear()
 })
 
@@ -110,7 +110,6 @@ test('should suggest when suggest is turned on for a specific query but disabled
   <button data-testid="foot">another</button>`)
 
   expect(() => screen.getByTestId('foo', {suggest: true})).toThrowError()
-  configure({throwSuggestions: true})
 })
 
 test('should suggest getByRole when used with getBy', () => {
--- a/src/__tests__/suggestions.js
+++ b/src/__tests__/suggestions.js
@@ -109,7 +109,7 @@ test('should suggest when suggest is turned on for a specific query but disabled
   <button data-testid="foo">submit</button>
   <button data-testid="foot">another</button>`)
 
-  expect(() => screen.getByTestId('foo', {suggest: true})).toThrowError()
+  expect(() => screen.getByTestId('foo', {suggest: true})).toThrowError("try this:\ngetByRole('button', { name: /submit/i })")
 })
 
 test('should suggest getByRole when used with getBy', () => {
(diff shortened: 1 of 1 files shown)


src/__tests__/suggestions.js
```diff
@@ -103,6 +103,16 @@ test('should not suggest when suggest is turned off for a query', () => {
...
 })
 
+test('should suggest when suggest is turned on for a specific query but disabled in config', () => {
+  configure({throwSuggestions: false})
+  renderIntoDocument(`
+  <button data-testid="foo">submit</button>
+  <button data-testid="foot">another</button>`)
+
+  expect(() => screen.getByTestId('foo', {suggest: true})).toThrowError()
+  configure({throwSuggestions: true})
```
eps1lon, line 113: These setup+cleanup tasks need to be in a `beforeEach`/`afterEach`. Otherwise the whole test run would be tainted if this single test fails.

eps1lon (changes requested):

src/__tests__/suggestions.js
MatanBobi: Since this is the only test needing them, I didn't want to add the overhead of running that code in a 'beforeEach' and 'afterEach' so I changed it at the beginning and returned it to what the 'beforeAll' hook defined. Do you think it will be better in this case to do it in the hooks instead?

eps1lon: every setup/cleanup for tests needs to be in hooks for the the reason I explained: once a test fails, the whole test run is tainted because because its cleanup didn't run. This can be super confusing if you try to debug the last test that fails when the only reason for that test failure is the first test.
Concretely:
say we have test 1 and 2. 1 needs environment A and 2 needs B. 1 sets up env A but fails. Now 2 also fails because it has env A+B. If you don't check the implementation of test 1 first, you might start debugging test 2 even though there's nothing in test 2 that needs fixing.

MatanBobi: You're right, that's a good point, thanks.
I've put the cleanup as part of the `afterEach` that was already there but I believe that the setup should remain within the test as it's the only test that needs this setup. Wdyt?

src/__tests__/suggestions.js
```diff
@@ -103,6 +103,15 @@ test('should not suggest when suggest is turned off for a query', () => {
   ).not.toThrowError()
 })
 
+test('should suggest when suggest is turned on for a specific query but disabled in config', () => {
+  configure({throwSuggestions: false})
+  renderIntoDocument(`
+  <button data-testid="foo">submit</button>
+  <button data-testid="foot">another</button>`)
+
+  expect(() => screen.getByTestId('foo', {suggest: true})).toThrowError()
```
eps1lon, line 112: This isn't really testing whether we're suggesting anything. Just that an error is thrown. I'll see if a snapshot makes more sense

eps1lon (approved):

PR #1220, opened 2023-03-08
test: Improve test perf for suggestion tests

eps1lon:
Brings down the duration of `suggestions.js` tests from ~9s to <1s.

Was always bugging me how sluggish `suggestions.js` tests were. Revisited them during https://github.com/testing-library/dom-testing-library/pull/1215/files and it turns out they're only slow because of real timers. 

However, the test is not concerned with time but the suggesions themselves. I originally just set the timeout to 1 for `find*` queries but this felt noisy and means we might regress in the future.

### DIFF
--- a/src/__tests__/suggestions.js
+++ b/src/__tests__/suggestions.js
@@ -6,7 +6,14 @@ beforeAll(() => {
   configure({throwSuggestions: true})
 })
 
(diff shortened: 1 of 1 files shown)


MatanBobi (dismissed):
Good catch :)

PR #1221, opened 2023-03-08
feat(ByRole): Allow filter by disabled state

eps1lon:
Part of a larger committment to unify APIs for `/react-native` and `/react` (which is just `/react-dom` at the moment)

**What**:

Enables `ByRole(role, { disabled })`

**Why**:

Makes writing tests for react-native and react-dom easier by removing barriers when changing the target platform that's being tested.

**How**:

Same patterns as for expanded, selected etc

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs) PENDING
- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/ariaAttributes.js
+++ b/src/__tests__/ariaAttributes.js
@@ -258,3 +258,57 @@ test('`expanded: true|false` matches `expanded` elements with proper role', () =
   expect(getByRole('button', {expanded: true})).toBeInTheDocument()
   expect(getByRole('button', {expanded: false})).toBeInTheDocument()
 })
+
+test('`disabled` throws on unsupported roles', () => {
+  const {getByRole} = render(
+    `<div role="alert" aria-disabled="true">Hello, Dave!</div>`,
+  )
+  expect(() =>
+    getByRole('alert', {disabled: true}),
+  ).toThrowErrorMatchingInlineSnapshot(
+    `"aria-disabled" is not supported on role "alert".`,
+  )
+})
+
+test('`disabled: true|false` matches `disabled` buttons', () => {
+  const {getByRole} = renderIntoDocument(
+    `<div>
+      <button disabled="true" />
+      <button />
+    </div>`,
+  )
+  expect(getByRole('button', {disabled: true})).toBeInTheDocument()
+  expect(getByRole('button', {disabled: false})).toBeInTheDocument()
+})
+
+test('`disabled: true|false` matches `aria-disabled` buttons', () => {
+  const {getByRole} = renderIntoDocument(
+    `<div>
+      <button aria-disabled="true" />
+      <button aria-disabled="false" />
+    </div>`,
+  )
+  expect(getByRole('button', {disabled: true})).toBeInTheDocument()
+  expect(getByRole('button', {disabled: false})).toBeInTheDocument()
+})
+
+test('`disabled` attributes overrides `aria-dsiabled`', () => {
+  const {getByRole} = renderIntoDocument(
+    `<div>
+      <button disabled="true" aria-disabled="false" />
+      <button />
+    </div>`,
+  )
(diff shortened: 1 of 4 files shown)


timdeschryver (approved):

src/__tests__/ariaAttributes.js
```diff
@@ -258,3 +258,57 @@ test('`expanded: true|false` matches `expanded` elements with proper role', () =
...
+})
+
+test('`disabled` attributes overrides `aria-dsiabled`', () => {
+  const {getByRole} = renderIntoDocument(
+    `<div>
+      <button disabled="true" aria-disabled="false" />
+      <button />
+    </div>`,
+  )
+  expect(getByRole('button', {disabled: true})).toBeInTheDocument()
```
MatanBobi, line 302: Just an idea, maybe for the completeness of this test, we should also add:
```
expect(queryByRole('button', {disabled: false})).not.toBeInTheDocument()
```

src/role-helpers.js
```diff
@@ -281,6 +281,29 @@ function computeAriaCurrent(element) {
   )
 }
 
+const elementsSupportingDisabledAttribute = new Set([
```
MatanBobi, line 284: I might be missing something, but this seems to me like something that should be defined in `aria-query` because it can change as part of the spec and having it here is like mixing the ARIA logic with our `role` implementation.
Wdyt?

src/role-helpers.js
```diff
@@ -281,6 +281,29 @@ function computeAriaCurrent(element) {
...
+ * @param {Element} element -
+ * @returns {boolean} -
+ */
+function computeAriaDisabled(element) {
+  return elementsSupportingDisabledAttribute.has(element.localName) &&
+    element.hasAttribute('disabled')
+    ? // https://www.w3.org/TR/html-aam-1.0/#html-attribute-state-and-property-mappings
+      true
+    : // https://www.w3.org/TR/wai-aria-1.1/#aria-disabled
+      element.getAttribute('aria-disabled') === 'true'
```
MatanBobi, line 304: I think it's worth adding this to [`jest-dom` assertion](https://github.com/testing-library/jest-dom/blob/main/src/to-be-disabled.js#L53)  too (I can do this on my own if you agree with me). 
Having two libraries in the organization considering different things as `disabled` is quite confusing IMO.

src/role-helpers.js
MatanBobi: Also, I see we have something like this in [`jest-dom`](https://github.com/testing-library/jest-dom/blob/main/src/to-be-disabled.js#L4) too, so it makes me think that this indeed shouldn't belong in one of our packages.

eps1lon: I don't want to put more stuff into `aria-query`. `dom-accesibility-api` makes more sense at this point. Feel free to raise a PR.

src/role-helpers.js
eps1lon: Querying should not be used for assertions. For jest-dom i.e. there was a clear case why this wasn't as straight forward because it might contain footguns. 

/dom: queries based on a11y tree
For assertions I'd rather remove the convenience query from jest-dom altogether in favor of an explicit query for just a11y disabled.

MatanBobi: I agree with that querying and asserting are two different operations.
Just so I'll understand, what are the footguns you're referring to?
I don't think removing the `toBeDisabled` assertion is a good idea as it's widely used but maybe I'm missing something.

src/role-helpers.js
MatanBobi: Done: https://github.com/eps1lon/dom-accessibility-api/pull/919
Please lmk if I can change stuff, I'm not happy with putting the change in the `util` file but I'm not familiar with the codebase, so I'm happy to change it.
Thanks!

jamesonhill:
Would be super helpful to have this

PR #1223, opened 2023-03-08
feat(ByRole): Allow filter by value state

eps1lon:
Part of a larger committment to unify APIs for `/react-native` and `/react` (which is just `/react-dom` at the moment)

**What**:

Enables `ByRole(role, { value })`

Note that we don't support `valuenow` but rather `value.now` (same pattern for other `aria-value*` attributes). One reason is parity with `@testing-library/react-native`. 

The other reason is that `value.*` works a bit different in that it looks at all `value.*` and combines them with `&&` instead of returning on the first match. Today, `{ pressed: true, selected: true}` has kind of an odd behavior since it only looks at whatever comes first in the implementation. We could combine them with `&&` as well but it's not clear to me yet that this is always good. So by putting them in a single object makes it clear that we look at the whole object which wouldn't be that obvious for `{ name, valuemin, valuemax}`. I don't feel too strongly about this specific rationale. I care more about parity with /react-native.

**Why**:

Makes writing tests for react-native and react-dom easier by removing barriers when changing the target platform that's being tested.

**How**:

Same patterns as for expanded, selected etc

**Checklist**:

- [x] Documentation: https://github.com/testing-library/testing-library-docs/pull/1230
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/ariaAttributes.js
+++ b/src/__tests__/ariaAttributes.js
@@ -237,6 +237,101 @@ test('`level` throws on unsupported roles', () => {
   )
 })
 
+test('`value.now` throws on unsupported roles', () => {
+  const {getByRole} = render(`<button aria-valuenow="1">Button</button>`)
+  expect(() =>
+    getByRole('button', {value: {now: 1}}),
+  ).toThrowErrorMatchingInlineSnapshot(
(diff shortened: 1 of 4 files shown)


timdeschryver (approved):
One small nit, but can be ignored if you want.

PR #1229, opened 2023-04-30
feat(waitFor): Support async `unstable_advanceTimersWrapper`

eps1lon:
`unstable_advanceTimersWrapper` now receives an `async` function. If your wrapper just returns the given callback, nothing should break (e.g. this is [not a breaking change for React Testing Library](https://github.com/testing-library/react-testing-library/pull/1205/files). But if you call it first with the expectation that it resolved synchronously, you may observe changed behavior. This method is marked as `unstable_` and not subject to semantic versioning. Still releasing as a feature for safe measure.

**What**:

Add support for having an `async` `unstable_advanceTimersWrapper`

**Why**:

To support experimentation with `async` `act`

**How**:

`await` `advanceTimersWrapper`. This also enables us to flush the microtask queue that exist after the macrotask.

This changes the observed semantics in `callback` a bit since `callback` now observes macrotask+microtask instead of microtask+macrotask. 
I don't think that matters though. We definitely have no test in this library and RTL tests passed using this version of `waitFor`.

**Checklist**:

- [x] `checkCallback` semantics were never documented so this is fine. `unstable_advanceTimersWrapper` has no docs either and I don't feel comfortable adding some.
- [x] Tests (also tested integration with https://github.com/testing-library/react-testing-library/pull/1205/files)
- [x] Ready to be merged

### DIFF
--- a/src/wait-for.js
+++ b/src/wait-for.js
@@ -67,12 +67,18 @@ function waitFor(
           reject(error)
           return
         }
-        // we *could* (maybe should?) use `advanceTimersToNextTimer` but it's
-        // possible that could make this loop go on forever if someone is using
-        // third party code that's setting up recursive timers so rapidly that
-        // the user's timer's don't get a chance to resolve. So we'll advance
-        // by an interval instead. (We have a test for this case).
-        advanceTimersWrapper(() => {
+
+        // In this rare case, we *need* to wait for in-flight promises
+        // to resolve before continuing. We don't need to take advantage
+        // of parallelization so we're fine.
+        // https://stackoverflow.com/a/59243586/971592
+        // eslint-disable-next-line no-await-in-loop
+        await advanceTimersWrapper(async () => {
+          // we *could* (maybe should?) use `advanceTimersToNextTimer` but it's
+          // possible that could make this loop go on forever if someone is using
+          // third party code that's setting up recursive timers so rapidly that
+          // the user's timer's don't get a chance to resolve. So we'll advance
+          // by an interval instead. (We have a test for this case).
           jest.advanceTimersByTime(interval)
         })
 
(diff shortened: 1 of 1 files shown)


timdeschryver (approved):
I expected a breaking change with this, but I tested this on a small project and everything continues to work fine 👍

src/wait-for.js
```diff
@@ -67,12 +67,18 @@ function waitFor(
...
-        // the user's timer's don't get a chance to resolve. So we'll advance
-        // by an interval instead. (We have a test for this case).
-        advanceTimersWrapper(() => {
+
+        // In this rare case, we *need* to wait for in-flight promises
+        // to resolve before continuing. We don't need to take advantage
+        // of parallelization so we're fine.
+        // https://stackoverflow.com/a/59243586/971592
+        // eslint-disable-next-line no-await-in-loop
+        await advanceTimersWrapper(async () => {
```
salazarm, line 76: Just curious as an outsider, why is `await` and `async` needed here when we don't explicitly `await` anything? Are we trying to wait a tick?

src/wait-for.js
```diff
@@ -85,18 +91,6 @@ function waitFor(
...
-
-        // In this rare case, we *need* to wait for in-flight promises
-        // to resolve before continuing. We don't need to take advantage
-        // of parallelization so we're fine.
-        // https://stackoverflow.com/a/59243586/971592
-        // eslint-disable-next-line no-await-in-loop
-        await advanceTimersWrapper(async () => {
-          await new Promise(r => {
-            setTimeout(r, 0)
-            jest.advanceTimersByTime(0)
```
salazarm, line 97: Curious if this was a bug but shouldn't the order of these 2 statement been reversed? ie. the promise should resolve after advancing timers?

bondiano:
You are the fucking idiot, I waste 4 hours to get know this was misstake in such little PR, thanks for good sex

PR #794, opened 2020-10-20
fix(types): make third and fourth params optional for findAllBy and findBy built queries

zorfling:
**What**: Updated types for built findBy / findAllBy custom queries and added extra type tests

**Why**: A fix made in PR #574 / Issue #534 added these parameters but accidentally made them required (see https://github.com/testing-library/dom-testing-library/pull/574#issuecomment-629213137)

**How**: Updated the types in `query-helpers.d.ts` to mimic the built-in types in `queries.d.ts`, and added more type tests to ensure it won't regress.

**Checklist**:

- [x] Documentation added to the
      [docs site](https://github.com/testing-library/testing-library-docs) N/A
- [x] Tests
- [x] Typescript definitions updated
- [x] Ready to be merged

PR #574, opened 2020-05-15
fix(types): add fourth param to build findAllBy and findBy queries

mzdunek93:
Fixes #534

PR #1230, opened 2023-05-05
In FindAllBy and FindAll query helper types: spread Arguments to support no arguments

dturcotte:
**What**: Custom queries returned by `buildQueries` whose implementation required no arguments (besides `container`) still required one argument for `find*` and `findAll*` even when used on `screen` or `within`. 

Without the `query-helpers.d.ts` changes introduced by this PR, see these TS errors in the updated `type-tests`:
Screen Shot 2023-05-05 at 11 48 12 AM

**Why**: There was a lot of great work done on these types [here](https://github.com/testing-library/dom-testing-library/pull/794) and [here](https://github.com/testing-library/dom-testing-library/pull/574), I want to build on that to solve this one additional edge case. 

For context on why we have queries with no arguments: we have a library with UI components that other teams consume. We also provide a few custom queries for things we don't want the consumers to worry about. For example: We provide a `DataGrid` component with the option to select rows. It's possible to find the input to select a row using base testing-library features but consumers shouldn't need to know how the sausage is made, for their convenience we want them to be able to simply say:
`await userEvent.click(within(row).getGridRowSelectionInput())`

This works, however I noticed that for the `find` and `findAll` queries the first arg is always required. It would end up looking like this to make TS happy:
`await within(row).findGridRowSelectionInput(undefined)`

**How**: Spreading `Arguments` in `FindBy` and `FindAllBy` types would fix this; the resulting queries should better match the types passed into `buildQueries`. If there are no arguments after `container` then the resulting query would have two: `container` and `waitForOptions`. This would also allow for more than two args if needed. Lastly, this doesn't change the fact that `container` is required:

Screen Shot 2023-05-05 at 11 46 16 AM

**Checklist**:

- [x] Tests (in type-tests.ts, I think this is right based on previous PRs but it failed the pre-commit hook validation)
- [x] Ready to be merged

### DIFF
--- a/types/__tests__/type-tests.ts
+++ b/types/__tests__/type-tests.ts
@@ -132,6 +132,50 @@ export async function testQueryHelpers() {
   screenWithCustomQueries.getByAutomationId(['id', 'automationId'])
   await screenWithCustomQueries.findAllByAutomationId('id', {}, {timeout: 1000})
   await screenWithCustomQueries.findByAutomationId('id', {}, {timeout: 1000})
+
+  // contrived example of a custom query with no arguments beyond container
+  function queryAllByFoo(container: HTMLElement) {
+    return queryAllByText(container, 'Foo')
+  }
+
+  const [
+    queryByTextFoo,
+    getAllByTextFoo,
+    getByTextFoo,
+    findAllByTextFoo,
+    findByTextFoo,
+  ] = buildQueries(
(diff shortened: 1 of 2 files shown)


astorije (approved):
@eps1lon, @MatanBobi, @timdeschryver, what do you all think about this? 🙏

PR #1236, opened 2023-06-14
Fix role assignment for imgs with alt attribute

### LINKED ISSUE #1235: Images with alt attributes cannot be grabbed ByRole
- `@testing-library/dom` version: 9.3.0
- Testing Framework and version: Jest/React
  
- DOM Environment:
  

### The Issue
If an `img` tag has an `alt` attribute, its role shows up as `presentation` instead of as `img`. So, e.g., this test:
 
```js
import { render, screen } from '@testing-library/react';

test("Displays an img", () => {
  render (<img src="https://http.cat/418" alt="404" />);
  screen.getByRole('img')
})
```

produces this result:

ImgError

### ARIA Background

According to the ARIA [docs](https://www.w3.org/TR/html-aria/#docconformance) an `img` should only default to the role of `presentation` if it has an `alt` attribute equal to "":
 
AriaImg

### Source of the Problem and Proposed Solution

The problem arises when the selector is being made in `role-helpers.js`: 

https://github.com/testing-library/dom-testing-library/blob/d09b3c25b07486c17603317f4c5a0320c188cbed/src/role-helpers.js#L80-L94 

Because a value of `""` resolves to false in line 87, the selector for a presentation `img` ends up returning `[alt]` instead of `[alt=""]` as its characteristic attribute, meaning that any `img` with an `alt` attribute will be marked with a `presentation` role. Simply adding an explicit check for `value === ""` on line 87 fixes the issue. (The check could be made more specific if that is needed.)

I will submit a PR with the proposed fix and updated tests.

brtrick:
**What**: 
This PR fixes a bug in assigning the role to imgs.
Fixes #1235  

**Why**:

**How**:

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/__snapshots__/role-helpers.js.snap
+++ b/src/__tests__/__snapshots__/role-helpers.js.snap
@@ -1,12 +1,38 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`logRoles calls console.log with output from prettyRoles 1`] = `
-region:
+section:
 
 Name "a region":
 <section
   aria-label="a region"
   data-testid="named-section"
+/>
+
+--------------------------------------------------
+img:
+
+Name "404":
+<img
+  alt="404"
+  data-testid="img-alt"
+  src="http://imgsrc.com"
+/>
+
+Name "":
+<img
+  data-testid="img-no-alt"
+  src="http://imgsrc.com"
+/>
+
+--------------------------------------------------
+presentation:
+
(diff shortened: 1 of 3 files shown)


MatanBobi:
I'm closing this one because AFAIU it's resolved by #1241.
@jlp-craigmorten, @brtrick if I got it wrong, please comment here and I'll re-open :)
Thanks for the time working on this one!

PR #1240, opened 2023-06-15
chore: add accessibility-alt-text-bot

MatanBobi:
**What**: Adds a workflow to add the accessibility alt text bot.
Following this [guide](https://github.blog/2023-06-12-make-your-github-projects-more-accessible-with-accessibility-alt-text-bot/).

**Why**: To enforce adding alt text to images in issues and PR's.

**Checklist**:

- [ ] ~~Documentation added to the
      [docs site](https://github.com/testing-library/testing-library-docs)~~
- [ ] ~~Tests~~
- [ ] ~~TypeScript definitions updated~~
- [X] Ready to be merged


MatanBobi:
> I like the idea. But the build seems to be having some troubles with it 👀

Oh, I think it's failing because I'm not synced with master and we pinned browserlist, I'll fix that :) Thanks!

nickmccurdy (approved):

nickmccurdy:
@MatanBobi Thanks! Feel free to merge this if you want, as this clearly isn't the cause of the CI failure, and merging should fix it anyway.

MatanBobi:
Ok, the reason it's failing is because we're testing on Node 14 and have an `overrides` solution for `browserlist` which doesn't work in npm 6. That's a problem because our tests will always fail for node 14 with this solution. I'm opening an issue about that.

nickmccurdy:
I think we can just merge this then, I'm not aware of any issues that aren't already on `main`.

nickmccurdy:
I think we accidentally merged to main, which still uses Node 14.

PR #1241, opened 2023-06-25
feat: upgrade aria-query to 5.3.0

### LINKED ISSUE #1100: can't find element by role `code`
- `@testing-library/dom` version: 12.1.2
jest 27.4.3
  
- DOM Environment:
jsdom 16.6.0
  

### Relevant code or config:

component
```js
export default ({ name }) => <code>{name}</code>;
```

test
```js
import "@testing-library/jest-dom/extend-expect";
import React from "react";
import { render } from "@testing-library/react";
import Code from "../hello";

test("To render a code element", () => {
  const { getByRole } = render(<Code name="Jill" />);
  expect(getByRole("code")).toBeInTheDocument();
});
```

### What you did:
I searched the code element by role `code`, I expected to find it by https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/structural_roles

### What happened:
The element is not found

### Reproduction:
https://codesandbox.io/s/fail-to-find-role-code-qpzrs?file=/src/__tests__/hello.js

### Problem description:
Per testing-library docs, it seems like it should work, as when I check the role of the element in chrome dev tools, I see 'code' as expected.

### Suggested solution:

not sure, I've seen that aria-query doesn't support this role.

### LINKED ISSUE #1150: Can not get by role "mark"

### LINKED ISSUE #1201: getByRole not finding Element by implicit role
- `@testing-library/dom` version: 16.9
- Testing Framework and version: 9.2
  
- DOM Environment: 4.1.0
  
See codesandbox

### Relevant code or config:

```js
test("getByRole", () => {
  render(
    <div>
      <strong>hello</strong>
    </div>
  );
  screen.getByRole("strong"); // errors
});
```

### What happened:

```js
Unable to find an accessible element with the role "strong"

  document:

  Name "":
  <body />

<body>
  <div>
    <div>
      <strong>
        hello
      </strong>
    </div>
  </div>
</body>
```

### Reproduction:

https://codesandbox.io/s/react-testing-library-demo-forked-lzjrl1?file=/src/__tests__/getByRole.js

### Problem description:

According to https://testing-library.com/docs/queries/byrole/, we should use the HTMLElement's role. According to [this table](https://www.w3.org/TR/html-aria/#docconformance),  `<strong>` has an implicit role of `strong` but it does not get found

### LINKED ISSUE #1234: ...ByRole with `paragraph` not working
- `@testing-library/dom` version: 9.3.0
- Testing Framework and version: vitest 0.28.5
  
- DOM Environment: jsdom 21.1.0
  

### Relevant code or config:

```js
render(<div><p>blah</p></div>);
console.log(screen.queryAllByRole('paragraph'));
```

### What you did:

Just searching for paragraphs in any rendered HTML, will fail. How the HTML gets rendered (React in this particular case) I don't think that matters anything.

### What happened:

It won't find any elements by the role `paragraph`. Other roles are found successfully.

### Reproduction:

https://codesandbox.io/s/react-testing-library-demo-forked-z4mf2z?file=/src/__tests__/hello.js

I've made the Hello component output a `<p>` and looked it up using `queryAllByRole('paragraph')`, which returns an empty array. I would expect an array of one item, namely the paragraph element.

### Problem description:

I need to test if a paragraph, when certain conditions are met, is present or absent in the resulting render. The [documentation on `...ByRole`](https://testing-library.com/docs/queries/byrole) links to [the specification on W3C](https://www.w3.org/TR/html-aria/#docconformance) where the `p` element is listed has having the `paragraph` role. So, it should work.

### Suggested solution:

I've been digging around to figure out how the role of a given element is calculated, and to me it seems rather over-engineered. There's probably a good reason for it though, but I did end up at [aria-query](https://github.com/A11yance/aria-query) from where you seem to be fetching a list of what appears to be a mapping between elements and roles?

Ultimately I couldn't figure out how and where this list is getting built. I don't mean to patronise, but it seems pretty basic to me, having a map of tagName=>role floating somwhere. It could be the heat wave hitting my brain though 😣

### LINKED ISSUE #1239: Upgrade `aria-query` to version 5.2.1

`aria-query` has released a new version yesterday which caused some regression on our end.
At the moment, we've pinned the version to prevent issues, we need to upgrade it soon to be compliant with the new ARIA spec changes.
Affected issues that were closed in the meantime because we've pinned the version:
https://github.com/testing-library/dom-testing-library/issues/1235
https://github.com/testing-library/dom-testing-library/issues/1238

We need to upgrade locally, see which tests are failing and why, if `aria-query` has issues, we need to open issues in their repo.

jlp-craigmorten:
**What**: Upgrade `aria-query` to version 5.3.0.

**Why**: Receive latest specification updates for role / element mappings to align with WAI ARIA 1.2.

Fixes #1100 
Fixes #1150
Fixes #1201 
Fixes #1234
Fixes #1239

**How**: Upgrading version in `package.json`. Updating tests to mirror changes.

**Checklist**:

- [ ] Documentation added to the - N/A
- [x] Tests
- [x] Ready to be merged
      

---

There are a couple changes of note as a result of this upgrade, namely:

- The `menuitem` element no longer has associated roles. This also removes the only example of an element with two role mappings.
- Many elements such as `p` and `em` now have dedicated roles (`paragraph` and `emphasis` respectively) inline with [HTML-ARIA](https://www.w3.org/TR/html-aria/) / [WAI-ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/). Associated tests have been updated / modified to suit.
- The introduction of the `generic` role also results in some changes. E.g. anchor tags with no `href` are now `generic` under [HTML-ARIA](https://www.w3.org/TR/html-aria/).

There are several other role / element mapping changes as a result of this change to align with specifications.

Given the number of role additions, removals, and modifications (including changes to role and/or changes to the constraints for a particular role) this should be considered a breaking change.


MatanBobi:
Hi @jlp-craigmorten!
Thanks for taking the time to create this and the corresponding fix in `aria-query`, this is extremely appreciated.
I'll try to look into the changes later on today, in the meantime, I pushed a change to fix the CI issues we had.

cmorten:
> Hi @jlp-craigmorten!
> Thanks for taking the time to create this and the corresponding fix in `aria-query`, this is extremely appreciated.
> I'll try to look into the changes later on today, in the meantime, I pushed a change to fix the CI issues we had.

Appreciate it - I’ve just raised https://github.com/testing-library/dom-testing-library/issues/1242 r.e. the CI issue

src/__tests__/__snapshots__/role-helpers.js.snap
```diff
@@ -52,32 +52,6 @@ Name "":
   data-testid="a-article"
 />
 
---------------------------------------------------
-command:
-
-Name "":
```
MatanBobi, line 58: I remember that `menuitem` has been deprecated, I'm hesitating whether this is a breaking change on our end or not. People who used `menuitem` before will now not get `menuitem` role or `command` role.
Any thoughts about this? If we decide to not push it as a major (and I don't think we should), we should clearly state that this has been deprecated due to the ARIA spec.

src/__tests__/role-helpers.js
```diff
@@ -19,7 +19,7 @@ function setup() {
 <header data-testid="a-header">Banner header</header>
 <section aria-label="a region" data-testid='named-section'>
   <a href="http://whatever.com" data-testid="a-link">link</a>
-  <a>invalid link</a>
+  <a data-testid="invalid-link">invalid link</a>
```
MatanBobi, line 22: Maybe we should add a test for `img` role with `alt` value (role img) and with an empty `alt` (role presentation) to see that we get the corresponding role for each. I know that this is not logic that we own but I believe it will help us to understand breakage later on.
Wdyt?

src/__tests__/suggestions.js
```diff
@@ -196,9 +196,9 @@ test('escapes regular expressions in suggestion', () => {
   const {container} = renderIntoDocument(`
       <label for="superInput">inp-t lab^l w{th c+ars to esc\\pe</label>
       <input id="superInput" type="text" value="my super string +-('{}^$)" placeholder="should escape +-'(/" />
-      <p>
+      <span>
```
MatanBobi, line 199: Can you please explain why we needed this change? :)

MatanBobi (commented):
That looks great, thanks :)
I left a few comments.

src/__tests__/suggestions.js
cmorten: The `p` element here is used exclusively (or at least the choice of element) for a test asserting on the outcome of:

```js
getSuggestedQuery(container.querySelector('p')
```

to check that the suggestion properly escapes regular expressions.

ARIA 1.2 introduced a `paragraph` role which is the implicit role for the `p` element. With the bump of `aria-query` which is now tracking 1.2 (and some 1.3 features) this test started to fail as the suggestion became (the more natural) `getByRole("paragraph")`.

I’m not familiar with the history of why this particular assertion exists, but am guessing the handling of `.` and `()` in the suggestion generation was the focus over the specific element choice, thus opted for `span`.

TBH I was surprised the `span` worked as these now have an implicit role of `generic` which I’m fairly certain `aria-query` implements. Would need to dig into the code to see how the role detection works - if it is ignoring `generic` or using `dom-accessibility-api` (see https://github.com/eps1lon/dom-accessibility-api/pull/937) then it not suggesting `generic` makes sense.

In light of this, it’s probably safe to say this test may need changed again in future if/when `generic` roles are fully supported.

An alternative idea is to keep the `p` and instead opt to introduce another `p` element somewhere to force the suggestion to be more specific that just getting by role.

WDYT?

src/__tests__/role-helpers.js
jlp-craigmorten: Agreed, good to have some coverage on this.

Interestingly adding the test cases highlighted an issue - a combination of the empty value issue (see also https://github.com/testing-library/dom-testing-library/pull/1236) as well as a limitation in the "specificity" sorting algorithm. I've pushed amends for both.

src/__tests__/suggestions.js
jlp-craigmorten: Ah it seems there is `generic` specific logic, so think `span` could be long-lived here safely, see https://github.com/testing-library/dom-testing-library/blob/main/src/suggestions.js#L82

src/__tests__/role-helpers.js
MatanBobi: I'm not sure if this was an issue before or that's a new `aria-query` feature and if so then we can understand why it won't work to begin with.

src/__tests__/__snapshots__/role-helpers.js.snap
jlp-craigmorten: Yeah I'm not sure how best to proceed. The situation as I understand it:

- The `menuitem` element is deprecated. REF: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menuitem and https://caniuse.com/?search=menuitem
- No browsers other than Firefox ever supported the element officially.
- Firefox stopped supporting the element in Firefox 85. For perspective, the latest current version of Firefox is 114 and the minimum supported version by `dom-testing-library` [is 91 according to the browserslist config](https://github.com/testing-library/dom-testing-library/blob/main/package.json#L34).
- In WAI ARIA 1.1, pre deprecation of the element, the role `menuitem` and the abstract role `command` were associated with the element. REFs: https://www.w3.org/TR/wai-aria-1.1/#menuitem and https://www.w3.org/TR/wai-aria-1.1/#command
- In WAI ARIA 1.2 the `menuitem` element and it's prior role associations were dropped prior to 2 March 2021 Candidate Snapshot (REF: https://www.w3.org/TR/2021/CR-wai-aria-1.2-20210302/)
- In HTML-ARIA the `menuitem` and `command` roles for the `menuitem` element were dropped from the element in 30 May 2019 working draft (https://www.w3.org/TR/2019/WD-html-aria-20190530/) and the element was later removed altogether before reaching the candidate stage of the spec proposal.

Ideally no-one should ever really used this element, and they certainly shouldn't be now. For most standards this element has either been dropped years ago, or never officially recognised.

I suspect it would be reasonably safe to go for a minor change, with a small risk of there being complaint. A major would be a certain measure, but feels potentially overkill for something that has never been supported by the standard suite of major browsers.

src/__tests__/role-helpers.js
```diff
@@ -78,13 +74,15 @@ function setup() {
...
     h1: getByTestId('a-h1'),
     h2: getByTestId('a-h2'),
     h3: getByTestId('a-h3'),
+    unnamedImg: getByTestId('a-img-1'),
+    presentationImg: getByTestId('a-img-2'),
+    namedImg: getByTestId('a-img-3'),
     nav: getByTestId('a-nav'),
     article: getByTestId('a-article'),
-    menuItem: getByTestId('a-menuitem-1'),
-    menuItem2: getByTestId('a-menuitem-2'),
```
eps1lon, line 87: Can we add new stuff at the end here? Makes the snapshot changes easier to review.

src/__tests__/role-helpers.js
jlp-craigmorten: I've just trialled the empty `alt=""` test case on `main` and it currently reports `img` for the role so I believe it's an existing issue with this lib rather than a consequence of the `aria-query` bump (as the old `5.1.3` version of `aria-query` had this ruleset, `5.2.0` regressed it).

package.json
```diff
@@ -35,7 +35,7 @@
     "ios_saf 12.2-12.5",
     "kaios 2.5",
     "op_mini all",
-    "op_mob 64",
+    "op_mob 73",
```
eps1lon, line 38: Can we pin browserslist/caniuselite first and then decide later what to do here? It's technically a breaking change although it might not be one in practice. But these changes should not be mixed in with other changes.

eps1lon (changes requested):

src/role-helpers.js
```diff
@@ -94,14 +95,32 @@ function buildElementRoleList(elementRolesMap) {
   }
 
   function getSelectorSpecificity({attributes = []}) {
-    return attributes.length
+    return {
+      primaryKey: attributes.length,
+      secondaryKey: attributes.reduce((acc, {value, constraints = []}) => {
```
eps1lon, line 100: Could you elaborate about this part a bit more? Ideally some links to how CSS selector specificity is supposed to work by spec.

jlp-craigmorten: I'm not aware of there being any spec based rules around CSS selector specificity, but happy to look around.

Do you know why there was the choice made for sorting by the number of attributes associated with the role assignment was chosen originally?

---

The issue currently is regarding the `img` element use-cases (and likely other similarly "complex" role assignments). `aria-query` represents each of the three `img` cases with a single attribute object where there is either a constraint for `undefined`, a constraint for `set`, or a `value` of `""`.

The issue arises with the `set` and the `""` cases. Through only sorting by length this lib is inheriting it's specificity ordering for the scenario where attribute lengths are equal somewhat arbitrarily from how the `roles.json` happens to be constructued in `aria-query`.

For a lot things this is fine, but it so happens in this scenario the `set` object is ordered prior to the `""`, and consequently gets evaluated first. The presence of an attribute is a superset of an attribute having a value of `""` so the role `img` is incorrectly applied.

Thinking on it now, the introduction here of a secondary sorting key is potentially a sticking plaster over `makeElementSelector` not being quite on the mark - I might spike seeing if can be more precise there instead.

jlp-craigmorten: I've reverted and opted for an update to `makeElementSelector` that feels better - see [cab934b](https://github.com/testing-library/dom-testing-library/pull/1241/commits/cab934b9e297d9ae782131009c2d84060c3eeb61).

package.json
jlp-craigmorten: FYI @MatanBobi I've reverted your change to this PR so this can be resolved separately.

This will mean this PR is blocked until resolved as the CI (and local install / tests) will fail until resolved.

jlp-craigmorten: Raised https://github.com/testing-library/dom-testing-library/pull/1243 as suggested

src/__tests__/__snapshots__/role-helpers.js.snap
jlp-craigmorten: This said, given the number of other changes this will bring to testing-library, the PR as a whole is hard to not consider as a breaking change - WAI ARIA 1.1 to 1.2 is not backward compatible IMO.

src/role-helpers.js
eps1lon: Yeah that reads a lot better, thanks!

eps1lon (approved):
Looks good to me. @MatanBobi good for you?

MatanBobi (approved):
This looks great to me. Thanks a lot for the work :)
@eps1lon do you think we should make this a breaking change?

nickmccurdy:
My understanding of SemVer is that we technically don’t need to release a breaking change when updating our API support to fix bugs because we don’t document supporting a specific version of ARIA (unless this affects `isInaccessible`, in which case we would need to be a breaking change).

MatanBobi:
> My understanding of SemVer is that we only need to publish breaking changes in a major release if we upgrade to a version of a spec that's no longer back compatible with a version that we've documented supporting.
> 
> Here are the specification versions we link to, organized by our APIs:
> 
> API	Specification
> `ByRole` option `description` and accessible name	Accessible Name and Description Computation 1.1
> `ByRole` option `description`	Accessible Rich Internet Applications 1.1
> `ByRole` table of roles	Accessible Rich Internet Applications 1.2
> `isInaccessible`	Accessible Rich Internet Applications 1.2

Thanks @nickmccurdy!
Where were these documented? I'm not sure they are correct.
The thing is, we have some role definition that have changed in 1.2, so people might look for `role="img"` for example and not get the elements they used to get.

package.json
```diff
@@ -62,7 +62,7 @@
     "@babel/code-frame": "^7.10.4",
     "@babel/runtime": "^7.12.5",
     "@types/aria-query": "^5.0.1",
-    "aria-query": "5.1.3",
+    "aria-query": "^5.3.0",
```
timdeschryver, line 65: 👋 Hey, sorry for my late "review" but since this isn't merged yet I wanted to give my 2cents.

I do think this should be a breaking change because some of the role are changed, which can result in failing tests after the upgrade.

That's why I also propose to pin the `aria-query` version.

```suggestion
    "aria-query": "5.3.0",
```

timdeschryver:
@MatanBobi that would mean to drop support for Node v14, right?
I like that, because also end of life.

MatanBobi:
> @MatanBobi that would mean to drop support for Node v14, right?
> 
> I like that, because also end of life.

Yes, definitely. In this major we'll also drop support for node 14 with these changes.

@eps1lon do you have any concerns?

MatanBobi:
Hi @jlp-craigmorten :)
Thanks again for you work on this one.
Since we're creating a new major version (due to the fact we're dropping support for node 14 and 16), we'll probably get this into `alpha` version first.
We're trying to gather a list of the changes for the change log, do you happen to know if there's a full list somewhere?

Thanks again for the effort you've put into this one!

nickmccurdy:
> Thanks @nickmccurdy! Where were these documented? I'm not sure they are correct. The thing is, we have some role definition that have changed in 1.2, so people might look for `role="img"` for example and not get the elements they used to get.

I was referring to the docs:
- [Core API ➡️ Queries ➡️ `byRole` API](https://testing-library.com/docs/queries/byrole#api)
- [Core API ➡️ Queries ➡️ `byRole` option `description`](https://testing-library.com/docs/queries/byrole)
- [Advanced ➡️ Accessibility ➡️ `isInaccessible`](https://testing-library.com/docs/dom-testing-library/api-accessibility/#isinaccessible)

I'll also add links to my comment above to add context. That being said it seems like we've already decided on a major version anyway, which I support.

nickmccurdy:
If you merge in from `alpha` CI should be fixed.

MatanBobi:
> I was referring to the docs, I've added links above to give more context.
> 

Thanks @nickmccurdy, we'll probably need to update the places where we direct to 1.1 as part of the major release :)

jlp-craigmorten:
> Hi @jlp-craigmorten :) Since we're creating a new major version (due to the fact we're dropping support for node 14 and 16), we'll probably get this into `alpha` version first. We're trying to gather a list of the changes for the change log, do you happen to know if there's a full list somewhere?
> 
> Thanks again for the effort you've put into this one!

Just done a scan through [the aria-query changeset from `5.1.3` to `5.3.0`](https://github.com/A11yance/aria-query/compare/v5.1.3...v5.3.0) and believe these are the pertinent changes:

- feat: blockquote element now has an implicit blockquote role
- fix: input element of type checkbox with a set aria-pressed attribute no longer has an implicit button role
- fix: summary element with aria-expanded set to false no longer has an implicit button role
- fix: summary element with aria-expanded set to true no longer has an implicit button role
- feat: caption element now has an implicit caption role
- feat: code element now has an implicit code role
- fix: th element now has implicit columnheader role regardless of scope attribute
- fix: select element with undefined multiple attribute and a size attribute with value 1 is no longer has an implicit combobox role
- feat: menuitem element no longer has an implicit command role
- feat: aside element with a set aria-label attribute now has an implicit complementary role
- feat: aside element with a set aria-labelledby attribute now has an implicit complementary role
- feat: del element now has an implicit deletion role
- fix: doc-backlink element now get's it's accessible name from contents
- fix: body element no longer has an implicit document role
- fix: html element now has an implicit document role
- feat: em element now has an implicit emphasis role
- feat: a element with no href attribute now has an implicit generic role
- feat: area element with no href attribute now has an implicit generic role
- feat: aside element with no label attribute now has an implicit generic role
- feat: b element now has an implicit generic role
- feat: bdo element now has an implicit generic role
- feat: body element now has an implicit generic role
- feat: data element now has an implicit generic role
- feat: footer element not scoped to the body element now has an implicit generic role
- feat: header element not scoped to the body element now has an implicit generic role
- feat: hgroup element now has an implicit generic role
- feat: i element now has an implicit generic role
- feat: pre element now has an implicit generic role
- feat: q element now has an implicit generic role
- feat: samp element now has an implicit generic role
- feat: section element with no label attribute now has an implicit generic role
- feat: small element now has an implicit generic role
- feat: span element now has an implicit generic role
- feat: u element now has an implicit generic role
- feat: address element now has an implicit group role
- feat: ins element now has an implicit insertion role
- fix: a element now requires a set href attribute to have an implicit link role
- fix: area element now requires a set href attribute to have an implicit link role
- fix: link element with an href attribute no longer has an implicit link role
- fix: select element no longer requires a set multiple attribute to have an implicit listbox role if it has a size attribute with value greater than 1
- feat: mark element now has an implicit mark role
- feat: menuitem element no longer has an implicit menuitem role
- feat: meter element now has an implicit meter role
- feat: p element now has an implicit paragraph role
- fix: img element with an empty alt attribute now has an implicit presentation role
- fix: frame element no longer has an implicit region role
- fix: rel element no longer has an implicit roletype role
- feat: strong element now has an implicit strong role
- feat: sub element now has an implicit subscript role
- feat: sup element now has an implicit superscript role
- feat: time element now has an implicit time role

Feel free to condense / summarise as feel is best!

nickmccurdy:
I don't seem to be getting any roles for `summary`. Shouldn't we at least have `button`?

cmorten:
> I don't seem to be getting any roles for `summary`. Shouldn't we at least have `button`?

See #1252

MatanBobi:
Since we're following the spec with this one, I think it's safe to at least merge to `alpha` so I'm merging this one :)

package.json
```diff
@@ -62,7 +62,7 @@
     "@babel/code-frame": "^7.10.4",
     "@babel/runtime": "^7.12.5",
     "@types/aria-query": "^5.0.1",
-    "aria-query": "5.1.3",
+    "aria-query": "5.3.0",
```
ogonkov, line 65: Why `aria-query` version hardly locked? As for not event `5.3.2` version cause package duplication in deps tree

PR #1248, opened 2023-08-05
feat(role-helpers): add some tests and refine TS types"

naorpeled:
- initial commit
- add tests

**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/role-helpers.js
+++ b/src/__tests__/role-helpers.js
@@ -3,6 +3,7 @@ import {
   logRoles,
   getImplicitAriaRoles,
   isInaccessible,
+  prettyRoles,
 } from '../role-helpers'
 import {render} from './helpers/test-utils'
 
@@ -22,6 +23,15 @@ function setup() {
   <a>invalid link</a>
 
   <nav data-testid='a-nav' />
+  
+  <!-- div without an explicit role is given a "generic" role by default -->
+  <div aria-hidden="true" data-testid="a-hidden-div-with-a-generic-role">
+    <span>Some hidden content</span>
+  </div>
+  
+  <div aria-hidden="true" data-testid="a-hidden-div-without-a-generic-role" role="region">
+    <span>Some hidden content</span>
+  </div>
   
   <h1 data-testid='a-h1'>Main Heading</h1>
   <h2 data-testid='a-h2'>Sub Heading</h2>
@@ -107,6 +117,10 @@ function setup() {
     dt: getByTestId('a-dt'),
     dd: getByTestId('a-dd'),
     header: getByTestId('a-header'),
+    hiddenDivWithGenericRole: getByTestId('a-hidden-div-with-a-generic-role'),
+    hiddenDivWithoutGenericRole: getByTestId(
+      'a-hidden-div-without-a-generic-role',
+    ),
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/__tests__/role-helpers.js
```diff
@@ -23,6 +24,15 @@ function setup() {
 
   <nav data-testid='a-nav' />
   
+  <!-- div without an explicit role is given a "generic" role by default -->
+  <div aria-hidden="true" data-testid="a-hidden-div-with-a-generic-role">
+    <span>Some hidden content</span>
+  </div>
+  
+  <div aria-hidden="true" data-testid="a-hidden-div-without-a-generic-role" role="region">
```
MatanBobi, line 32: AFAIR, `aria-hidden` isn't the only flag that signals a hidden element, it can also be the attribute `hidden`. Maybe we should also test for that?

MatanBobi (commented):
Thanks for this one @naorpeled!
Left a minor comment.

src/__tests__/role-helpers.js
naorpeled: Sounds good to me,
pushed it :)

src/role-helpers.js
```diff
@@ -186,7 +186,7 @@ function getRoles(container, {hidden = false} = {}) {
     }, {})
 }
 
-function prettyRoles(dom, {hidden, includeDescription}) {
+function prettyRoles(dom, {hidden, includeDescription} = {}) {
```
MatanBobi, line 189: Is this change necessary as part of this commit? :)

naorpeled: When invoking prettyRoles in tests without this change,
I'm getting the following error:

image

types/role-helpers.d.ts
```diff
@@ -1,9 +1,19 @@
-export function logRoles(container: HTMLElement): string
+export function logRoles(
+  container: HTMLElement,
+  options?: LogRolesOptions,
+): string
+
+type LogRolesOptions = {
```
timdeschryver, line 6: I think we use `interface`s in other locations.
Or is there a reason why this is a `type`?

```suggestion
interface LogRolesOptions {
```

src/__tests__/role-helpers.js
```diff
@@ -23,6 +24,27 @@ function setup() {
 
   <nav data-testid='a-nav' />
   
+  <!-- div without an explicit role is given a "generic" role by default -->
```
timdeschryver, line 27: To keep it consistent with the other cases, I wouldn't add a comment here to describe the test because it isn't added for the others.

```suggestion
```

timdeschryver (commented):
Thanks for the PR @naorpeled.
Because this PR contains 2 changes (types, and additional tests) I think it would be better to also create 2 focused PR.
This makes it easier to follow, and we have more control over the releases/history.

src/role-helpers.js
timdeschryver: Because this is just for internal usage I don't think this change is needed.

naorpeled:
> Thanks for the PR @naorpeled.
> Because this PR contains 2 changes (types, and additional tests) I think it would be better to also create 2 focused PR.
> This makes it easier to follow, and we have more control over the releases/history.

Fair point.
Will do.

types/role-helpers.d.ts
naorpeled: The reason was that I couldn't think of a reason to allow interface inheritance with that type, but it's not super crucial.
Will change it tomorrow.

src/__tests__/role-helpers.js
naorpeled: I personally find this case more confusing and I'd rather reduce the "Wait what??" moments as much as I can.

I don't mind removing it, just giving my 2 cents :P

naorpeled:
@timdeschryver thanks for the great feedback.
I'm heading to sleep, will make the adjustments tomorrow 🙏

src/role-helpers.js
naorpeled: done.

naorpeled:
Hey @timdeschryver,
I've updated the PR based on your CR.
Also opened [this PR](https://github.com/testing-library/dom-testing-library/pull/1249) for the type defs.

timdeschryver (approved):
LGTM.
Sorry, I missed the notification

naorpeled:
> LGTM.
> Sorry, I missed the notification

All good :)
Thanks!

Issue #1242, opened 2023-06-25
CI failing with op_mob browserslist error


cmorten:
- `@testing-library/dom` version: latest
- Testing Framework and version: N/A
- DOM Environment: N/A
  

### Relevant code or config:

N/A

### What you did:

Run `npm run setup -s` locally or in CI.

### What happened:

```console
$ kcd-scripts build  --no-ts-defs --ignore "**/__tests__/**,**/__node_tests__/**,**/__mocks__/**" && kcd-scripts build --no-ts-defs --bundle --no-clean
Error [BrowserslistError]: Unknown version 64 of op_mob
```

### Reproduction:

`npm run setup -s`

### Problem description:

op_mob 64 is not supported, as noted in https://github.com/testing-library/dom-testing-library/pull/1170#issuecomment-1595300619.

### Suggested solution:

Move to use `op_mob 73` provided it doesn’t result in a breaking change, otherwise a different config to unblock build.

MatanBobi:
Hi @cmorten :)
I already pushed a fix to your PR for this one, I saw this before and didn't get a chance to fix it.

eps1lon:
Please fix this in a different PR so that we can better bisect/revert later.

MatanBobi:
I'm re-opening this because what we have at the moment is a workaround (we've pinned the version) :)

MatanBobi:
This issue is more urgent than we thought because our workaround fails our node 14 CI because npm 6 doesn't support `overrides`.

nickmccurdy:
So #1255 should fix this, right?

MatanBobi:
> So #1255 should fix this, right?

It should fix the failure but it's still a workaround. We should probably update the browsers we support at some point I guess.

nickmccurdy:
That's just a matter of updating `browserslist`, right? Since we're doing a major anyway, let's include a fix then.

MatanBobi:
> That's just a matter of updating `browserslist`, right? Since we're doing a major anyway, let's include a fix then.

Yes, that's a really good point. I'll do it and create a PR to the alpha branch.

nickmccurdy:
This issue seems to only reproduce with Node 16, which is no longer officially supported. Please upgrade to Node 18+ and npm 7+ (if used) and then reinstall dependencies.

We can leave this open for visibility, but it will be fixed in DTL 10. You can also install the `alpha` channel now, but you'll still need to upgrade Node and npm (if used).

[...]

PR #1243, opened 2023-06-26
chore: pin browserslist deps

jlp-craigmorten:
**What**:

Pin `browserslist` and `caniuse-lite` development dependencies.

Fixes #1242

**Why**:

There has been minor release of browserslist packages that remove support of `op_mob 64` which this library currently makes use of. Updates to the support matrix are considered a breaking change, so by pinning dependencies we can unblock minor and patch changes while pending a major release to update these dependencies and the version support for opera mobile.

**How**:

Added `browsers@4.21.8` and `caniuse-lite@1.0.30001502` to the `package.json` `overrides` object (for Node >=16) and `devDependencies` (for Node 14).

Versions chosen to align with those used in the last commit to `main` on 14 June 2023.

**Checklist**:

~- [ ] Documentation added to the~
      ~[docs site](https://github.com/testing-library/testing-library-docs)~
~- [ ] Tests~
~- [ ] TypeScript definitions updated~
- [x] Ready to be merged

PR #1255, opened 2023-09-05
feat: Drop support for Node.js 14.x and Node.js 16.x and add support for Node 20.x

MatanBobi:
BREAKING CHANGE: Minimum supported Node.js version is now 18.x

**What**: This PR drops support for Node 14 and Node 16.

**Why**: Node 14 was declared EOL on May 23, Node 16 will be declared EOL on September 11.

**How**: Upgrade all relevant places.

**Checklist**:

~~- [ ] Documentation added to the
      [docs site](https://github.com/testing-library/testing-library-docs)~~
- [X] Tests
~~- [] TypeScript definitions updated~~
- [X] Ready to be merged

PR #1257, opened 2023-09-10
feat: update browserslist and remove overrides workaround

MatanBobi:
**What**: Resolves https://github.com/testing-library/dom-testing-library/issues/1242. This updates the `browserslist` entry in our package.json.

**Why**: As part of our major version, we're updating the `browserslist` support.

**How**: Ran `npx browserslist defaults` and updated the browserslist support in our package.json file. I've also removed the temporary workarounds introduced in https://github.com/testing-library/dom-testing-library/pull/1243.

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,22 +24,25 @@
     "node": ">=18"
   },
   "browserslist": [
-    "and_chr 103",
-    "and_ff 101",
-    "and_qq 10.4",
-    "and_uc 12.12",
-    "android 103",
-    "chrome 102",
-    "edge 102",
-    "firefox 91",
-    "ios_saf 12.2-12.5",
+    "and_chr 109",
+    "and_ff 109",
+    "and_qq 13.1",
+    "and_uc 13.4",
+    "android 109",
+    "chrome 108",
+    "edge 108",
+    "firefox 102",
+    "ios_saf 16.0",
+    "ios_saf 15.6",
+    "ios_saf 14.5-14.8",
     "kaios 2.5",
     "op_mini all",
-    "op_mob 64",
-    "opera 88",
-    "safari 15.5",
-    "samsung 17.0",
-    "samsung 16.0",
+    "op_mob 73",
+    "opera 94",
+    "safari 16.2",
+    "safari 15.6",
+    "samsung 19.0",
+    "samsung 18.0",
     "node 18.0"
   ],
   "scripts": {
@@ -70,18 +73,12 @@
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^5.11.6",
-    "browserslist": "4.21.8",
-    "caniuse-lite": "1.0.30001502",
     "jest-in-case": "^1.0.2",
     "jest-snapshot-serializer-ansi": "^1.0.0",
     "jest-watch-select-projects": "^2.0.0",
     "jsdom": "20.0.0",
     "kcd-scripts": "^13.0.0",
     "typescript": "^4.1.2"
-  },
-  "overrides": {
-    "browserslist": "4.21.8",
-    "caniuse-lite": "1.0.30001502"
   },
   "eslintConfig": {
     "extends": [


cmorten (approved):

nickmccurdy:
Alternatively can we set the `browserslist` to `defaults` (or something including Node) and pin it to prevent breakage?

nickmccurdy (changes requested):
Alternatively can we set the `browserslist` to `defaults` (or something including Node) and pin it to prevent breakage?

MatanBobi:
> Alternatively can we set the `browserslist` to `defaults` (or something including Node) and pin it to prevent breakage?

Doing that will require us to use `overrides` since `browserslist` is not our direct dependency, it's from `kcd-scripts`. This means we'll need to maintain that `overrides`, that's why I prefer having it hard-coded here.
Wdyt?

timdeschryver (approved):
A green build 🥳
I also prefer this, as its more flexible imho.

nickmccurdy:
I don't think we should remove the `overrides`: The bug fix in this PR is unnecessary because it only targets npm versions installed by Node 16, which are no longer supported in the `alpha` branch as of #1255.

Therefore we should be able to switch to `defaults` (or something including Node) and automate updating the Browserslist query with Dependabot, rather than manually running a command and reviewing its results ever time we need to update targets.

package.json
```diff
@@ -24,22 +24,25 @@
...
-    "samsung 17.0",
-    "samsung 16.0",
+    "op_mob 73",
+    "opera 94",
+    "safari 16.2",
+    "safari 15.6",
+    "samsung 19.0",
+    "samsung 18.0",
     "node 18.0"
   ],
```
nickmccurdy, line 47: This could be simplified to `"browserslist": "defaults"` (or something including Node)

package.json
```diff
@@ -70,19 +73,13 @@
...
     "jest-snapshot-serializer-ansi": "^1.0.0",
     "jest-watch-select-projects": "^2.0.0",
     "jsdom": "20.0.0",
     "kcd-scripts": "^13.0.0",
     "typescript": "^4.1.2"
   },
-  "overrides": {
-    "browserslist": "4.21.8",
-    "caniuse-lite": "1.0.30001502"
-  },
```
nickmccurdy, line 85: Can be reverted

nickmccurdy (changes requested):

package.json
```diff
@@ -70,19 +73,13 @@
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^5.11.6",
-    "browserslist": "4.21.8",
-    "caniuse-lite": "1.0.30001502",
```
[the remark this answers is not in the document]
MatanBobi: This was just a "hack" to make `node 14` build pass but it didn't work so I think either way we won't need it.

package.json
MatanBobi: The thing is that having `"browserslist": "defaults"` isn't really explicit from our end, it relies on the `browserslist` and `caniuse` versions which are an implicit dependency for us.
I prefer having a closed list of browsers for two reasons:
1. We define them explicitly. When using `defaults` we have an implicit broweserslist which will be hard to debug/understand later.
2. We need to use `overrides` which is an `npm` feature. That's fine as an idea but I think that this feature is mostly used to workaround issues and not as a preferred solution.

@timdeschryver @eps1lon do you have any opinion here?

nickmccurdy: Changing my mind now that I realize we've had the hard coded Browserslist longer than the overrides. Not my favorite solution but I think it's a reasonable workaround for now considering we don't commit lockfiles (I'm rethinking if we should still do this but will probably open an issue/pull) and we need a major release anyway.

nickmccurdy (approved):

package.json
timdeschryver: I agree  👍

eps1lon (changes requested):
Let's do this as the last step right before shipping the next major. The browserslist override needs to stay. Otherwise it'll break again.

nickmccurdy:
Can we merge this once we revert the overrides? Since we're already released alphas and this is also targeting the alpha branch, I'd prefer our CI to be passing sooner rather than later.

nickmccurdy (changes requested):
See [my initial review](https://github.com/testing-library/dom-testing-library/pull/1257#issuecomment-1716442602) (excluding the resolved comment) or @eps1lon's review

eps1lon:
CI should be green with https://github.com/testing-library/dom-testing-library/pull/1262.

We'll continue to need the overrides otherwise we'll encounter this issue again.

PR #1262, opened 2023-09-16
chore: Pin NPM version

eps1lon:
Closes https://github.com/testing-library/dom-testing-library/issues/1242

`overrides` is not available in NPM bundled with Node.js 14. But we can hopefully use later NPM versions in older Node.js version.

That way we can keep CI alive without having to make breaking changes.


MatanBobi (approved):
Haven't thought of this idea.
That's a good change to unblock ci.
Thanks :)

PR #1263, opened 2023-09-16
test: Run with Node.js 20

eps1lon:
Current Node.js release that will go into LTS next month


MatanBobi:
This was already done as part of https://github.com/testing-library/dom-testing-library/pull/1255 which is merged to alpha,
is there any specific reason to have another pr?

eps1lon:
To have it in main.

PR #1264, opened 2023-09-16
test: Checking Bun compatibility

eps1lon:
Checking if the repo is compatible with Bun. There's on test assertion that needed changing and I'm pretty sure that's a bug in Bun.

I wonder if all the `kcd-script` calls actually use Bun since they have a shebang that wants Node.js. Type-checking shows only 10% difference:
```bash
$ hyperfine "npx kcd-scripts typecheck --
build types"
Benchmark 1: npx kcd-scripts typecheck --build types
  Time (mean ± σ):      2.384 s ±  0.049 s    [User: 4.826 s, System: 0.204 s]
  Range (min … max):    2.326 s …  2.493 s    10 runs
 
$ hyperfine "bunx kcd-scripts typecheck -
-build types"
Benchmark 1: bunx kcd-scripts typecheck --build types
  Time (mean ± σ):      2.191 s ±  0.082 s    [User: 4.552 s, System: 0.183 s]
  Range (min … max):    2.125 s …  2.401 s    10 runs
```

`bun run format` works locally but not in CI:
```bash
Run bun run format --check --no-write
$ kcd-scripts format --check --no-write
/usr/bin/bash: line [1](https://github.com/testing-library/dom-testing-library/actions/runs/6206581608/job/16850977595?pr=1264#step:4:1): kcd-scripts: command not found
```

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,5 +1,5 @@
 {
   "installCommand": "install:csb",
   "sandboxes": ["github/kentcdodds/react-testing-library-examples"],
-  "node": "14"
+  "node": "20"
 }
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,5 +1,5 @@
 {
   "installCommand": "install:csb",
   "sandboxes": ["github/kentcdodds/react-testing-library-examples"],
-  "node": "20"
+  "node": "18"
 }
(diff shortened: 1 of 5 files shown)


eps1lon:
All the CI steps work locally but not in CI. This is not something I wish to debug. But good to know we can potentially debug Bun compat issues locally.

Issue #1270, opened 2023-10-03
Leaking `waitFor` callback against `jasmine` when the `clock` is mocked


KevinBon:
- `@testing-library/dom` version: 9.3.1
- `@testing-library/react` version 14.0.0
- Testing Framework and version: `jasmine@5.1.0`
- DOM Environment: Chrome (Headful and Headless)

### Relevant code or config:

```js
/**
 * For making dom-testing-library compatible with jasmine
 */
function enableFakeJestMode() {
  /**
   * For jestFakeTimersAreEnabled
   * @see https://github.com/testing-library/dom-testing-library/blob/a7b725257e61370675ee043813003d36fae23e9d/src/helpers.ts#L5
   * For usingJestFakeTimers
   * @see https://github.com/testing-library/dom-testing-library/blob/a7b725257e61370675ee043813003d36fae23e9d/src/wait-for.js#L51-L52
   *
   */
  Object.assign(globalThis.setTimeout, { _isMockFunction: true });

  globalThis.jest = {
    /**
     * Use jasmine.clock instead
     */
    advanceTimersByTime: (delay: number) => {
      globalThis.jasmine.clock().tick(delay);
    },
  };
}

describe('jasmine', () => {
  beforeEach(() => {
    enableFakeJestMode();
    jasmine.clock().install();
  });

  afterEach(() => {
    console.log(
      '%c afterEach',
      'color: #f80; font-weight: bold',
      'uninstall:before'
    );
    jasmine.clock().uninstall();
    console.log(
      '%c afterEach',
      'color: #f80; font-weight: bold',
      'uninstall:after'
    );
  });

  it('simulates a timeout', async () => {
    render(<span>Will never be found</span>);

    try {
      await waitFor(
        () => {
          console.log(
            '%c spec',
            'color: lime; font-weight: bold',
            'waitfor:callback'
          );
          // The button shouldn't be found, so it will throw/fail the inner waitFor callback predicate
          screen.getByRole('button');
        },
        { interval: 6000 } // to have less logs
      );
      fail('Should have failed');
    } catch (error) {
      console.log('%c spec', 'color: #f80; font-weight: bold', 'finished ✅');
      expect().nothing();
    }
  });
});
```

### What you did:

I added logs to demonstrate that the `waitFor` callback has been called even after it has been resolved from a timeout.

### What happened:

[...]

PR #1271, opened 2023-10-03
fix: Stop calling `waitFor` callback after timeout

KevinBon:
**What**:

Prevent `waitFor` callback from being invoked even after it resolved, on `waitFor` timeout.

**Why**:

More context can be found [in this issue](https://github.com/testing-library/dom-testing-library/issues/1270), but with that context in mind, it's preventing test side-effects:

`afterEach` should be enough to "clean" the previous test from any side-effect.  However, because of the callback leak issue, `window.variable` will remain to be `123` even after being set back to `1` for a short period of time. 

```js
afterEach(() => {
 window.variable = 1;
})
it('my test', () => {
  return waitFor(() => {
    window.variable = 123;
    throw new Error('I want it to timeout');
  })
})
```

**How**:

When the clock is mocked, only call `checkCallback` when `finished` is `false` 

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs): N/A
- [x] Tests
- [ ] TypeScript definitions updated: N/A
- [x] Ready to be merged
      

resolves https://github.com/testing-library/dom-testing-library/issues/1270

### DIFF
--- a/src/__tests__/wait-for.js
+++ b/src/__tests__/wait-for.js
@@ -431,24 +431,19 @@ test(`when fake timer is installed, on waitFor timeout, it doesn't call the call
     },
   })
 
-  let waitForHasResolved = false
-  let callbackCalledAfterWaitForResolved = false
+  const callback = jest.fn(() => {
+    throw new Error('We want to timeout')
+  })
+  const interval = 50
 
   await expect(() =>
-    waitFor(
-      () => {
-        // eslint-disable-next-line jest/no-conditional-in-test -- false-positive
-        if (waitForHasResolved) {
-          callbackCalledAfterWaitForResolved = true
-        }
-        throw new Error('We want to timeout')
-      },
-      {interval: 50, timeout: 100},
-    ),
+    waitFor(callback, {interval, timeout: 100}),
   ).rejects.toThrow()
-  waitForHasResolved = true
+  expect(callback).toHaveBeenCalledWith()
 
-  await Promise.resolve()
+  callback.mockClear()
 
-  expect(callbackCalledAfterWaitForResolved).toBe(false)
+  await jest.advanceTimersByTimeAsync(interval)
+
+  expect(callback).not.toHaveBeenCalledWith()
 })
--- a/src/__tests__/wait-for.js
+++ b/src/__tests__/wait-for.js
@@ -335,10 +335,10 @@ test('does not work after it resolves', async () => {
     },
   })
 
-  let data = null
+  let timeoutResolved = false
   setTimeout(() => {
     contextStack.push('timeout')
-    data = 'resolved'
+    timeoutResolved = true
   }, 100)
 
   contextStack.push('waitFor:start')
@@ -346,7 +346,7 @@ test('does not work after it resolves', async () => {
     () => {
       contextStack.push('callback')
       // eslint-disable-next-line jest/no-conditional-in-test -- false-positive
-      if (data === null) {
+      if (!timeoutResolved) {
         throw new Error('not found')
       }
     },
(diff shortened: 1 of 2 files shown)


eps1lon (changes requested):
Thanks for working on the fix. However, we should rather make sure that `callback` is not being invoked again. The proposed fix just hides the issue.

Also: Can you provide a minimal repro against which we can test this fix? Ideally we could extract a unit test from it.

KevinBon:
@eps1lon I changed the _solution_ implementation.

You can find a reproduction example on [this repository](https://github.com/KevinBon/jasmine-tl-waitfor-leak/tree/master).

Output example:

image

It's been a while since I have used `jest`, so I'm not sure how to test "this behavior" on jest to satisfy the test code coverage

Don't hesitate if you have any other questions 🙇

KevinBon:
I created a branch to show-case that applying this patch will fix this behavior: https://github.com/KevinBon/jasmine-tl-waitfor-leak/pull/1

src/wait-for.js
```diff
@@ -160,6 +163,9 @@ function waitFor(
     }
 
     function handleTimeout() {
+      if (finished) {
```
eps1lon, line 166: Shouldn't we cancel the timeout instead when we finish?

src/wait-for.js
```diff
@@ -134,7 +137,7 @@ function waitFor(
     }
 
     function checkCallback() {
-      if (promiseStatus === 'pending') return
+      if (finished || promiseStatus === 'pending') return
```
eps1lon, line 140: Can we cancel whatever is calling `checkCallback` instead?

eps1lon (changes requested):

src/wait-for.js
KevinBon: When we finish, `onDone` should have been called with `clearTimeout` ([src](https://github.com/testing-library/dom-testing-library/blob/a0bb0050b1ccb7652a60b72d617453400c300837/src/wait-for.js#L113)), therefore I don't really think this one is needed, but it's more for extra-safety.

src/wait-for.js
KevinBon: I would love to, but I'm afraid it might introduce breaking changes.

We'll need to move the `if(finished)` before `checkCallback` here: https://github.com/testing-library/dom-testing-library/pull/1271/files#diff-68584107d5f50eb95702a47b23bb6e0f6c2d78f9bab7456a1a4150c79229db78R84-R92

But the following comment is daunting: https://github.com/testing-library/dom-testing-library/pull/1271/files#diff-68584107d5f50eb95702a47b23bb6e0f6c2d78f9bab7456a1a4150c79229db78R84-R87

KevinBon: Actually https://github.com/testing-library/dom-testing-library/pull/1271/files#diff-68584107d5f50eb95702a47b23bb6e0f6c2d78f9bab7456a1a4150c79229db78R84-R92 is un-related, so I should move the `if(finished)` before `checkCallback` 👍

src/wait-for.js
KevinBon: Amended in https://github.com/testing-library/dom-testing-library/pull/1271/commits/6b41483386804687e145119bc1966a2da711eb01

src/wait-for.js
KevinBon: Amended in https://github.com/testing-library/dom-testing-library/pull/1271/commits/6b41483386804687e145119bc1966a2da711eb01

src/wait-for.js
```diff
@@ -81,15 +81,15 @@ function waitFor(
           jest.advanceTimersByTime(interval)
         })
 
+        // Could have timed-out
+        if (finished) {
```
eps1lon, line 85: With this factoring, I wonder if we could simplify further by moving `checkCallback` before the `advanceTimers`. That way the loop condition makes sure we exit as soon as we're `finished`. And we can remove another `checkCallback` on L54.

At this point I would remove the "It's really important that checkCallback [...]" comment. We might have fixed the issue along the way. And since I'd rather ship this with the next major, we have a good opportunity to resurface the bug so that we can at least link a repro. It doesn't have to be a Jest test but somehow the reordering fixed "something". That "something" needs to be linked a t least. Doesn't need to be an automated test but it needs to be something. Otherwise we're chasing ghosts.

KevinBon: If we move `checkCallback` before the `advanceTimers`, the `waitFor` `callback` will be called twice before we advance the clock, instead of once right now.

Which will break this fix: https://github.com/testing-library/dom-testing-library/pull/1073/files

src/__tests__/wait-for.js
```diff
@@ -292,12 +292,11 @@ test('the real timers => fake timers error shows the original stack trace when c
 
 test('does not work after it resolves', async () => {
   jest.useFakeTimers('modern')
-  let context = 'initial'
+  const contextStack = []
```
KevinBon, line 295: Or

```suggestion
  const contextCallStack = []
```

KevinBon:
@eps1lon I added a reproducible test for `jest` through https://github.com/testing-library/dom-testing-library/pull/1271/commits/ce11c0f9f51f51f1e84cb790016d83f44502ec87

src/__tests__/wait-for.js
```diff
@@ -307,54 +306,149 @@ test('does not work after it resolves', async () => {
...
+      act:end,
+      callback,
+      no-act:end,
+      waitFor:end,
+    ]
+  `)
+})
 
-  expect(context).toEqual('initial')
+test(`when fake timer is installed, on waitFor timeout, it doesn't call the callback afterward`, async () => {
```
KevinBon, line 395: This test ensures that the fix is working

src/__tests__/wait-for.js
```diff
@@ -292,12 +292,11 @@ test('the real timers => fake timers error shows the original stack trace when c
 
 test('does not work after it resolves', async () => {
```
KevinBon, line 293: I made some changes on this test to help understand the outcome: everything that starts needs to end

src/wait-for.js
eps1lon: > the waitFor callback will be called twice before we advance the clock, instead of once right now.

but not if we remove the `checkCallback` call earlier from L54, no?

KevinBon: There will be a change of behavior in https://github.com/testing-library/dom-testing-library/blob/37ccada9ed26dec9a50bee8df3af4ea0e45d61b9/src/__tests__/fake-timers.js#L43, for a `timeout` smaller than the `interval`, and will instead be rejected with `[Error: Timed out in waitFor.]`

KevinBon:
@eps1lon bumping, let me know if you need anything from me. Thank you!

eps1lon:
Thanks for the patience. Checking this out now so that we can land it ASAP.

eps1lon (commented):
Looks great, thank you. I verified the tests are actually failing on `main`

eps1lon (dismissed):
Looks great, thank you. I verified the tests are actually failing on `main`

eps1lon (approved):

eps1lon:
@all-contributors add @KevinBon for code, bugs

Issue #1179, opened 2022-10-17
Specify ontimeout at configuration level


SimonGodefroid:
Is there a way to specify an `onTimeout` function at top level?

We are having some flaky tests and I'd like to be able to print the screen when some of the timeouts occur because we figured in some cases the printed DOM during timeouts is really not what we're expecting.

Context: 

some tests fails every once in a while because we render the app and redirect to a specific route and sometimes 2 tests that run in a row, the second one is not starting at the redirect location at the beginning of the test but at the location where the previous test stopped (when you click on the submit button of a form, mock with msw a patch request and that the code redirects to the details page). For that case we have set the `DEBUG_PRINT_LIMIT` in the CI command to be able to see the mounted DOM and figured that weird case.

Motivation:

We have new flaky tests and this time the error message in the CI is:

```bash
Error: expect(element).toBeInTheDocument()

element could not be found in the document
    at Object.<anonymous> (/home/circleci/project/src/***.test.tsx:39:53)
    at _callCircusTest (/home/circleci/project/node_modules/jest-circus/build/run.js:212:5)
    at _runTest (/home/circleci/project/node_modules/jest-circus/build/run.js:149:3)
    at _runTestsForDescribeBlock (/home/circleci/project/node_modules/jest-circus/build/run.js:63:9)
    at _runTestsForDescribeBlock (/home/circleci/project/node_modules/jest-circus/build/run.js:57:9)
    at run (/home/circleci/project/node_modules/jest-circus/build/run.js:25:3)
    at runAndTransformResultsToJestFormat (/home/circleci/project/node_modules/jest-circus/build/legacy-code-todo-rewrite/jestAdapterInit.js:176:21)
    at jestAdapter (/home/circleci/project/node_modules/jest-circus/build/legacy-code-todo-rewrite/jestAdapter.js:109:19)
    at runTestInternal (/home/circleci/project/node_modules/jest-runner/build/runTest.js:380:16)
    at runTest (/home/circleci/project/node_modules/jest-runner/build/runTest.js:472:34)
``` 

Locally I wrote a purposefully broken statement to see whether I could leverage on `onTimeout` function to print the DOM as it was when the code failed.

```tsx
expect(
      await screen.findByText(
        /LOLLLLLL/i,
        {},
        {
          onTimeout: err => {
            console.log('>>>'.repeat(200),err);
            screen.debug(null, 10000000000);
          }
        }
      )
    ).toBeInTheDocument();
```

[...]

PR #1272, opened 2023-10-11
#1179: Specify default ontimeout at configuration level

Areiser:
**What**:

Make it possible to override the default onTimeout method for all waitFor methods in the config.

**Why**:

We use the testing library in our organization and would like to print a clear user information that the test ran into a timeout, since we sometimes struggle with flaky tests. Only printing the error leads some users to think that the condition did not time out.

This is my first contribution, I added a test, but if there are any change requests I'll gladly update this PR, as the feature would be very helpful for us.


src/__tests__/fake-timers.js
```diff
@@ -43,6 +43,23 @@ test('fake timer timeout', async () => {
   ).rejects.toMatchInlineSnapshot(`[Error: always throws]`)
 })
 
+test('fake timer timeout uses default ontimeout', async () => {
+  configure({
```
timdeschryver, line 47: Since this updates global state, we should also revert it back to it's original state to prevent these changes from leaking through other test cases.
As an example see https://github.com/testing-library/dom-testing-library/blob/0e528c4d805de0e057ad3c7211269077913ebe1c/src/__tests__/element-queries.js#L11

timdeschryver (commented):
This LGTM, but I have a comment about the test case.

src/__tests__/fake-timers.js
Areiser: Thanks for the hint, I have reset the configuration after the tests as in the example you provided.

MatanBobi:
Thanks for this contribution. I'm hesitating whether this is a use-case that should be built-in in the library. Do we believe users will want to configure it on their own? Why not just add a hint that it was caused by a timeout in our codebase and not make it configurable?

timdeschryver:
@MatanBobi both can be useful. 
Modifying the message can be helpful to a broader audience (is this a breaking change)?
But since we already allow that the message can be overwritten for single cases, I think having a global config for it is OK (as with other config properties). It's also a relative small change that doesn't have an impact to a big area.

PR #1273, opened 2023-10-13
feat(ByRole): improve `current` option description

thefalked:
**What**: 

Updating `current` options JSDOC to be more clear how to use it.

**Why**:

Isn't clear reading the docs that we could use another value, even with the `string` type.

**How**: 

Changing the JSDOC description on `types/queries.d.ts` file.

**Checklist**:

- [ ] Tests N/A
- [x] Ready to be merged

### DIFF
--- a/types/queries.d.ts
+++ b/types/queries.d.ts
@@ -96,7 +96,9 @@ export interface ByRoleOptions {
    */
   pressed?: boolean
   /**
-   * Filters elements by their `aria-current` state. `true` and `false` match `aria-current="true"` and `aria-current="false"` (as well as a missing `aria-current` attribute) respectively.
+   * Filters elements by their `aria-current` state.
+   * Note that no `aria-current` attribute will match `current: false` since `false` is the default value for `aria-current`.
+   * https://www.w3.org/TR/wai-aria-1.2/#aria-current
    */
   current?: boolean | string
   /**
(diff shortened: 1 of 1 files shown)


MatanBobi:
Hi @thefalked, thanks for taking the time to create this one.
I don't think that this is better than what we have (based on the type (`boolean | string`) and the current comment.

I'm leaning towards keeping it as it is but I'll leave it open if any other maintainer has a different opinion.
Thanks again.

thefalked:
> Hi @thefalked, thanks for taking the time to create this one. I don't think that this is better than what we have (based on the type (`boolean | string`) and the current comment.
> 
> I'm leaning towards keeping it as it is but I'll leave it open if any other maintainer has a different opinion. Thanks again.

The only reason that i though it was a good idea to change is that in the [docs site](https://testing-library.com/docs/queries/byrole#current) and in the JSDOC description is only talking about the true & false, i spend my hole day trying to find why the test was failing and i had `aria-current="page"`. (I though that `current: true` would validate if the property was on the element independent of the value).

Maybe someone who has less experience like me find this a better solution than what was on the previous description.

MatanBobi:
If that's the case, I think it will be better to update our docs and not the JSDOC comment. Any chance you can open a PR to our docs site for this one? I'm closing this as I highly believe that this should be a part of the doc and not the JSDOC.

Thanks!

PR #1300, opened 2024-04-17
Update package.json

rajagopalpalani:
To support latest react version

**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "chalk": "^4.1.0",
     "dom-accessibility-api": "^0.5.9",
     "lz-string": "^1.5.0",
-    "pretty-format": "^27.0.2"
+    "pretty-format": "^29.7.0"
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^5.11.6",


MatanBobi:
@rajagopalpalani Can you please provide more context into why this change is needed?

lukeapage:
I just coincidentally came across this because typescript performance analyse pointed this out as a duplicate package in node_modules. Its to upgrade so they can be deduped.

MatanBobi:
What's the reason it's duplicated? Is that only for your package? I don't see a reason for this to be duplicated in our repo.

lukeapage:
We depend on testing library and jest and jest is upgraded and testing library is not.

I have no idea if this pr is correct or not but why wouldn’t you want to regularly update to the latest version ?

MatanBobi:
Since we're following semver, we need to understand if this should cause a major upgrade in our library which we just made. I'm not sure and I currently don't have the time to invest in understanding what are the changes in `pretty-format`. If we're exporting some of it's functionalities, this might break our API.

rajagopalpalani:
I thought version was an issue. I am closing this MR. Its an issue from our end

PR #1307, opened 2024-04-27
feat: Allow null type for node on fireEvent and its shortcuts

jdufresne:
**What**:

Allow the `null` type for `node` argument on fireEvent and its shortcuts.

**Why**:

A common pattern in some testing code is:

```js
const btn = document.getElementById('my-button')
fireEvent.click(btn)
```

However, `document.getElementById()` may return `null` and then TypeScript will report an error:

```
error TS2345: Argument of type 'HTMLElement | null' is not assignable to parameter of type 'Document | Node | Element | Window'.
```

To resolve the TypeScript error, library users can assert or check the returned value with something like:

```js
const btn = document.getElementById('my-button')
if (!btn) {
   throw new Error("Unable to find 'my-button')
}
fireEvent.click(btn)
```

But dom-testing-library is already doing this check, so let fireEvent accept null as a convenience to call sites.

The error already reported by dom-testing-library is:

```
Unable to fire a "click" event - please provide a DOM element.
```

Some might suggest that using `document.getElementById()` should be replaced by a dom-testing-library query method, but that is non-trivial on legacy codebases with many tests trying to adopt and introduce dom-testing-library.

**How**:

Augmented the types of fireEvent & friends to allow `null` for the `node` argument.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/events.js
+++ b/src/__tests__/events.js
@@ -420,6 +420,18 @@ test('throws a useful error message when firing events on non-existent nodes (sh
   )
 })
 
+test('throws a useful error message when firing events on null', () => {
+  expect(() => fireEvent(null, new MouseEvent('click'))).toThrow(
+    'Unable to fire a "click" event - please provide a DOM element.',
+  )
+})
+
+test('throws a useful error message when firing events on null (shortcut)', () => {
+  expect(() => fireEvent.click(null)).toThrow(
+    'Unable to fire a "click" event - please provide a DOM element.',
+  )
+})
+
 test('throws a useful error message when firing non-events', () => {
   expect(() => fireEvent(document.createElement('div'), undefined)).toThrow(
     'Unable to fire an event - please provide an event object.',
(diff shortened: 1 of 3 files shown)


eps1lon:
It should error when it receives `null` because that's also a runtime error. If you know `getElementById(elementID)` returns an element, you should use non-null assertions: `fireEvent.click(getElementById(elementID)!)`

eps1lon (changes requested):
It should error at compile time since that's also a runtime error. If you know `getElementById(elementID)` returns an element, use non-null assertions instead: `fireEvent.click(getElementById(elementID)!)`

jdufresne:
For me, that runtime error is desired as that gets converted into a test failure by Jest.

Using the TypeScript `!` is an escape hatch to bypass type checking. Generally speaking, I try to avoid these escape hatches as much as possible and instead adjust the types to better reflect the state and capability of things. 

In this case, the `fireEvent` function is prepared to handle `null`. True, it handles it by throwing an error, but it is handled none-the-less and so, IMO, isn't invalid input from a typing perspective.

My goal is to reduce the burden on the call site within tests as the same expected outcome happens: either continue with the non-null value or fail the test. Asking every call site to add `!` reintroduces a burden (albeit a small one).

If this proposal is against the principles of this library, no worries, feel free to close.

eps1lon:
The principle is that runtime errors should be caught at compile time.

PR #1321, opened 2024-06-25
Fix pretty dom exception

DavidRieman:
Tests and code improvement to avoid unexceptional exceptions from being thrown during `shouldHighlight` in `pretty-dom`.

### DIFF
--- a/src/pretty-dom.js
+++ b/src/pretty-dom.js
@@ -5,25 +5,20 @@ import {getDocument} from './helpers'
 import {getConfig} from './config'
 
 const shouldHighlight = () => {
-  let colors
+  // Try to safely parse env COLORS: We will default behavior if any step fails.
   try {
-    colors = JSON.parse(process?.env?.COLORS)
-  } catch (e) {
-    // If this throws, process?.env?.COLORS wasn't parsable. Since we only
-    // care about `true` or `false`, we can safely ignore the error.
+    const colors = process?.env?.COLORS
+    if (colors) {
+      const b = JSON.parse(colors)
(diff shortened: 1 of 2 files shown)


eps1lon:
Basically we reduce the number of caught exceptions?

DavidRieman:
Indeed, unexceptional exceptions are a thorny problem in the JS ecosystem due to continued lack of sufficient IDE support for the kind of "just my code" debugger halting features that have existed for decades in other ecosystems.  (This one in particular stings me at least once a year so I finally resolved to fix it at the source.)

eps1lon (approved):

src/pretty-dom.js
```diff
@@ -5,25 +5,20 @@ import {getDocument} from './helpers'
...
+    // Ignore (non-critical) - Make a defaulting choice below.
   }
 
-  if (typeof colors === 'boolean') {
-    // If `colors` is set explicitly (both `true` and `false`), use that value.
-    return colors
-  } else {
-    // If `colors` is not set, colorize if we're in node.
-    return (
-      typeof process !== 'undefined' &&
```
eps1lon, line 22: This always needs to stay for environments that don't have `process` e.g. browsers: https://github.com/testing-library/dom-testing-library/pull/1323

eps1lon:
Needs to be reverted in https://github.com/testing-library/dom-testing-library/pull/1325 since it caused https://github.com/testing-library/dom-testing-library/issues/1322. Relanding in https://github.com/testing-library/dom-testing-library/issues/1322

PR #1331, opened 2024-08-01
feat: Add strictVisibility check option to ByRole queries

benlesh:
**What**:

Added an option called `strictVisibilityCheck` that defaults to `true` (for now). If `true` it will use the existing code path that does a `getComputedStyle` on each node to check visibility during queries `ByRole`.  If set to `false`, the visibility check settles for a more lose algorithm that doesn't run `getComputedStyle`.

**Why**:

This is being added as a means of improving performance of tests using queries like `getAllByRole` on larger sets of DOM elements, as can be found in many real-world applications. Anecdotally, using this setting fixed one test I had locally from `8000ms` to `500ms` total. (still not great, but miles better).

**How**:

Added an option to the type definition, threaded it through the existing logic, and added a "loose" check for inaccessibility. I added a test with a few checks to make sure the default is still working as expected, and the new setting still works in a similar manner.

It's worth noting that I tried to add a test that checked "indirect" visibility alterations through CSS classes, however, I was unable to get the test harness to recognize the CSS class and trigger layout (I was probably doing something wrong there, happy to follow up).

**Checklist**:

- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -813,6 +813,37 @@ test('queryAllByRole returns semantic html elements', () => {
   expect(queryAllByRole('listbox')).toHaveLength(1)
 })
 
+test('getAllByRole matchers with strictVisibility enabled and disabled', () => {
+  const {getAllByRole} = render(`
+    <div aria-hidden="true">
+      <button>Aria Hidden Button</button>
+    </div>
+    <div style="display: none;">
+      <button>Display None Button</button>
+    </div>
+    <div style="visibility: hidden;">
+      <button>Visibility Hidden Button</button>
+    </div>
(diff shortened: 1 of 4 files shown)


eps1lon (changes requested):
I don't want to define any custom interpretation what it means to be in the a11y tree. This will be hard to teach and probably change too often depending on the performance needs. Users should use the [`hidden` option](https://testing-library.com/docs/queries/byrole#hidden) instead if they don't want to check if an element is in the a11y tree or not.

benlesh:
`hidden: true` isn't quite accurate to the need, but whatever. Not my library. I'll fork it if I need this consistently, I guess.

Issue #1340, opened 2024-11-19
Replace `chalk` with `picocolors`


Fuzzyma:
### Describe the feature you'd like:

I want to propose to replace the `chalk` dependency with the smaller and faster `picocolors` package.
As part of the [e18e](https://e18e.dev/) initiative (ecosystem cleanup), we try to reduce the burden of old or big libs and modernize the ecosystem.
Considering that chalk is used for exactly **one** style in the console output in this lib, it might be worthwhile to think about geeting rid of it at all :D.
`chalk` alone accounted for a whopping `907.93 GB` of traffic last month whenever it got installed as part of `dom-testing-library`.
The tar size of `13.40Kb` doesn't seem much (`dom-testing-library`s tar is `434.14KB`) but every bit counts :).

### Suggested implementation:

Either use picocolor or inline the one styling you have.
In your case it would be a full drop-in repacement.

I would be willing to create the PR for you

Fuzzyma:
@MatanBobi there you go https://github.com/testing-library/dom-testing-library/pull/1341

I wasnt able to run the tests because I hit an error. Details in the PR

PR #1341, opened 2024-11-19
chore(deps): replace chalk with picocolors

Fuzzyma:
Replaced `chalk` with `picocolors` as discussed in #1340.
I ended up using picocolors to support the NO_COLOR env (https://no-color.org/)

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

I wasnt able to run the tests. It errored with `TypeError: _browserslist.findConfigFile is not a function`.
I dont want to call this mergable unless I can see the green tests.

// EDIT: The CI seems to fail with the same error

// OFFTOPIC: I saw that `@testing-library/jest-dom` is also using chalk but only in a test. However, it is listed as dependency. Maybe we can move it to devDependencies and also replace it?


eps1lon:
>  I saw that @testing-library/jest-dom is also using chalk but only in a test. However, it is listed as dependency. Maybe we can move it to devDependencies and also replace it?

It's listed in `devDependencies` as far as I can tell.

Fuzzyma:
@eps1lon I looked into it and this is a deadlock. I tried several combinations of pinned dependencies and got the tests to work eventually. But now `validate` fails.

I really wonder why nobody ever bothered to commit a lock-file. Thats exactly what it is for :D.

The reason this is all not working is, is because some babel dependencies that are brought in require at least babel 7.22. However, that version calls browserlist with a different api (and therefore fails). Installing older versions of those packages gets rid of that error but now the validate fails with (to me unknown) reason because it complains that a function isnt found (which is clearly there).

So I have to give up for now. What other possibilities to we have?
Does anyway of you have a working setup? If so, please dump all the exact versions of your dependencies by (deleting any shrinkwrap and lock file and then) running

```
npm shrinkwrap
```

Give me that file :D

Fuzzyma:
OMG I DID IT!!!!
I had to commit the package-lock.json to convince the CI to install the correct versions.
So not sure what to do with this.

It would be really beneficial to get rid of old supported browsers so we can update browserslist so we can update babel so we can get rid of all that overwrites and really unstable depenceny situation

.gitignore
```diff
@@ -5,5 +5,4 @@ dist
 
 # these cause more harm than good
 # when working with contributors
-package-lock.json
```
timdeschryver, line 8: > I had to commit the package-lock.json to convince the CI to install the correct versions. So not sure what to do with this.

I might be wrong about this, but we're not a big fan of introducing lock files (https://github.com/testing-library/dom-testing-library/pull/624).
That's why we're also not using it during our CI pipeline:
https://github.com/testing-library/dom-testing-library/blob/a86c54ccda5242ad8dfc1c70d31980bdbf96af7f/.github/workflows/validate.yml#L49

If you want you can revert this change, and I can give it a look in order to fix this.

package.json
```diff
@@ -59,29 +59,33 @@
...
     "aria-query": "5.3.0",
-    "chalk": "^4.1.0",
-    "dom-accessibility-api": "^0.5.9",
-    "lz-string": "^1.5.0",
-    "pretty-format": "^27.0.2"
+    "dom-accessibility-api": "0.5.9",
+    "lz-string": "1.5.0",
+    "picocolors": "1.1.1",
+    "pretty-format": "27.0.2",
+    "typescript": "4.5"
```
timdeschryver, line 70: `typescript` needs to be listed as a `devDependency`, or is there a reason to move it into the dependencies?

timdeschryver (changes requested):
Thanks for the pull request @Fuzzyma !
I did leave some comments, feel free to ask for assistance if needed.

package.json
Fuzzyma: Just an oversight and probably a leftover from trying to get the versions right :D

Fuzzyma:
@timdeschryver I reverted the commit of the package-lock and the ci is red again. So would love some help here. Clearly the CI does _something_ with the lock-file :D

Also moved typescript back to devDependencies where it rightfully belongs

.gitignore
MichaelDeBoey: This originates from @kentcdodds' package starters.

If I remember correctly (but Kent can probably go more in depth here) the main reason is for when different people are adding/removing/updating different (dev)dependencies at the same time.
When merging one PR, the other one could automatically be in conflict with the one just merged, which is a pain to resolve.

MichaelDeBoey: A good solution (for now) would probably be to pin the versions to an exact number and investigate this further in another PR, so we don't hold back this PR

Fuzzyma:
It seems like that there was a to introduce a lockfile in 2022 already but Kent wanted the aproval from one other maintainer (https://github.com/testing-library/dom-testing-library/pull/624#issuecomment-643260300). Maybe that could be you? :D



Kent also seems to use lockfiles now:
https://github.com/kentcdodds/kentcdodds.com

MichaelDeBoey:
@Fuzzyma Kent's not opposed to using lock files in projects/apps, but the way I understood it he is opposed using them in packages

.gitignore
Fuzzyma: That is what I tried exactly. I pinned the versions but the pinned dependencies also have dependencies and they are not pinned. The version I have locally runs for whatever reason. So my set of dependencies and sub dependencies work together.
I spend quite a while to find a set of overwrites and versions that ran at all. So it might take another while to find out what I have installed vs what the ci is installing :D.

Yes, package-lock conflicts are annoying. But npm can handle merge conflicts nowadays and will resolve them for you if you just run "npm install".

We can also move this discussion into its own issue as you suggested. I think it would be worthwhile to decide that first before putting more time into investigating the exact dependencies we need to overwrite :D

timdeschryver:
I will take a look at this in the next days (I might push directly into this branch). 
AFAIK @MichaelDeBoey is right about the reasons of why not to include a lock file, additionally it also allows people to use their package manager of choice. We might revisit this later, but I prefer not to do this within this PR.

timdeschryver (approved):
Pinning `@babel/helper-compilation-targets` seems to do the trick.
I used the last version of a successful build.

MichaelDeBoey (approved):

timdeschryver:
@eps1lon does this seem good to you? 
As mentioned in the comments, `@babel/helper-compilation-targets` was pinned to make the build happy without upgrading deps/changing the browserlist support (as was done in #1346)

MatanBobi (approved):

benmccann:
This has three approvals. Could we merge it?

MatanBobi:
Thanks @benmccann for the ping.
I've just merged https://github.com/testing-library/jest-dom/pull/659, let's see we're not seeing any regressions there and I'll merge this one un the upcoming week :)

PR #1346, opened 2025-01-15
Fix 'TypeError: _browserslist.findConfigFile is not a function'

nifgraup:
**What**:

I remove browserslist override which was causing errors on: npm run setup -s

**Why**:

Have to build.

**How**:

removed pin in package.json

**Checklist**:

- N/A [ ] Documentation added to the
- N/A [ ] Tests
- N/A [ ] TypeScript definitions updated
- [ x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,16 +51,12 @@
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^5.11.6",
-    "caniuse-lite": "1.0.30001502",
     "jest-in-case": "^1.0.2",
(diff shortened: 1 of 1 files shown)


eps1lon:
Have you checked why we added this? It used to ensure we always target the same syntax level. Changing it randomly on publish is unsafe since we might produce syntax that's only valid for newer Node.js/Browser versions i.e. a breaking change.

nifgraup:
It was added for node 14 & 16 compatability #1243, both dropped in #1255.

PR #1347, opened 2025-01-15
createEvent: Only run dataTransfer workarounds in json environments

nifgraup:
**What**:

Fixes for browser based environments, not present when used with jest/jsdom.

**Why**:

createEvent has a workaround for missing event.dataTransfer objects as they don't exist in jsdom environments like jest #339 . Browser-based environments will have native DataTransfer objects and that should be used as the tests may rely on them being the real thing.

**How**:

A simple check to see if the workaround is needed.

**Checklist**:

- N/A[ ] Documentation added to the
- I wish I could write a test case but this requires a test environment other than node.js/jest/jsdom [ ] Tests
- N/A [ ] TypeScript definitions updated
- [x] Ready to be merged


nifgraup:
build failure is dealt with in #1346

PR #1362, opened 2025-07-26
chore: unpin aria-query

benmccann:
**What**:

Unpin `aria-query`

**Why**:

I have multiple versions of `aria-query` in my project. This allows it to dedupe

**How**:

See https://github.com/testing-library/dom-testing-library/pull/1341#issuecomment-2486918317 for a discussion of why the CI is failing. It would be nice to merge that PR and then add a lockfile to this project

**Checklist**:

- [x] Tests - N/A
- [x] TypeScript definitions updated - N/A
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "@babel/code-frame": "^7.10.4",
     "@babel/runtime": "^7.12.5",
     "@types/aria-query": "^5.0.1",
-    "aria-query": "5.3.0",
+    "aria-query": "^5.3.2",
     "chalk": "^4.1.0",
     "dom-accessibility-api": "^0.5.9",
     "lz-string": "^1.5.0",


MatanBobi:
Thanks for this @benmccann but we're currently not willing to remove the pin. aria-query aren't following semantic versioning tightly and every minor change which seems like an addition on their end might break existing tests for users of testing-library.
If you have any issues with conflicting versions of aria-query, I recommend using overrides to enforce one version is installed.

benmccann:
Can we reopen this and do the upgrade without the pin?

MatanBobi:
Every upgrade we do to aria-query requires several checks based on aria-query's release notes to verify if there were any changes that might cause a breaking change to our users.
If you're willing to take that on, I'll be happy to re-open the PR. Unfortunately I don't have the capacity currently to review the changes in aria-query.

benmccann:
5.3.2 is certainly not breaking and 5.3.1 has only a few changes: https://github.com/A11yance/aria-query/blob/main/CHANGELOG.md#531

I'm unsure of what types of changes have caused users of this project problems in the past. We use this library in Svelte and would be fine to upgrade based on the changelog, but I'm not exactly sure how the library is used here and if these changes would be considered safe or not.

PR #1368, opened 2025-11-12
Switch to trusted publishing

eps1lon:
Followed https://docs.npmjs.com/trusted-publishers

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -6,8 +6,6 @@ on:
       # e.g. "12.x" or "8.x"
       - '[0-9]+.x'
       - 'main'
-      - 'next'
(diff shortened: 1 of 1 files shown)


.github/workflows/validate.yml
```diff
@@ -90,18 +92,15 @@ jobs:
         run: npm run build
 
       - name: 🚀 Release
-        uses: cycjimmy/semantic-release-action@v2
+        uses: cycjimmy/semantic-release-action@v5
         with:
-          semantic_version: 17
+          semantic_version: 25
```
eps1lon: See https://github.com/cycjimmy/semantic-release-action/issues/269#issuecomment-3503478001

PR #1369, opened 2025-11-12
feat: Use new `waitForWrapper` around `waitFor` scope 

eps1lon:
BREAKING CHANGE: `asyncWrapper` is no longer used around `waitFor` scope

`user-event` has started using `asyncWrapper` around small async steps. RTL needs to disable the act environment around `waitFor`. We're resolving this conflict with a new wrapper specific to `waitFor`.

RTL will use `waitFor` to disable act environment and configure `asyncWrapper` to use `act`.  All user-event has to do is ensure it awaits the `fireEvent` since these are now async.

### DIFF
--- a/src/__tests__/wait-for.js
+++ b/src/__tests__/wait-for.js
@@ -328,7 +328,7 @@ test('does not work after it resolves', async () => {
         return undefined
       }
     },
-    asyncWrapper: async callback => {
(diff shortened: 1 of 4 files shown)


jotaRenan:
hi, @eps1lon! Any updates on this?

PR #1371, opened 2026-01-14
fix: propagate hidden flag to computeAccessibleName

### LINKED ISSUE #820: waitFor + getByRole causing severe delays

foxaltus:
**What**: this PR fixes https://github.com/testing-library/dom-testing-library/issues/820#issuecomment-3434470548

**Why**: *ByRole APIs are slow, because the DOM environment relies on `getComputedStyles` to compute elements visibility. This can be mitigated by using the `{ hidden: true }` option. However, when using *ByRole APIs along with the `name` option, the performance toll is still being paid because `computeAccessibleName` is called internally, which does not currently ignore hidden elements.

**How**: this PR propagates the `hidden` option to `computeAccessibleName`, which avoids paying the performance toll. 

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -197,6 +197,20 @@ test('can include inaccessible roles', () => {
   expect(getByRole('list', {hidden: true})).not.toBeNull()
 })
 
+test('can include inaccessible roles when searching by name', () => {
+  // this behavior deviates from the spec which includes "any descendant"
+  // if visibility is hidden. However, chrome a11y tree and nvda will include
+  // the following markup. This behavior might change depending on how
+  // https://github.com/w3c/aria/issues/1055 is resolved.
+  const {getByRole} = render(
+    '<button>hello <span hidden>hidden</span> world!</button>',
+  )
+
+  expect(
+    getByRole('button', {name: 'hello hidden world!', hidden: true}),
+  ).not.toBeNull()
+})
+
 test('can be filtered by accessible name', () => {
   const {getByRole} = renderIntoDocument(
     `
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -197,11 +197,7 @@ test('can include inaccessible roles', () => {
   expect(getByRole('list', {hidden: true})).not.toBeNull()
 })
 
-test('can include inaccessible roles when searching by name', () => {
-  // this behavior deviates from the spec which includes "any descendant"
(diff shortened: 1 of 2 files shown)


foxaltus:
Hmm I realize this is not enough...

The changes allow to bypass the expensive `isHidden` check at https://github.com/eps1lon/dom-accessibility-api/blob/main/sources/accessible-name-and-description.ts#L601, but not the ones at https://github.com/eps1lon/dom-accessibility-api/blob/main/sources/accessible-name-and-description.ts#L388. So the toll is still being paid :(

See: most time is being spent in `computeMiscTextAlternative`:
Screenshot of the flame graph showing most time being spent inside computeMiscTextAlternative
(FYI this test looks by table rows by name in a 25 rows x 10 cols table)

foxaltus:
At this point I think that maybe `testing-library` should offer an option to fake `getComputedStyle` (similar to the `computedStyleSupportsPseudoElements` option). In some of my tests (dealing with tables) this improves the performance by a factor of 10.

If this sounds too much of a bazooka solution, maybe we should instead consider adding an option at `*ByRole` level, such as `namingStrategy?: 'fast' | 'accurate'`:

- `fast` would send a mock `getComputedStyle` implementation to `computeAccessibleName`.
- `accurate` (default) would use the current implementation (+ I think it could honour the hidden option, as already done in this PR)

Thoughts?

PR #1375, opened 2026-04-29
fix: escape square brackets in ID selectors

mixelburg:
## Problem

When an element has an ID containing square brackets (e.g. `foo[1].bar`), calling `getByLabelText` (or queries that resolve `aria-labelledby`) throws:

```
SyntaxError: foo[1].bar is not a valid selector
```

This happens because `querySelector("[id=\"foo[1].bar\"]")` interprets `[1]` as a CSS attribute selector, which is invalid.

## Fix

Use `CSS.escape()` to properly escape IDs before interpolating them into attribute selectors. Applied in two locations:

- `src/queries/label-text.ts` — `getTagNameOfElementAssociatedWithLabelViaFor`
- `src/label-helpers.ts` — `getLabels` (aria-labelledby lookup)

`CSS.escape()` is available in all environments supported by dom-testing-library (Node >= 18).

Fixes #1404

### DIFF
--- a/src/label-helpers.ts
+++ b/src/label-helpers.ts
@@ -64,7 +64,7 @@ function getLabels(
   return labelsId.length
     ? labelsId.map(labelId => {
         const labellingElement = container.querySelector<HTMLElement>(
-          `[id="${labelId}"]`,
+          `[id="${CSS.escape(labelId)}"]`,
         )
         return labellingElement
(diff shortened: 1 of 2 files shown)


mixelburg:
Closing due to inactivity. Over 8 weeks with no review.

PR #1376, opened 2026-04-30
fix: support inert attribute and <search> element in role queries

### LINKED ISSUE #1359: `getByRole('search')` does not find `<search />`
- `@testing-library/dom` version: Unknown
I ran `npm list | grep @testling-library` and I didn't get `@testing-library/dom`:
```
├── @testing-library/jest-dom@6.5.0
├── @testing-library/react@15.0.6
├── @testing-library/user-event@14.6.1
```

- Testing Framework and version: Vitest 3.1.3
- DOM Environment: jsdom 22.1.0

### Relevant code or config:

```js
  it('contains search role', async () => {
    render(<search />);
    expect(screen.getByRole('search')).toBeInTheDocument();
  });

```

### What you did:

I searched the search element by role `search` and expected it will be found. As it seems the [proper role for that element](https://www.w3.org/TR/html-aria/#el-search).

### What happened:

My test failed:

```
TestingLibraryElementError: Unable to find an element with the role "search"

There are no available roles.

Ignored nodes: comments, script, style
<body>
  <div>
    <search />
  </div>
</body>
 ❯ Object.getElementError ../../node_modules/@testing-library/dom/dist/config.js:37:19
 ❯ ../../node_modules/@testing-library/dom/dist/query-helpers.js:76:38
 ❯ ../../node_modules/@testing-library/dom/dist/query-helpers.js:52:17
 ❯ ../../node_modules/@testing-library/dom/dist/query-helpers.js:95:19
 ❯ test.spec.tsx:30:19
     28|   it('contains search role', async () => {
     29|     render(<search />);
     30|     expect(screen.getByRole('search', { hidden: true })).toBeInTheDocument();
       |                   ^
     31|   });
```

### Reproduction:

https://stackblitz.com/edit/dtl-template-ryhamkss?file=src%2Fmain.test.ts

### Problem description:

The search element is not found by `getByRole('search')`. I was expecting the role search was okay for that element since that's mentioned in [this table](https://www.w3.org/TR/html-aria/#el-search).

### Suggested solution:

Maybe a change like the one done [to solve this issue](https://github.com/testing-library/dom-testing-library/issues/1100) could be the answer.

### LINKED ISSUE #1364: Elements with an inert="" attribute aren't being hidden from the accessibility tree
- `@testing-library/dom` version: 10.4.0
  - testing-library/cypress@10.0.1
  - cypress@13.17.0
- DOM Environment: 
  - Chrome v138
  - node v22.11.0

### Relevant code or config:

Codesandbox URL: https://codesandbox.io/p/sandbox/fwqv27

```js
describe('elements hidden from the accessibility tree', () => {
  it('inert elements should be hidden', () => {
    cy.visit('https://fwqv27.csb.app/', { failOnStatusCode: false });

    // Required to view the codesandbox preview
    cy.findByRole('link', { name: 'Yes, proceed to preview' }).click();

    cy.findByRole('button', { name: 'First', hidden: true }).click();
    cy.findByRole('button', { name: 'Second', hidden: true }).click();
  });
});

```

### What you did:

Tried to target a button nested in a container with an inert attribute with the `hidden` option set to `true`.

### What happened:

The assertion could not target the button as it isn't recognised as a hidden element.

### Reproduction:

Please refer to the codesandbox URL and code snippet above.

### Problem description:

According to MDN any inert elements and their children should be hidden from the accessibility tree ([see source](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert#:~:text=Hides%20the%20element%20and%20its%20content%20from%20assistive%20technologies%20by%20excluding%20them%20from%20the%20accessibility%20tree.)):

> Specifically, `inert` does the following:
> ...
> - Hides the element and its content from assistive technologies by excluding them from the accessibility tree.

An example scenario from the `inert` HTML specification ([see source](https://html.spec.whatwg.org/multipage/interaction.html#modal-dialogs-and-inert-subtrees)):

> A [Document](https://html.spec.whatwg.org/multipage/dom.html#document) document is blocked by a modal dialog subject if subject is the topmost [dialog](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element) element in document's [top layer](https://drafts.csswg.org/css-position-4/#document-top-layer). While document is so blocked, every node that is [connected](https://dom.spec.whatwg.org/#connected) to document, with the exception of the subject element and its [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) descendants, must become [inert](https://html.spec.whatwg.org/multipage/interaction.html#inert).

It also provides a technical example in the following section ([see example](https://html.spec.whatwg.org/multipage/interaction.html#modal-dialogs-and-inert-subtrees)).

### Suggested solution:

We should perform a check within the `isSubtreeInaccessible` method to see if the element is inside of an inert subtree:

https://github.com/testing-library/dom-testing-library/blob/2c570553d8f31b008451398152a9bd30bce362b3/src/role-helpers.js#L15-L29

We could potentially use the `.closest()` method to determine whether there is an inert ancestor ([see docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)):

```js
if (element.hasAttribute('inert') || element.closest('[inert]')) {
  return true;
}
```

mixelburg:
## Summary

This PR addresses two related issues in role-based queries:

### 1. `inert` attribute not hiding elements from a11y tree (#1364)

Elements with the [`inert`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert) attribute (and their descendants) should be excluded from the accessibility tree. Per the [HTML spec](https://html.spec.whatwg.org/multipage/interaction.html#inert), `inert` makes the element and all its contents hidden from assistive technologies.

**Fix:** Added a check in `isSubtreeInaccessible` that returns `true` when:
- The element itself has the `inert` attribute, OR
- The element has an ancestor with the `inert` attribute (via `element.closest('[inert]')`)

### 2. `<search>` element not found by `getByRole('search')` (#1359)

The `<search>` HTML element has an implicit ARIA role of `search` per the [HTML-ARIA spec](https://www.w3.org/TR/html-aria/#el-search), but `aria-query@5.3.0` does not include this mapping in its `elementRoles` map.

**Fix:** Added a fallback in `getImplicitAriaRoles` that returns `[search]` for `<search>` elements. This workaround can be removed once `aria-query` publishes a version that includes this mapping (the mapping exists in aria-query's `main` branch but is not yet released).

## Test plan

- Added test case for `<div inert />` to the existing `shouldExcludeFromA11yTree` test matrix
- Added test case for elements nested inside an inert subtree
- Added test case for `<search>` element implicit role
- All 669 existing tests pass
- Full test suite passes: `29 passed, 29 total`

## Related issues

- Fixes #1364
- Fixes #1359

### DIFF
--- a/src/__tests__/role-helpers.js
+++ b/src/__tests__/role-helpers.js
@@ -197,6 +197,13 @@ test('getImplicitAriaRoles returns expected roles for various dom nodes', () => 
   expect(getImplicitAriaRoles(input)).toEqual(['textbox'])
 })
 
+test('getImplicitAriaRoles returns search role for <search> element', () => {
+  const {container} = render('<search>Search content</search>')
+  expect(getImplicitAriaRoles(container.querySelector('search'))).toEqual([
+    'search',
+  ])
+})
+
 test.each([
   ['<div />', false],
   ['<div aria-hidden="false" />', false],
@@ -205,9 +212,17 @@ test.each([
   ['<div style="display: none;"/>', true],
   ['<div style="visibility: hidden;"/>', true],
   ['<div aria-hidden="true" />', true],
+  ['<div inert />', true],
 ])('shouldExcludeFromA11yTree for %s returns %p', (html, expected) => {
   const {container} = render(html)
   container.firstChild.appendChild(document.createElement('button'))
 
   expect(isInaccessible(container.querySelector('button'))).toBe(expected)
+})
+
+test('elements nested in inert subtree are inaccessible', () => {
+  const {container} = render(
+    '<div inert><span><button>click me</button></span></div>',
+  )
+  expect(isInaccessible(container.querySelector('button'))).toBe(true)
 })
--- a/src/role-helpers.js
+++ b/src/role-helpers.js
@@ -21,6 +21,10 @@ function isSubtreeInaccessible(element) {
     return true
   }
 
+  if (element.hasAttribute('inert') || element.closest('[inert]')) {
+    return true
+  }
+
   const window = element.ownerDocument.defaultView
   if (window.getComputedStyle(element).display === 'none') {
     return true
@@ -72,6 +76,15 @@ function getImplicitAriaRoles(currentNode) {
     if (match(currentNode)) {
       return [...roles]
     }
+  }
+
+  // Handle elements not yet in aria-query's elementRoles map
+  // See: https://github.com/testing-library/dom-testing-library/issues/1359
+  // The <search> element has an implicit 'search' role per the HTML-ARIA spec
+  // but aria-query 5.3.0 does not include it in elementRoles.
+  // This can be removed once aria-query includes this mapping.
+  if (currentNode.tagName === 'SEARCH') {
+    return ['search']
   }
 
   return []


philwolstenholme:
Nice! I think I'll have a use case for this soon so might try it out via patching the main package

philwolstenholme:
Although have you seen the history on https://github.com/testing-library/dom-testing-library/pull/1366? @eps1lon had some concerns about how `dialog` elements handle `inert`, so it's not as simple as just 'anything inside an inert parent is also inert', unfortunately.

That said, for my use case this will probably be fine, so I'll probably still apply the patch.

mixelburg:
Closing due to inactivity. The reviewer raised valid concerns about dialog/inert interaction that need further investigation. Happy to reopen with a revised approach.

PR #1377, opened 2026-05-01
fix: wrap filterNode in arrow function to ignore extra callback args

### LINKED ISSUE #1360: prettyDOM filterNode argument is called with more args than expected based on type
### Relevant code or config:

```js
prettyDOM(container, undefined, {
  filterNode: (node, unexpectedArg1, unexpectedArg2, shouldBeUndefined) => {
    console.log("arg 1: node - ", node);
    console.log("arg 2: unexpected - ", unexpectedArg1);
    console.log("arg 3: unexpected - ", unexpectedArg2);
    console.log("arg 4: undefined (none) - ", shouldBeUndefined);
    debugger;
    return true;
  },
})
```

### What you did / what happened

The `filterNode` arg in `prettyDOM` is [typed as having one argument](https://github.com/testing-library/dom-testing-library/blob/a86c54ccda5242ad8dfc1c70d31980bdbf96af7f/types/pretty-dom.d.ts#L8):
```
filterNode?: (node: Node) => boolean
```

so I was expecting it to be called with one argument, but it was called with three.

### Reproduction:

https://codesandbox.io/p/sandbox/wizardly-shirley-yp96n8?file=%2Fsrc%2F__tests__%2Findex.test.js%3A8%2C27

See in the console that there are three args passed to `filterNode`:
Image

### Problem description:

I think it's being called with three arguments because `filterNode` is [being passed into `filter`](https://github.com/testing-library/dom-testing-library/blob/a86c54ccda5242ad8dfc1c70d31980bdbf96af7f/src/DOMElementFilter.ts#L256) and `filter` [has three args](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter#callbackfn): `element`, `index`, and `array`.

### Suggested solution:

Can either the typing of `filterNode` be updated to indicate that it is called with three arguments or the usage be updated to only call it with the one expected `node` argument?

Thank you!

mixelburg:
Fixes #1360

When printing filtered DOM nodes, Array.prototype.filter passes three arguments (element, index, array) to the callback, but filterNode expects only (node: Node). While this works in JS due to ignored extra arguments, wrapping with an explicit arrow function `.filter(node => filterNode(node))` ensures only the element is passed, matching the type signature.

### DIFF
--- a/src/DOMElementFilter.ts
+++ b/src/DOMElementFilter.ts
@@ -253,7 +253,7 @@ export default function createDOMElementFilter(
         printChildren(
           Array.prototype.slice
             .call(node.childNodes || node.children)
-            .filter(filterNode),
+            .filter(node => filterNode(node)),
           config,
           indentation + config.indent,
           depth,


mixelburg:
Closing due to inactivity. Over 8 weeks with no review.

PR #1380, opened 2026-05-13
fix: support form-associated custom elements in isLabelable

### LINKED ISSUE #1343: `isLabelable` does not support custom elements with `formAssociated = true`
- `@testing-library/dom` version: 10.0.0
- Testing Framework and version: jest 29.7.0 / @testing-library/jest-dom 6.5.0 / @testing-library/react

### Relevant code or config:

```js
    const { getByLabelText } = render(<PktTextinput labelText="Input Label" id="inputId" />)
    await window.customElements.whenDefined('pkt-textinput')
    const inputElement = getByLabelText('Input Label')
```

### What you did:

Tried to use getByLabelText to query a form associated custom element by label text.

### What happened:

> ´TestingLibraryElementError: Found a label with the text of: Input Label, however the element associated with this label (<pkt-textinput />) is non-labellable [https://html.spec.whatwg.org/multipage/forms.html#category-label]. If you really need to label a <pkt-textinput />, you can use aria-label or aria-labelledby instead.´

### Reproduction:

Not needed as the problem is evident from this code in dom-testing-library:

```
function isLabelable(element: Element) {
  return (
    /BUTTON|METER|OUTPUT|PROGRESS|SELECT|TEXTAREA/.test(element.tagName) ||
    (element.tagName === 'INPUT' && element.getAttribute('type') !== 'hidden')
  )
}
```

### Problem description:

Web component API has `formAssociated` and `ElementInternals` functionality to let a custom element in HTML behave like a fully valid form element that should support being targeted with a label. This is not supported in dom testing because only the “standard” built-in list of HTML form elements is supported in the `isLabelable` function.

I checked the W3C specs and formAssociated custom elements should indeed support being targeted by a `label` with `for`.

### Suggested solution:

Add a test in `isLabelable` to check if the element is a form associated custom element.

mixelburg:
## Description

Form-associated custom elements using the `ElementInternals` API (`static formAssociated = true`) can be targeted by `<label>` elements and should be considered labelable by `getByLabelText()`.

Previously, `isLabelable()` only checked for standard HTML form elements (`BUTTON`, `INPUT`, `METER`, `OUTPUT`, `PROGRESS`, `SELECT`, `TEXTAREA`), causing `getByLabelText()` to fail for form-associated custom elements with the error:

> Found a label with the text of: Input Label, however the element associated with this label (<custom-input/>) is non-labellable

## Solution

Added a check for custom elements (tags containing a hyphen) that have the `attachInternals()` method. When available, we call it and check if `internals.formAssociated === true`. This is wrapped in a try-catch to gracefully handle:
- Environments that don't support `ElementInternals`
- Elements that throw when `attachInternals()` is called on non-form-associated custom elements

## Changes

- `src/label-helpers.ts`: Extended `isLabelable()` to detect form-associated custom elements

## Testing

- All 666 existing tests pass
- The change is fully backward compatible
- In environments without `ElementInternals` support (older jsdom), behavior is unchanged

Fixes #1343

### DIFF
--- a/src/label-helpers.ts
+++ b/src/label-helpers.ts
@@ -58,9 +58,11 @@ function isLabelable(element: Element) {
   // Support form-associated custom elements
   if (
     element.tagName.includes('-') &&
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call
     typeof (element as any).attachInternals === 'function'
   ) {
     try {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call
       const internals = (element as any).attachInternals()
       return internals.formAssociated === true
     } catch {
--- a/src/label-helpers.ts
+++ b/src/label-helpers.ts
@@ -48,10 +48,27 @@ function getRealLabels(element: Element) {
 }
 
 function isLabelable(element: Element) {
-  return (
+  if (
     /BUTTON|METER|OUTPUT|PROGRESS|SELECT|TEXTAREA/.test(element.tagName) ||
     (element.tagName === 'INPUT' && element.getAttribute('type') !== 'hidden')
-  )
+  ) {
+    return true
+  }
+
+  // Support form-associated custom elements
+  if (
+    element.tagName.includes('-') &&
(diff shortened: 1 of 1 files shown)


mixelburg:
Closing due to inactivity. Over 5 weeks with no review.

PR #1385, opened 2026-08-14
fix: support the <search> element in role queries

### LINKED ISSUE #1359: `getByRole('search')` does not find `<search />`

FrancoKaddour:
Closes #1359

The `<search>` element has an implicit ARIA role of `search` per the HTML-AAM spec, but `aria-query@5.3.0` doesn't map it. As a result `getByRole('search')` can't find a `<search>` element, and `getImplicitAriaRoles` returns `[]` for it.

The fix adds `<search>` in the **two** places that consume aria-query's maps:
- `getImplicitAriaRoles` (computes an element's implicit role)
- `makeRoleSelector` (builds the candidate set for a role query)

Both are gated behind a comment noting they can be removed once aria-query maps `<search>`. I found that patching only `getImplicitAriaRoles` (as in the earlier #1376) isn't enough — `makeRoleSelector` filters the candidate elements first, so `<search>` never reaches the role check. The added integration test (`getByRole('search')`) covers that path; there's also a unit test for `getImplicitAriaRoles`.

Full suite green (668 tests). #1376's contentious `inert` change is intentionally left out — this is only the uncontroversial `<search>` support.

### DIFF
--- a/src/__tests__/role-helpers.js
+++ b/src/__tests__/role-helpers.js
@@ -197,6 +197,14 @@ test('getImplicitAriaRoles returns expected roles for various dom nodes', () => 
   expect(getImplicitAriaRoles(input)).toEqual(['textbox'])
 })
 
+test('getImplicitAriaRoles returns search role for a <search> element', () => {
+  const {container} = render('<search>content</search>')
+
+  expect(getImplicitAriaRoles(container.querySelector('search'))).toEqual([
+    'search',
+  ])
+})
+
 test.each([
   ['<div />', false],
   ['<div aria-hidden="false" />', false],
--- a/src/__tests__/role.js
+++ b/src/__tests__/role.js
@@ -56,6 +56,13 @@ test('when hidden: true logs available roles when it fails', () => {
   `)
 })
 
+test('can be found by its implicit role when using a <search> element', () => {
+  const {getByRole, container} = render(
+    '<search><input type="search" /></search>',
+  )
+  expect(getByRole('search')).toBe(container.querySelector('search'))
+})
+
 test('logs error when there are no accessible roles', () => {
   const {getByRole} = render('<div />')
   expect(() => getByRole('article')).toThrowErrorMatchingInlineSnapshot(`
--- a/src/queries/role.ts
+++ b/src/queries/role.ts
@@ -313,6 +313,15 @@ function makeRoleSelector(role: ByRoleMatcher) {
     Array.from(roleRelations).map(({name}) => name),
   )
 
+  // aria-query 5.3.0 does not map the `<search>` element to its implicit
+  // `search` role, so it is missing from `roleElements`. Add it here to keep
+  // the candidate selector in sync with `getImplicitAriaRoles`. This can be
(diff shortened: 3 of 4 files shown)


FrancoKaddour:
The `validate` failure here is the `@types/node` × `typescript@4` typecheck breakage in `node_modules/@types/node/ffi.d.ts` (`TS1139: Type parameter declaration expected`, etc.) — not this change. The diff is source + tests only, my source has no type errors, and the full suite passes locally (668 tests). It's currently hitting other open PRs too: there's no committed lockfile, so CI installs the latest `@types/node`, which ships syntax `typescript@^4.1.2` can't parse.

I can't re-run the workflow from a fork PR — could a maintainer give it a re-run? And happy to open a separate PR pinning `@types/node` (or adding an `overrides` entry) to the last TS4-compatible version if that'd help unblock CI repo-wide.

FrancoKaddour:
Quick update: #1386 (the `@types/node` pin) is now green on CI, which confirms the failure here is the unrelated `@types/node` × TypeScript 4 typecheck issue, not this change. Once #1386 lands, this PR's CI will pass — happy to rebase onto it if that makes review easier.

